PR #3, opened 2017-11-03
Add the packaging metadata to build the zokrates snap

elopio:
This package will let you publish the latest zokrates in the Ubuntu store, and from there reach many users on all the supported Ubuntu versions, and Linux distributions. You just have to go to https://build.snapcraft.io and enable the automated continuous delivery.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,23 @@
+name: zokrates
+version: git
+summary: A toolbox for zkSNARKS on Ethereum
+description: |
+  A higher-level language and compiler, which transforms a more convenient
+  representation into verifiable programs based on zkSNARKS.
+  Additionally, generates Ethereum Smart Contraccts, which verify the results
+  on-chain.
+
+grade: devel # must be 'stable' to release into candidate/stable channels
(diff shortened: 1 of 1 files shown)


elopio:
The continuous delivery will build a new snap for every change on your repo, and publish it to the edge channel. This is a great way to get more testers and contributors, who will be able to install with a single command and will get automatic updates. 

The projects from parity are already there:
https://github.com/paritytech/parity#install-from-the-snap-store
https://github.com/paritytech/parity-bitcoin#installing-the-snap

It would be great for us to get feedback from more projects written in rust. If you have any questions, please let me know.

Schaeff:
I don't think this is necessary now, shall we close @JacobEberhardt ?

JacobEberhardt:
Agree, I don't think this is necessary at this point.

elopio:
@Schaeff @JacobEberhardt alright, let me know if you change your mind, or have any remaining questions. Recently we have seen a huge bump of our stats with the recent big releases of skype, spotify and slack. And parity also has just been released to stable: https://snapcraft.io/parity
I would love to have more software written in rust available in the store, but I can understand your position.

pura vida :)

Issue #7, opened 2017-11-09
Possible Less-Than Operator Issue?


curlysemi:
Hello,

I've been playing with this toolbox for a few days and am attempting some somewhat complex code, but I have ran into some strange behavior when using the the less-than operator.

The issue doesn't arise when the code is simple. An example starting point:

```
def lt(a,b):
    return if a < b then 1 else 0 fi

def main():
    return lt(31337, 2197)
```

As expected, the returned value when `main()` is called is `0`.

A minor note (mostly for my own sanity): we can make a slight adjustment and get the same result:
```
def lt(a,b):
    return if a < b then 1 else 0 fi

def cutoff():
    return 31337

def main():
    return lt(cutoff(), 2197)
```

Now, let's introduce some slight complexity, allowing `31337` to be compared to another value—a value that is computationally arrived upon based off of an input:
```
def getThing(index):
    result = 3
    result = if index == 0 then 13 else result fi
    result = if index == 1 then 23 else result fi
    result = if index == 2 then 43 else result fi
    result = if index == 3 then 53 else result fi
    result = if index == 4 then 73 else result fi
    result = if index == 5 then 83 else result fi
    return result

def cubeThing(thing):
    return thing**3

def main(index):
    thing = getThing(index)
    thing = cubeThing(thing)
    return thing
```

We can call our new `main()` function with an argument of `0`, and the returned value is `2197`. So far so good.

Things seem to go awry when we try to combine our new stuff with the old stuff:

```
def lt(a,b):
    return if a < b then 1 else 0 fi

def cutoff():
    return 31337

def getThing(index):
    result = 3
    result = if index == 0 then 13 else result fi
    result = if index == 1 then 23 else result fi
    result = if index == 2 then 43 else result fi
    result = if index == 3 then 53 else result fi
    result = if index == 4 then 73 else result fi
    result = if index == 5 then 83 else result fi
    return result

def cubeThing(thing):
    return thing**3

def main(index):
    thing = getThing(index)
    thing = cubeThing(thing)
    return lt(cutoff(), thing)
```
This most recent `main()` is the mostly the same as the previous one that yielded `2197`. The only difference between this new code and the previous is that, instead of returning the value, we compare `31337` to the value (just as in the *very* first example). When we call this newest `main()` with an argument of `0`, we get an unexpected result!
```
Returned (~out): 1
```
[...]

PR #9, opened 2017-11-10
Fix for Issue #7: Fix for 'sym_{i}_{j}' and 'sym_{i}_{j}_b{k}' flattening

curlysemi:
Addressing Issue #7 
Fix for symbol reuse when flattening (normally not an issue in simple programs, as normal cases were accounted for; the problem was that the `_b`s were not using the adjusted root names) — I don't know rust at all, so there may be possible issues with these changes. Likely will want a second commit/fixup to change the order of the params (the new `all_vars` should probably be first).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,6 +1,6 @@
 [root]
 name = "zokrates"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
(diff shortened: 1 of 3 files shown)


curlysemi:
Unfortunately, while 670ab63e allows the code in issue #7 to run, an even more complex program has a similar symbol issue. Currently working on a resolution.

curlysemi:
Closes until the more advanced cases are sorted out.

PR #10, opened 2017-11-16
Add "Where's Waldo" example

Schaeff:
I played around with ZoKrates and implemented a version of Where's Waldo.
The idea is that Waldo is a non-prime number, and other people are prime numbers.
There's no guarantee that there's exactly one Waldo (or that there's one at all)

I also updated the factorization example to only allow non trivial solutions.

### DIFF
--- a/examples/factorization.code
+++ b/examples/factorization.code
@@ -1,4 +1,7 @@
 // a and b are factorization of c
 def main(c):
+  a1 = if a == 1 then 0 else 1 fi // "a != 1"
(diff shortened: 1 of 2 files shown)


nicola:
Hey @Schaeff do you mind writing some more explanation at the top of the document?

Schaeff:
@nicola sure, which part isn't clear? Enforcement of non-trivial factorisation, or the Waldo representation?

nicola:
The Waldo representation and the waldo problem

PR #15, opened 2017-11-26
Make private inputs explicit

Schaeff:
See github.com/JacobEberhardt/ZoKrates/issues/13

Todo:
- [ ] Tests
- [ ] Agree on syntax (for now `(a, b, private c, private d)`)
- [ ] Ban undeclared variables at compile time (currently at compute-witness time)


src/absy.rs
```diff
@@ -288,19 +289,11 @@ impl<T: Field> Expression<T> {
                         }
                         assert_eq!(num, T::zero());
                     } else {
-                        println!(
-                            "Could not calculate variable {:?}, inputs: {:?}",
+                        panic!(
```
Schaeff: Theoretically we could remove this part as semantic checks ensure all variables are defined and the CLI ensures all main arguments are specified. Thoughts?

PR #20, opened 2018-02-01
Update sudokuchecker.code

hanyunxu:
In the `check row correctness` part, the 3rd row should be c11, c12...
I ran the example with a 4*4 case but it failed. This change fixed it.


JacobEberhardt:
Thank you!

JacobEberhardt:
Thank you!

PR #30, opened 2018-03-05
Update readme

Schaeff:
Make the README more... readable.

One issue is that the access to `templates/sol_verification.template` forces to run the CLI from project root (`./target/release/zokrates` rather than `cd ./target/release` and `./zokrates`)


Schaeff:
[Update README](https://trello.com/c/GSudogRs/1-update-readme)

JacobEberhardt:
We can do minor changes in additional commits, but this i much better than what we have now, so I merge right away.

PR #31, opened 2018-03-05
Feature/r1cs import

JacobEberhardt:
Import of Standard R1CS files for interoperability.


src/r1cs.rs
```diff
@@ -331,6 +344,80 @@ pub fn r1cs_program<T: Field>(
...
+/// * The Rank 1 Constraint System (R1CS) is defined as:
+/// * `<A,x>*<B,x> = <C,x>` for a witness `x`
+/// * Since the matrices in R1CS are usually sparse, the following encoding is used:
+/// * For each constraint (i.e., row in the R1CS), only non-zero values are supplied and encoded as a tuple (index, value).
+///
+/// # Arguments
+///
+/// * r1cs - R1CS in standard JSON data format
+
+pub fn flattened_program<T: Field>(
```
Schaeff: Could this be in an `impl` for `R1CS` to call `R1CS::from(json).to_flattened_program()` ?

JacobEberhardt: Yes, good idea. Let's add that after the format has been standardized and the data structure fixed.

Schaeff: You mean not in this PR?

JacobEberhardt: That depends if we want to merge this preliminary version or finalize it first.

nicola:
woho! this is big!

Kyroy:
jenkins-test

PR #45, opened 2018-04-12
Add an optimizer

Schaeff:
The flattener generates a lot of redundant code like

```
b = a
c = b
d = c
```

and also keeps long variable names and flattened functions

Introduce an opt-out optimizer to reduce overhead


Schaeff:
[Reduce .code size by removing "a = b" constraints](https://trello.com/c/fkb9Hg1n/20-reduce-code-size-by-removing-a-b-constraints)

PR #46, opened 2018-04-12
Support function overloading in flattener

Schaeff:
Currently, the flattener names variables generated by function calls based on function name and index of the call:  `foo_42_bar_33` for the 33rd call to `bar` inside the 42nd call to `foo`

This can break when functions are overloaded, like the following example:

```
def foo()
     return 1
def foo()
     return 1, 2
def main()
     a = foo()
     b, c = foo()
     return 1
```

I'm not 100% familiar with how the flattener works, so would welcome some feedback @JacobEberhardt

### DIFF
--- a/src/flatten.rs
+++ b/src/flatten.rs
@@ -903,4 +903,69 @@ mod multiple_definition {
             Statement::Definition("a".to_string(), Expression::Number(FieldPrime::from(1)))
         );
     }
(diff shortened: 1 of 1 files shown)


JacobEberhardt:
We decided to add the signature after the function name separated by underscore.

PR #48, opened 2018-04-12
Add missing argument.

lovesh:
Fixes https://github.com/JacobEberhardt/ZoKrates/issues/47
Signed-off-by: Lovesh Harchandani <lovesh.bond@gmail.com>

### DIFF
--- a/examples/add.code
+++ b/examples/add.code
@@ -1,5 +1,5 @@
 // only using add, no need to flatten
-def main():
+def main(a):  // or def main(private a):
   b = a + 5
   c = a + b + a + 4
   d = a + c + a + b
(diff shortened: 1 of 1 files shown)


examples/add.code
```diff
@@ -1,5 +1,5 @@
 // only using add, no need to flatten
-def main():
+def main(a):  // or def main(private a):
```
Schaeff: Thanks for the contribution! I think you can remove that comment, this example is fine without `private`

lovesh: Done. Thanks for this project :+1:

Schaeff: Merged! Thank you!

PR #51, opened 2018-04-14
Catch undefined variables in multireturn calls

### LINKED ISSUE #39: Semantic checker should catch undefined variable in multi return function call
```
def foo(x):
	return 1

def main():
	a = foo(x) // undefined caught
	return 1	
```

```
def foo(x):
	return 1, 2

def main():
	a, b = foo(x) // undefined not caught
	return 1	
```

Schaeff:
Fixes #39


Kyroy:
@JacobEberhardt This merge broke the develop branch. If you don't want this to happen, enable the "up to date" option in the protected branches.

@Schaeff Could you please fix this? XD

PR #52, opened 2018-04-16
Add an importer for .code files

Schaeff:
Current syntax:

```
import "./foo.code" as bar

def main():
   return bar(1, 2, 3)
```

To decide:

- [x] How to handle overload? Easiest: forbid overloading for imported functions
- [x] How to handle naming? Easiest: imported file has to have only one exposed function, say `main`. Otherwise we'd have to implement `import <fun> from <file>`. Maybe something for later
- [x] Can we import flattened files to prevent recompiling? Maybe not MVP
- ~[ ] Update grammar in parser~ postponed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,3 @@ RUN cd libsnark-$libsnarkcommit \
 ENV LD_LIBRARY_PATH $LD_LIBRARY_PATH:/usr/local/lib
 
 COPY . /root/ZoKrates
-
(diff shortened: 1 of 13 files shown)


JacobEberhardt:
Only needs to become compatible with new flattened type abstraction merged to develop. Can then be merged.
Note: relative Path syntax not perfect yet, e.g., "../" does not work but needs to be "./../".

Issue #43, opened 2018-04-06
panick at 'no entry found for key'


neeps:
Already mentioned this issue in gitter.im chat, but also posting it here so others can see it aswell. Been playing around with the less than "<" operator, to determine if ZoKrates in its current state is capable of performing taxation in a proof-of-concept. However, I met a error in the development.

The source code is:
```
def wtax(debt,wealth):
  x = if wealth < debt then 0 else (wealth-debt) fi
  return x

def main(debt, wealth):
  tax = wtax(debt,wealth)
  return tax
```

Although my application compiles fine, it gives the following error when I try to compute witnesses:

> root@de4db328ebdf:~/ZoKrates/target/release/test# ../zokrates compute-witness -a 2 500000
> Computing witness for:
> def main(debt,wealth):
> ...
> thread 'main' panicked at 'no entry found for key', libcore/option.rs:917:5
> stack backtrace:
>    0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
>              at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
>    1: std::sys_common::backtrace::_print
>              at libstd/sys_common/backtrace.rs:71
>    2: std::panicking::default_hook::{{closure}}
>              at libstd/sys_common/backtrace.rs:59
>              at libstd/panicking.rs:380
>    3: std::panicking::default_hook
>              at libstd/panicking.rs:396
>    4: std::panicking::rust_panic_with_hook
>              at libstd/panicking.rs:576
>    5: std::panicking::begin_panic
>              at libstd/panicking.rs:537
>    6: std::panicking::begin_panic_fmt
>              at libstd/panicking.rs:521
>    7: rust_begin_unwind
>              at libstd/panicking.rs:497
>    8: core::panicking::panic_fmt
>              at libcore/panicking.rs:71
>    9: core::option::expect_failed
>              at libcore/option.rs:917
>   10: <zokrates::absy::Expression<T>>::solve
>   11: <zokrates::absy::Expression<T>>::solve
>   12: <zokrates::absy::Function<T>>::get_witness
>   13: zokrates::main
>   14: std::rt::lang_start::{{closure}}
>   15: std::panicking::try::do_call
>              at libstd/rt.rs:59
>              at libstd/panicking.rs:479
>   16: __rust_maybe_catch_panic
>              at libpanic_unwind/lib.rs:102
>   17: std::rt::lang_start_internal
>              at libstd/panicking.rs:458
[...]

PR #56, opened 2018-04-24
Add prefix substitution

Schaeff:
### Problem

Using a simple HashMap to store substitutions leads to multiple issues linked to binary representations:

1. 
using entire variable names in substitutions leads to the loss of the binary structure. For example, `a_b12` and `a_b13` would be renamed to `_42` and `_43`, so that the witness generation won't "detect" them as being binary coefficients of `a`. This is relevant in the context of an optimizer which needs to preserve the semantics of the unoptimized code.

~2. 
when flattening function calls which include binary coefficient, each binary coefficient is renamed to include the function call's prefix based on its own record in the HashMap. The problem with that is that with binary structures, we cannot assume that a variable to be renamed will appear first as the left side of a definition (see the implementation of `<` which has a bunch of simple definitions of the type `sym_2_b42 = sym_2_b42 * sym_2_b42`, and at the end a complex definition `sym_2 = sum(sym_2_bi * 2**i)` which includes identifiers which were not seen yet in the previous "simple" statements, thus not substituted because not in the map). Related: #43~ this doesn't hold as it's an issue with the flattener

### Solution

Instead of keeping track of binary variables independently, only keep track of prefixes in the hash map. This allows for any binary coefficient of a given variable to be substituted as long as one binary coefficient or the variable itself have been added to the map.

Example:
```
Insert(foo_bar_b12, _123)
Get(foo_bar_b42) # returns _123_b42
```

This solves 1. as the binary aspect is preserved: `x_bi` maps to `y_bi`
This solves 2. as in order to substitute a binary coefficient, it doesn't have to have been seen before: only any other coefficient of the same variable.

### Justification

I don't think it's impossible to solve 2. without the change proposed here, possibly by refactoring how function calls are flattened, removing the assumption that variables to be renamed appear first as the left side of a statement. Basically walking the AST in a different way.

I however can't see how to solve 1. without the proposed change, as the binary aspect of the variables need to somehow be encoded in any new names to be applied.

### Risk

What if a variable has multiple `_b`? Is it even possible?
How do be confident our HashMap wrapper is safe?


Schaeff:
Closing in favor of #45

PR #57, opened 2018-04-27
Add sha256 from libsnark

barryWhiteHat:
This PR adds sha256 witness generation, constraint generation directly from libsnark. Thanks to @kobigurk for help with this.

It also includes wraplibsnark.cpp upgrade to support the latest version of libsnark. 

This brakes ZoKrates as `exportInput` , `_setup` and `_generate_proof` functions were replaced with place holders as i could not see easily how to port this code. 

The build process is changed slightly so remember to run `git submodule update --init --recursive` before `cargo +nightly build`

This change breaks the docker setup. 

The thinking is that @JacobEberhardt will take over from here and use the `_sha256Constraints` and `_sha256Witness` functions i provide to integrate sha256 in ZoKrates at a cost of ~27k constraints. Which is much better than our native implementation ~500k constraints. Which will hopefully make merkle trees practical in ZoKrates.


christianlundkvist:
Hey @barryWhiteHat, this is very cool! You mentioned “our native implementation” of sha256. Is this an implementation written in the ZoKrates DSL? Do you have a link to the code?

kobigurk:
@christianlundkvist - I believe he means @Schaeff's implementation here: https://github.com/Schaeff/ZoKrates/tree/sha-256/examples/sha256

that said, unless we're testing it incorrectly - it seems that it passes the test for an input of all 0s, but not for others - would probably benefit from another eye :-)

christianlundkvist:
@kobigurk Ah, thanks for the link! Ok this is very interesting, would be cool to try this out in practice! I guess from what you're saying we need to test it a bit more to make sure it works. For actual use in Solidity we would probably also need to create a wrapper for the sha256compression that takes two uint256 values rather than 512 uint256 values :)

lib/wraplibsnarkgadgets.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include <libsnark/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.hpp>
+
+
+using namespace libsnark;
+using namespace libff;
+
+using std::vector;
+
+
+template<typename FieldT>
```
xwvvvvwx: Do these functions need to be templated? 

I'm not sure how to easily call a templated C++ function from Rust (Rust only speaks C not C++) and if I'm reading the code right it's not required (there is a `typedef` for `FieldT` in [`sha256_ethereum`](https://github.com/JacobEberhardt/ZoKrates/pull/57/files#diff-ab7ad5652f9ddd2e8337327a30127bc0R22)).

lib/wraplibsnarkgadgets.cpp
```diff
@@ -0,0 +1,139 @@
...
+            ss << ", ";
+        }
+    }
+    ss << "],\n";
+    ss << "\"constraints\":[";
+     
+    for (size_t c = 0; c < constraints.num_constraints(); ++c)
+    {   
+        ss << "[";// << "\"A\"=";
+        constraint_to_json(constraints.constraints[c].a, ss);
```
xwvvvvwx: This function does not seem to be defined anywhere, did you forget to include it in the PR?

lib/wraplibsnarkgadgets.cpp
```diff
@@ -0,0 +1,139 @@
...
+    return ss.str(); 
+}
+
+template<typename FieldT>
+std::string _sha256Constraints() {
+
+    protoboard<FieldT> pb;
+    std::shared_ptr<sha256_ethereum> hash;
+
+    block_variable<FieldT> input;
```
xwvvvvwx: Unless I'm missing something`block_variable` does not seem to have a default constructor (see [here](https://github.com/scipr-lab/libsnark/blob/92a80f74727091fdc40e6021dc42e9f6b67d5176/libsnark/gadgetlib1/gadgets/hashes/hash_io.hpp#L38)).

I think we need to provide some initial values here.

xwvvvvwx:
Hey @barryWhiteHat really cool PR 👌🙂 

I've been playing around on a branch trying to get this bound into rust ([here](https://github.com/JacobEberhardt/ZoKrates/compare/develop...xwvvvvwx:bind-libsnark-sha256)), but I've run into a few compilation issues.

I've left some comments inline, maybe you could clarify a few questions?

Schaeff:
We're actively working on this one! Stay tuned

Schaeff:
Closing for #66, thanks! @christianlundkvist I'd love to hear your thoughts on the interface with Solidity, we're thinking about how to add types to ZoKrates itself (rather than only field elements)

PR #61, opened 2018-05-15
Fixed power expression with a number base in flattener

lubieplacki:
I've found out that power expressions with a number base weren't flattened properly - it was just a single multiplication before.
So before the fix:
`x = 5**5`
was returning:
`x = 5*5`
Now it works correctly.

I guess my editor removed some whitespaces from the end of the lines. :)


Schaeff:
Good catch @lubieplacki !

Schaeff:
Better late than never :) @lubieplacki I made some changes but thanks for giving the direction 🏅

PR #66, opened 2018-06-05
Add integration with sha256 from libsnark

Schaeff:
[WIP]

Inject a function named `sha256libsnark` in ZoKrates programs, available globally, which computes sha256 on its input using a libsnark gadget.

Also introduces the latest version of libsnark with an updated build process

Todo:
- [x] Make sure the generated witness satisfies the constraints
- [ ] Enable passing arguments in (currently `sha256(0...0)`)
- [ ] Merge #67
- [ ] Add Error handling to witness generation

Special thanks to @barryWhiteHat and @xwvvvvwx !


Kyroy:
I think this PR needs a new Dockerfile in the CI. Let me know when you think you are done. I will adapt the Dockerfile :)

Schaeff:
Hey @Kyroy I think the Dockerfile is pretty much stable now!

Dockerfile
```diff
@@ -2,33 +2,34 @@ FROM ubuntu:14.04
...
+    sh -s -- --default-toolchain $rust_toolchain -y
 
 ENV PATH=/root/.cargo/bin:$PATH
 
-RUN cd libsnark-$libsnarkcommit \
-  && make install lib PREFIX=/usr/local \
-    NO_PROCPS=1 NO_GTEST=1 NO_DOCS=1 CURVE=ALT_BN128 FEATUREFLAGS="-DBINARY_OUTPUT=1 -DMONTGOMERY_OUTPUT=1 -DNO_PT_COMPRESSION=1"
-
-ENV LD_LIBRARY_PATH $LD_LIBRARY_PATH:/usr/local/lib
+RUN git clone https://github.com/scipr-lab/libsnark.git $LIBSNARK_PATH
```
Kyroy: Using the latest libsnark commit may break the Dockerfile and/or ZoKrates

build.rs
```diff
@@ -1,23 +1,52 @@
+extern crate cmake;
 #[cfg(not(feature = "nolibsnark"))]
 extern crate gcc;
 
 fn main() {
     #[cfg(not(feature = "nolibsnark"))]
     {
+        let libsnark = cmake::Config::new("/root/libsnark")
```
Kyroy: I will also not work on the CI pipeline until this is fixed ;)

Dockerfile
Schaeff: Thanks, I'll add a specific commit

build.rs
Schaeff: Yes, in progress here #67

Dockerfile
Kyroy: Why not include libsnark as submodule in this repo?

Schaeff: I think it should be an integration rather than a dependency. @JacobEberhardt thoughts?
Also I think submodules are a bit awkward, and I don't think we need real time updates to libsnark.

JacobEberhardt: I 100% agree!
Submodules should not be employed here. As I understand, it was a temporary fix to move to the latest libsnark version employing cmake.

src/main.rs
```diff
@@ -336,7 +345,7 @@ fn main() {
 
             let witness_map = main_flattened.get_witness(arguments);
             // let witness_map: HashMap<String, FieldPrime> = main_flattened.get_witness(args);
```
xwvvvvwx: do you want to keep this commented out code?

Schaeff:
@Kyroy this still had a bunch of todos, don't think it should have been merged so early?

Kyroy:
I am not sure what happened. But I did not merge this. I just merged #68 which included A LOT of commits.

Kyroy:
I think your branches build on each other or something like this

Schaeff:
OK that's possible. I guess I'll resolve the conflicts and merge again

PR #74, opened 2018-06-20
Fixed less-than function call bug

lubieplacki:
There was a bug with less-than operator #43  - it crashed during witness generation when used in function call. 

I've figured a bug when in result code I've found:
`naturalCheck_i1o1_1_sym_2_b0 = (sym_2_b0 * sym_2_b0)`
which obviously should look like 
`naturalCheck_i1o1_1_sym_2_b0 = (naturalCheck_i1o1_1_sym_2_b0 * naturalCheck_i1o1_1_sym_2_b0)`
so function name wasn't added to sym_2_b0.

So I've found out that replacement_map was updated with naturalCheck_i1o1_1_sym_2_b0 after applying substitution to rhs. (line 278 and 286)
Also there was more bits used than defined so I fixed that as well. (line 96)

### DIFF
--- a/src/flatten.rs
+++ b/src/flatten.rs
@@ -93,7 +93,7 @@ impl Flattener {
                         box FlatExpression::Mult(box FlatExpression::Number(T::from(2)), box FlatExpression::Identifier(rhs_name.to_string())),
                     ),
                 ));
-                for i in 0..self.bits - 2 {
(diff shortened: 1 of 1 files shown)


Schaeff:
@JacobEberhardt is that the fix you were planning to make? Is it correct?

PR #75, opened 2018-06-22
Make helpers generic

Schaeff:
Support libsnark and rust, allowing to add more gadgets and other util


Schaeff:
Just a refactor, will merge this.

PR #80, opened 2018-07-04
Implement groth16

Schaeff:


Schaeff:
jenkins-test

Schaeff:
@Kyroy I think this PR would require updating CI as the libsnark commit and repo changed. I'm wondering if this could be captured in the repo?

Kyroy:
I'd also like to use the Dockerfile from the repo. But since it uses the root user, I cannot use it to run the tests.

But I think we can easily change this. I will make a PR to your branch if you like.

Kyroy:
Done 🎉

Schaeff:
@Kyroy what would you think about submitting a different PR for this? I'm realising it will be quite useful for another bit I'm working on (so that I can update the Dockerfile to pass CI) and this branch needs a security review before merge.

Kyroy:
I agree #81

HarryR:
I am concerned about the pairing operation, and malleability of proofs.

```solidity
        if(!Pairing.pairingProd4(
             proof.A, proof.B,
             Pairing.negate(vk_x), vk.gamma,
             Pairing.negate(proof.C), vk.delta,
             Pairing.negate(vk.a), vk.b)) return 1;
```

I'm struggling to fully comprehend much of the notation in Groth16 paper, however in https://eprint.iacr.org/2016/260.pdf the validity check is something like (pg14):

> From the Abstract: In our SNARK we work with asymmetric pairings for higher efficiency, a proof is only 3 group elements, and verification consists of checking a single pairing product equations using 3 pairings in total

> Compute a quadratic multi-variate polynomial t such that `t(σ,π) = 0` corresponds to the test

Where `σ` is the Proving key and `τ` is the verifying key of `alpha`, `beta`, `gamma`, `delta`, `x`

    A * B = alpha * beta + ( Z[....] * gamma) + (C * delta)

So

    e(A, B) = e(alpha, beta) * e(Z[...], gamma) * e(C * delta)

Then in your `verify` you're doing:

    1 == e(A,B) * e(-VK_x, gamma) * e(-C, delta) * e(-alpha, beta)

Where `Z[...]` is `VK_x`. From the following right?

```solidity
       Pairing.G1Point memory vk_x = Pairing.G1Point(0, 0);
        for (uint i = 0; i < input.length; i++)
            vk_x = Pairing.addition(vk_x, Pairing.scalar_mul(vk.gammaABC[i + 1], input[i]));
        vk_x = Pairing.addition(vk_x, vk.gammaABC[0]);
```

I am having a hard time following the notation for the sum in `Vfy()` on §3.1 (pg14) and verifying that it matches `vk_x`.

----------------

Anyway, the version in the paper is somewhat like

    A + B = (alpha + beta) + (vk_x + gamma) + (C + delta)

Where, because of the pairing check constraint in EVM, where product equals 1:

    (gamma - vk_x) + (delta - C) + (beta - alpha) + A + B = 1

I am still digging into the statement, and I will get back to you with further comments, however if you have any comments, corrections or insight in the meantime I'd appreciate it.

HarryR:
So, I can negate any one statement in the three pairings using the variables A, B and C, that is - I can pair an item with its self so the product = `FQ12.one`. However, I'm still working on two statements where I have to do jedi crypto trix with `C` and `B`, and three is looking highly unlikely.

Cursory glances say the product to one (or... sum to zero) statement you've devised is correct and not malleable, however... I'm still learning about the intricacies of pairings, so don't take my word on it.

Schaeff:
Hi Harry, thanks for your feedback, we'll look into it. What I can say for now is:
- The solidity implementation here hasn't been extensively tested as we first want to make sure that not precomputing the GT element `e(alpha,beta)` doesn't have any unintended consequences
- We're aware of the malleability of Groth16 as it was pointed out here https://eprint.iacr.org/2017/540.pdf

Schaeff:
Closing as we used bellman for this!

PR #81, opened 2018-07-06
Use Dockerfile for CI tests

Kyroy:
Different branches with changes on the libraries or Dockerfile will no longer require a change of an extra Dockerfile for CI.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,9 +3,16 @@
 def majorVersion
 def minorVersion
 def patchVersion
+def dockerImage
 
 pipeline {
     agent any
+    options {
+        skipStagesAfterUnstable()
+        timeout(time: 2, unit: 'HOURS')
+        timestamps()
+        buildDiscarder(logRotator(numToKeepStr: '10', artifactNumToKeepStr: '10'))
+    }
     stages {
         stage('Init') {
             steps {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -3,9 +3,16 @@
 def majorVersion
 def minorVersion
 def patchVersion
+def dockerImage
 
 pipeline {
     agent any
+    options {
+        skipStagesAfterUnstable()
```
Schaeff: Just curious what these do? Could you add a comment?

Jenkinsfile
```diff
@@ -25,16 +32,25 @@ pipeline {
         }
         stage('Build') {
             steps {
-                withDockerContainer('kyroy/zokrates-test:1') {
-                    sh 'RUSTFLAGS="-D warnings" cargo build'
+                script {
+                    ansiColor('xterm') {
+                        dockerImage = docker.build("kyroy/zokrates")
```
Schaeff: Would it make sense to have a Zokrates dockerhub account?

Jenkinsfile
Kyroy: Normally, if tests fail, the build should be marked UNSTABLE, not FAILED. Then this option prevents the pipeline to continue other stages.

https://jenkins.io/doc/book/pipeline/syntax/

Jenkinsfile
```diff
@@ -56,14 +76,15 @@ pipeline {
...
             steps {
                 script {
-                    def dockerImage = docker.build("kyroy/zokrates")
-                    docker.withRegistry('https://registry.hub.docker.com', 'dockerhub-kyroy') {
-                        dockerImage.push(patchVersion)
-                        dockerImage.push(minorVersion)
-                        if (majorVersion > '0') {
-                            dockerImage.push(majorVersion)
+                    ansiColor('xterm') {
+                        docker.withRegistry('https://registry.hub.docker.com', 'dockerhub-kyroy') {
```
Schaeff: Oops @Kyroy missed this, isn't it supposed to push to zokrates on dockerhub?

Schaeff: Oh ok great!

PR #86, opened 2018-07-15
Add basic types to the semantics and flattening

Schaeff:
Add a type system with two types:
- FieldElements
- Booleans

FieldElements are used for everything we currently have without types, and booleans are used for IfElse statements.

No changes in the parser, this will be part of a separate PR.

Todo:

- [x] unpack in flattener
- [x] clean up
- [x] figure out how to treat imports
- [x] remove compiled output from integration tests


Schaeff:
Hey @Kyroy seems like CI is down?

src/types/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+}
+
+impl Type {
+	fn get_constraints<T: Field>(&self) -> Constraints<T> {
+		match self {
+			Type::FieldElement => Constraints::none(),
+			Type::Boolean => Constraints::boolean(),
+		}
+	}
+
```
JacobEberhardt: Maybe add comment:
# of field elements the type maps to.

Kyroy:
@Schaeff Yes it was down. Thanks for the info

src/types/mod.rs
Schaeff: Done

PR #87, opened 2018-07-16
Fix Example in README

3psil0n:
The example should be a computation that returns a+b
Otherweise, the later witness does not make sense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ cd ZoKrates/target/release
 To execute the program, perform the setup for the program, generate a proof
 ```
 def add(a, b, c):
-  return a + b + c
+  return a + b
 ```
 with `add(1, 2, 3)`, call
 ```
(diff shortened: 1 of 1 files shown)


Schaeff:
`compute-witness` finds the witness for you based on the inputs, so in your example it would be `compute-witness -a 1 2`

Schaeff:
Hey @3psil0n 

> The example method has 3 arguments, namely a b c. This would imply that a valid witness takes 4 arguments. However, in the example, the compute-witness function just takes 3 arguments.

ZoKrates finds the full witness based on the arguments of the main function, so compute-witness takes 3 arguments, not 4 (the 4th you're thinking of is the result, and this is generated by ZoKrates)

I agree however that we should have the same example all the way through so if you can make it that way I'll be happy to merge this.

For example:

def main(a, b):
   return a + b

main(1, 2)

compute-witness -a 1 2

PR #91, opened 2018-07-29
Read Solidity template from file given as [--template] argument

naure:


Voxelot:
Would be really nice to have this since the solidity output doesn't play well with projects that have multiple verifiers and is also pretty dated (^4.19.0)

Schaeff:
Hey @Voxelot the verifier will be updated to 0.5 in the next release. For custom verifiers, for now we advise users to create them based on the `verification.key` file.

Voxelot:
Sounds good, and yeah the verification.key file is easy enough to parse. JSON formatting or some other standard serialization format would be nice though.

PR #92, opened 2018-07-29
Output proof as YAML to be consumed by scripts

naure:


Schaeff:
Thanks @naure! @JacobEberhardt any thoughts on yaml? Also, maybe we want to have this on the Rust side as it's not libsnark specific?

JacobEberhardt:
Nice feature, should be implemented in Rust to merge, though. We do not want to have more cpp than necessary.

PR #93, opened 2018-07-29
Support for libsnark gadgets through a C interface

naure:


JacobEberhardt:
Since there is now a PR integrating the zkInterface, this is outdated.

PR #108, opened 2018-08-14
Implement u8 type

Schaeff:
Experimental branch for now.
Type inference is used as the parser doesn't have type annotations.


Schaeff:
This is super outdated. Good to keep in mind as we'll soon have better tools to make it happen.

PR #130, opened 2018-09-20
Remove typed substitutions and linear checks

Schaeff:
We currently apply substitutions both on the typed AST and flat AST during flattening. There's no reason this should be, and substituting only on flat AST would be simpler and cleaner.

To merge before #131


zokrates_core/src/flatten/mod.rs
```diff
@@ -402,54 +403,29 @@ impl Flattener {
...
-                                        arguments_flattened,
-                                        statements_flattened,
-                                        expr_subbed,
-                                    );
-                                    new_var = format!("{}param_{}", &prefix, i);
-                                    statements_flattened
-                                        .push(FlatStatement::Definition(new_var.clone(), rhs));
-                                }
-                            }
+                            // for field elements, flatten the input and assign it to a new variable
```
Schaeff: @JacobEberhardt just wanna make sure I'm not missing something here, any thoughts on why matching Identifier separately above was necessary?

JacobEberhardt: Not necessary, new version is fine.

PR #131, opened 2018-09-24
Remove Strings from the flat absy

Schaeff:
The current core of the compiler uses string substitution to flatten constructs. This has a considerable cost especially during function call resolution.

Introduce `FlatVariable` which is cheap and `Copy`, and keep track of bindings between non-flattened variable names and `FlatVariable`s in a bijection table.

Other changes introduced:
- Get rid of substring-based witness generation (x_b5 being interpreted as the 6th bit or x at run time) and instead use explicit Directives
- Remove loops from substitution calls: we only need to know the latest `FlatVariable` for a given `Variable`, not each intermediate one

To merge after #130


JacobEberhardt:
> jenkins-test

PR #133, opened 2018-10-04
Implement Boolean Operators

tjade273:
Implements operators on boolean conditions in if statements.

# Done:
* Lexing, parsing and flattening boolean and (&&)

# Todo:
* Boolean or, negation, xor
* Boolean operators in assignment/arbitrary expressions
* Parenthesized boolean expressions


zokrates_core/src/typed_absy/mod.rs
```diff
@@ -298,6 +298,7 @@ pub enum BooleanExpression<T: Field> {
     Eq(Box<FieldElementExpression<T>>, Box<FieldElementExpression<T>>),
     Ge(Box<FieldElementExpression<T>>, Box<FieldElementExpression<T>>),
     Gt(Box<FieldElementExpression<T>>, Box<FieldElementExpression<T>>),
+    AndAnd(Box<BooleanExpression<T>>, Box<BooleanExpression<T>>),
```
Schaeff: Looks great so far! Just wondering if we could call this `And` at this point as we're referring to the boolean And operation and not the `&&` character sequence.

zokrates_core/src/flatten/mod.rs
```diff
@@ -349,12 +349,33 @@ impl Flattener {
...
+                let y = self.flatten_boolean_expression(
+                    functions_flattened,
+                    arguments_flattened,
+                    statements_flattened,
+                    rhs
+                );
+                let name_new = format!("sym_{}", self.next_var_idx);
+
+                statements_flattened.push(FlatStatement::Definition(
+                    name_new.to_string(), FlatExpression::Mult(box x, box y))
```
Schaeff: One thing to be careful about is that `FlatExpression`s are not necessarily linear, so multiplying two of them is not guaranteed to be quadratic. This is a bit awkward and may change in the future (using linear expressions as the main intermediate concept rather than quadratic now) but for now using `.is_linear()` is sometimes necessary to make sure we always end up with something quadratic in the flat program. Check out the flattening of `Mult` in `flatten_field_ewpression` for an approach, and feel free to add `assert(x.is_linear() && y.is_linear())` before pushing the definition

zokrates_core/src/typed_absy/mod.rs
tjade273: That's fair, I chose the name AndAnd to distinguish boolean and (&&) from bitwise and (&), which might implemented in the future. Maybe BoolAnd and BitAnd would be clearer.

zokrates_core/src/flatten/mod.rs
tjade273: Since x and y are flattened boolean expressions, and `faltten_boolean_expression` always returns an identifier, I believe they should always be linear. I've added an assertion to document that

zokrates_core/src/typed_absy/mod.rs
Schaeff: Oh I see, however in this context isn't it already clear we're talking about booleans as we're defining BooleanExpressions?

zokrates_core/src/flatten/mod.rs
Schaeff: Great thank you, good to have for when `flatten_boolean_expression` starts returning something else.

zokrates_core/src/flatten/mod.rs
```diff
@@ -357,12 +357,35 @@ impl Flattener {
...
+                    functions_flattened,
+                    arguments_flattened,
+                    statements_flattened,
+                    rhs
+                );
+                let name_x_and_y = self.use_sym();
+
+                // flatten_boolean_expression always returns a linear term
+                assert!(x.is_linear() && y.is_linear());
+                statements_flattened.push(FlatStatement::Definition(
```
Schaeff: This new definition is not needed in theory as x*y is guaranteed to be quadratic, but I see how it makes sure the return is an identifier and we can compose boolean expressions. Maybe we could have a comment to that effect?

zokrates_core/src/parser/tokenize/tokenizer.rs
```diff
@@ -201,6 +201,24 @@ pub fn next_token<T: Field>(input: &String, pos: &Position) -> (Token<T>, String
...
+        Some('&') => match input.chars().nth(offset + 1) {
+            Some('&') => (
+                Token::AndAnd,
+                input[offset + 2..].to_string(),
+                Position {
+                    line: pos.line,
+                    col: pos.col + offset + 2,
+                },
+            ),
+            _ => (
```
BrianGuo: Why do we have logic to tokenize bitwise And in this PR, but no further compiling? I'm assuming it's to make this pattern matching exhaustive, but I think we might be better off throwing an error than writing a half-implemented feature. It would make more sense to have all the logic for bitwise And implemented together in a future PR

zokrates_core/src/typed_absy/mod.rs
BrianGuo: I agree with renaming AndAnd here to just And - the BooleanExpression parameters are enough to distinguish it, and Eq follows the same naming convention, where the token is lexed as EqEq but written as Eq in this enum.

BrianGuo: Made the change - andand is now and

zokrates_core/src/parser/tokenize/tokenizer.rs
Schaeff: Good catch! I agree we should get rid of `&` for now

Schaeff:
Hey @BrianGuo you mentioned this should be close to mergeable, sweet! I guess the few outstanding comments and change requests should be updated, as well as the description that still has a TODO list: would be nice to have a summary of the changes there even if it's not the whole thing as I understand you're working on more. Let me know how I can help

zokrates_core/src/parser/tokenize/tokenizer.rs
BrianGuo: This was fixed and there's no and operator now

tjade273:
@Schaeff This is should be mergeable now, the major difference is that we now parse boolean expressions in a similar LL style to the rest of the parser, which allows more natural associativity and parentheses. Also boolean negation is implemented.

zokrates_core/src/parser/tokenize/tokenizer.rs
```diff
@@ -225,6 +225,28 @@ pub fn next_token<T: Field>(input: &String, pos: &Position) -> (Token<T>, String
...
+        Some('|') => match input.chars().nth(offset + 1) {
+            Some('|') => (
+                Token::Or,
+                input[offset + 2..].to_string(),
+                Position {
+                    line: pos.line,
+                    col: pos.col + offset + 2
+                },
+            ),
+            _ => panic!("| should not appear alone"),
```
Schaeff: Here I think we should use 
```
Token::Unknown(String::from('|')),
                 input[offset + 1..].to_string(),
                 Position {
                     line: pos.line,
                     col: pos.col + offset + 1,
                 },
```

to get a better error

zokrates_core/src/parser/tokenize/tokenizer.rs
```diff
@@ -225,6 +225,28 @@ pub fn next_token<T: Field>(input: &String, pos: &Position) -> (Token<T>, String
...
+        Some('&') => match input.chars().nth(offset + 1) {
+            Some('&') => (
+                Token::And,
+                input[offset + 2..].to_string(),
+                Position {
+                    line: pos.line,
+                    col: pos.col + offset + 2,
+                },
+            ),
+            _ => panic!("Bitwise AND (&) is unimplemented. Did you mean &&?"),
```
Schaeff: same, see below

zokrates_core/src/typed_absy/folder.rs
```diff
@@ -177,6 +177,20 @@ pub fn fold_boolean_expression<T: Field, F: Folder<T>>(f: &mut F, e: BooleanExpr
...
+            let e2 = f.fold_boolean_expression(e2);
+            BooleanExpression::Or(box e1, box e2)
+        }
+        BooleanExpression::And(box e1, box e2) => {
+            let e1 = f.fold_boolean_expression(e1);
+            let e2 = f.fold_boolean_expression(e2);
+            BooleanExpression::And(box e1, box e2)
+        }
+        BooleanExpression::Not(box e) => {
+            let e1 = f.fold_boolean_expression(e);
```
Schaeff: rename to e to follow same pattern

zokrates_core/src/parser/parse/expression.rs
```diff
@@ -42,39 +42,168 @@ fn parse_then_else<T: Field>(
     }
 }
 
-fn parse_if_then_else<T: Field>(
+fn parse_prim_cond<T: Field>(
```
Schaeff: function name could be clearer

zokrates_core/src/typed_absy/folder.rs
Schaeff: fixed

zokrates_core/src/parser/tokenize/tokenizer.rs
Schaeff: fixed

zokrates_core/src/parser/tokenize/tokenizer.rs
Schaeff: fixed

PR #136, opened 2018-10-09
Add Ethereum SHA-256 gadget

sanchopansa:
@stefandeml and myself have been working on getting an Ethereum-compliant SHA-256 gadget based on the libsnark code from @kobigurk: https://github.com/kobigurk/sha256_ethereum.

You can find the changes below.


zokrates_cli/tests/code/sha_ethereum.code
```diff
@@ -0,0 +1,21 @@
...
...d, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field):
```
Schaeff: As these were added to the tests (great), we need an `.arguments.json` file and an `.expected.witness` file to be able to run integration tests. Also `a` is not being used so could be removed.

zokrates_cli/tests/code/sha_libsnark.code
```diff
@@ -1,3 +1,5 @@
...
...d, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field, field):
```
Schaeff: Same here

zokrates_core/lib/wraplibsnarkgadgets.cpp
```diff
@@ -1,20 +1,182 @@
...
-
+//https://gist.github.com/kobigurk/24c25e68219df87c348f1a78db51bb52	
 #include <iostream>
 
+#include "wraplibsnarkgadgets.hpp"
+
 #include "libsnark/gadgetlib1/gadget.hpp"
+#include "libsnark/gadgetlib1/protoboard.hpp"
 #include "libff/common/default_types/ec_pp.hpp"
+#include "libsnark/reductions/r1cs_to_qap/r1cs_to_qap.hpp"
```
Schaeff: Is this needed if we only deal with r1cs?

zokrates_core/src/helpers/libsnark_gadget.rs
```diff
@@ -8,12 +8,14 @@ use helpers::{Signed, Executable};
...
 pub enum LibsnarkGadgetHelper {
 	Sha256Compress,
+	Sha256Ethereum,
 }
 
 impl fmt::Display for LibsnarkGadgetHelper {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
     	match *self {
     		LibsnarkGadgetHelper::Sha256Compress => write!(f, "Sha256Compress"),
+			LibsnarkGadgetHelper::Sha256Ethereum => write!(f, "Sha256Ethereum"),
```
Schaeff: Indentation looks off

zokrates_core/src/helpers/libsnark_gadget.rs
```diff
@@ -28,6 +30,15 @@ impl<T: Field> Executable<T> for LibsnarkGadgetHelper {
 					return Err(format!("{}", e));
 				}
 
+				Ok(witness_result.unwrap().variables.iter().map(|&i| T::from(i)).collect())
```
Schaeff: Instead of duplicating the whole case, could we have only the minimum in there and the rest after the match as it's common?

zokrates_core/src/standard.rs
```diff
@@ -37,6 +37,11 @@ pub struct Constraint {
 	c: BTreeMap<usize, String>,
 }
 
+pub struct DirectiveR1CS {
+    pub r1cs : R1CS,
+    pub directive : Option<LibsnarkGadgetHelper>
```
Schaeff: Is there any case where we wouldn't want a directive? Wouldn't it end being a bunch of constraints the wires of which are not assigned?

sanchopansa: Correct me if I'm wrong, but I believe the directive is only necessary when we have a computation done off-circuit, which is then verified on-circuit? So it would be the case that for some gadgets this is not necessary.

sanchopansa:
Thanks for the review @Schaeff! Most comments are addressed, we'll add the `arguments.json` and `expected.witness` files shortly.

zokrates_core/src/standard.rs
stefandeml: I guess @Schaeff is right. `R1CS` is only being used if libsnark is involved, hence there also will be usage of some directive / helper.
`None` is currently used for example in the following test:
https://github.com/JacobEberhardt/ZoKrates/pull/136/files#diff-e4af2ccde827780269c4d04b27eef2ccR262
I guess this should be changed as well.

My main reasoning was to only have on conversion into `FlatFunction`.

Schaeff: Yep @sanchopansa as this yields Constraints, the variables involved need to have been defined before either in a Definition or a Directive, so I can't see the case for not having a Directive here. Adjusting the test by specifying the sha256 directive instead of None would work, and hopefully we can keep directives and constraints at the same place in the future

zokrates_cli/tests/code/sha_libsnark.code
stefandeml: Maybe I'm wrong, imo the `integration.rs` has the flag `-a` hardcoded. Hence the integration tests will fail if you run an example without argument. 
I added dummy arguments on both examples and all tests pass.

zokrates_core/src/standard.rs
stefandeml: Changed as discussed.

stefandeml:
Thanks for the review @Schaeff!

zokrates_cli/tests/code/sha_libsnark.code
Schaeff: You're right my bad

PR #146, opened 2018-10-25
Change sha256packed to take 4 parameters

Schaeff:
Instead of taking 2 values of 254 bits, take 4 values of 128 bits. Also return 2 values of 128 bits.


JacobEberhardt:
CI was down. Trigger rebuild.

Dockerfile.prod
```diff
@@ -0,0 +1,43 @@
+FROM ubuntu:18.04
+
+MAINTAINER JacobEberhardt <jacob.eberhardt@tu-berlin.de>, Dennis Kuhnert <mail@kyroy.com>
```
JacobEberhardt: Could add you there @Schaeff

Schaeff: Oops this file was not meant to be commited, I'll remove it

PR #147, opened 2018-10-25
Add production Dockerfile

Schaeff:
Since we moved to the new build process for libsnark, the size of our Docker images exploded.
It's hard to keep the same Dockerfile for both development and usage, as users do not need Rust, or the source code, or libsnark: they only need the ZoKrates binary.

This PR introduces a "production" Dockerfile which is lighter than the dev one thanks to:
- removal of rust
- removal of superfluous dependencies: mostly `libboost-dev-all` where `libboost-dev` and `libboost-program-options-dev` seem to be enough for what we're doing
- removal of source code
- keeping examples
- ubuntu 18:04

Locally, the size of the image as measured by `docker images` is 632mb (for context, libsnark-only images are around 420mb), down from 2.2gb (!)

Feedback would be much appreciated as this is not my domain of expertise @kyroy @JacobEberhardt @stefandeml :)

Todo:
- [ ] documentation

### DIFF
--- a/dev.Dockerfile
+++ b/dev.Dockerfile
@@ -1,0 +1,43 @@
+FROM ubuntu:14.04
+
+MAINTAINER JacobEberhardt <jacob.eberhardt@tu-berlin.de>, Dennis Kuhnert <mail@kyroy.com>
+
+RUN useradd -u 1000 -m zokrates
+
+ARG RUST_TOOLCHAIN=nightly-2018-06-04
+ARG LIBSNARK_COMMIT=f7c87b88744ecfd008126d415494d9b34c4c1b20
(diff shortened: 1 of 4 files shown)


JacobEberhardt:
- Tested build. Works fine.
- Examples also work fine.
- Cargo tests unavailable, since src is unavailable. However, any problems should be caught by CI.
- CI build currently failing, hence no test of Dockerhub upload yet.

dev.Dockerfile
```diff
@@ -0,0 +1,43 @@
+FROM ubuntu:14.04
```
JacobEberhardt: We should probably be on one version. Not urgent.

JacobEberhardt:
Moved to Org Repo.

PR #152, opened 2018-10-31
Implement Le, Gt, Ge

tjade273:
Currently the operations

    <=, >=, >
are not flattened and fail with "unimplemented". 

This PR adds these cases to make `flatten_boolean_expr` exhaustive and adds tests for these operators.


zokrates_core/src/flatten/mod.rs
```diff
@@ -355,13 +355,31 @@ impl Flattener {
 
                 FlatExpression::Identifier(name_1_y)
             },
+            BooleanExpression::Le(box lhs, box rhs) => {
+                let lt = self.flatten_boolean_expression(functions_flattened,
```
Schaeff: Wouldn't it be possible to flatten `Or(Lt, Eq)` instead of flattening `Lt` and `Eq` and then adding them?

tjade273: Yes, I just did this so it would be independent from the boolean operator branch, but since that's merged we can use that.

It is actually more efficient to do it this way, since boolean Or uses a multiplication gate but this is guaranteed to be boolean with just an addition.

tjade273:
This branch seems to keep failing in Jenkins because of an email error? Whatever is happening, it's happening after the integration tests so I'm not sure how to fix it.

zokrates_core/src/flatten/mod.rs
Schaeff: Ok let's keep it like this for now

Schaeff: I think it would be clearer to use `Or` in this case. It's a minor performance hit, and we can potentially make `Or` flattening better if needed.

Schaeff:
@tjade273 I think I know the cause of the build fail here: the current flattening process expects only additions, no subtractions. This is being addressed by #180 but for the time being introducing an intermediate variable would fix it.

PR #166, opened 2018-11-13
Implement GM17 scheme

Schaeff:
GM17 has smaller proofs and faster verification. See the paper [here](https://eprint.iacr.org/2017/540.pdf) and a comparison [here](https://github.com/scipr-lab/libsnark/blob/master/libsnark/zk_proof_systems/ppzksnark/README.md)

Make the scheme available for
- [x] setup
- [x] generate-proof
- [x] export verifier
- [x] tests


Schaeff:
The current issue in this PR is that addition in G2 is not available as a precompile, so it has to be implemented in Solidity, possibly using G1 precompiles.

Schaeff:
Fixed using https://github.com/musalbas/solidity-BN256G2

Issue #167, opened 2018-11-19
associativity of subtraction of field values with function calls is wrong


aspiers:
This is best demonstrated by a test case:

```
def x() -> (field):
  return 3

def main() -> (field):
  // These all work
  7 == 10 - x()
  8 == 10 + 1 - x()
  8 == 10 - 3 + 1
  8 == (10 - 3) + 1
  8 == (10 - x()) + 1
  8 == 10 - (x() - 1)

  // These are wrong
  8 == 10 - x() + 1
  6 == 10 - x() - 1

  return 1
```

All of these assertions should pass; however the final two fail:

```
setup successful: true
Computing witness for:
def main():
        7 == 7
        8 == 8
        8 == 8
        8 == 8
        8 == 8
        8 == 8
        8 == 6
        6 == 8
        return 1
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { message: "Condition not satisfied: 8 should equal 6" }', libcore/result.rs:945:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

It seems that `10 - x() + 1` is interpreted as `10 - (x() + 1)` rather than `(10 - x()) + 1`.  I guess it's a problem with the way associativity is handled in combination with subtraction.  Taking a wild guess, maybe it's an issue with `flatten_field_expression()`?  But I never looked at Zokrates before this weekend, and only looked at the implementation for 30 minutes or so, so take that guess with a large pinch of salt ;-)

This bug was found in partnership with @GuthL, so thanks to him for being a great partner in crime :-)

Schaeff:
Thanks for this, good spot. I'd suspect a problem in the parser, not the flattener, but not sure.

Schaeff:
@GuthL just wondering if you made a start on this?

GuthL:
I just had a quick look but nothing conclusive yet. What make you think it depends on the parser?

Schaeff:
I had a crack at it as it's not a trivial piece of our codebase, maybe not the best to get started..

GuthL:
FIne with it! Could you explain how the bug came to be?

Schaeff:
@GuthL Sure, I commented on the PR.

Schaeff:
0.3.2 is out with a fix

aspiers:
Awesome, thanks!  BTW for future issues you might enjoy this trick: https://help.github.com/articles/closing-issues-using-keywords/

PR #173, opened 2018-11-23
Fix terms precedence

Schaeff:
#167 

Todo
- [x] check if we can use `parse_identified1` in `parse_statement` to avoid code rewrite: we can, but it has significant impact on statement parsing, so not added to this PR, see https://github.com/Schaeff/ZoKrates/tree/use-parse-identified-everywhere

### DIFF
--- a/zokrates_core/src/parser/parse/expression.rs
+++ b/zokrates_core/src/parser/parse/expression.rs
@@ -240,6 +240,19 @@ fn parse_factor1<T: Field>(
     }
 }
 
+// parse an identifier or select or function call
+fn parse_identified1<T: Field>(
+    x: String,
+    input: String,
+    position: Position,
+) -> Result<(Expression<T>, String, Position), Error<T>> {
+    match next_token::<T>(&input, &position) {
+        (Token::Open, s1, p1) => parse_function_call(x, s1, p1),
+        (Token::LeftBracket, s1, p1) => parse_array_select(x, s1, p1),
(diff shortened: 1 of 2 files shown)


zokrates_core/src/parser/parse/expression.rs
```diff
@@ -469,13 +472,12 @@ pub fn parse_expr<T: Field>(
             },
             Err(err) => Err(err),
         },
-        (Token::Ide(x), s1, p1) => match next_token::<T>(&s1, &p1) {
-            (Token::Open, s2, p2) => parse_function_call(x, s2, p2),
-            (Token::LeftBracket, s2, p2) => parse_array_select(x, s2, p2),
-            _ => match parse_term1(Expression::Identifier(x), s1, p1) {
-                Ok((e2, s2, p2)) => parse_expr1(e2, s2, p2),
-                Err(err) => Err(err),
+        (Token::Ide(x), s1, p1) => match parse_identified1(x, s1, p1) {
```
Schaeff: @GuthL this is the main issue here imo, the one that cause the bug you witnessed: before this change in `parse_expr`, we only run `parse_term1`(which is "parse term starting with expression x") when x is an identifier. Otherwise (when x is a function call or an array access), it just returns x. Therefore we end up with `Sub(x, Sub(y, z)) == x - (y - z)` instead of the expected `Sub(Sub(x, y), z) == (x - y) - z`

PR #176, opened 2018-11-26
Refactor optimizer with Folder pattern

Schaeff:
Currently the optimizer is a single custom written pass on the flat AST.
Implement the Folder (pure Visitor) pattern on flat AST and port the current optimizer to two Folder steps.


Schaeff:
Closing as the plan is to use ZIR for the optimiser

PR #178, opened 2018-11-27
Remove flat division

Schaeff:
Our IR currently supports divisions. To move closer to R1CS, remove divisions and emulate them with multiplications.

Example: `a = b/c`
Currently flattens to 
```
a = b/c
```
We can do
```
# inv_c = Helper::Inverse(c) // return 1/c
c * inv_c == 1
a = b * inv_c
```

Or
```
# a = Helper::Div(b, c) // returns b/c
a * c == b
```

~Seems to me that the first version reads a bit better, but no strong opinion. @JacobEberhardt ?~
We take the latter approach.


JacobEberhardt:
I think the second option is cleaner. The first option exposes details that are not required. Saving a constraint here makes sense.

Issue #175, opened 2018-11-26
Base of an exponentiation should be anything


Schaeff:
We currently only allow the base of an exponentiation to be a number or identifier.
```
// NOT allowed
a = (2x + y)**3
```
This restriction was added as a static analyser step because of an issue with how complex bases were handled. Find the issue, fix it, and remove the static analyser step.

The flattening issue is the following:
```
a*2**3
```
is flattened to
```
(a*2)**3
```
but should be flattened to
```
a*(2**3)
```

dhruvluthra:
I've been exploring this, but I can't seem to identify what the problem is. 

`def main() -> (field): 
    field x = 1
    field y = 2
    field a = (2*x + y)**2
    return (if a == 15 then 1 else 0 fi)`

I wrote this basic code to test, and it seems to work properly when I remove the power check. It seems like in the above example the complex base is being handled properly, with the function return 0 when we check a==15 and 1 when a==16 (as it should). 

Is there something I am missing here?

Schaeff:
My bad @dhruvluthra I updated the description, also we added an example that the static analyser step should block until this is fixed, it's at zokrates_cli/examples/error/forbidden_power_base.code

Schaeff:
Hey @JacobEberhardt we got rid of this now woohoo

JacobEberhardt:
Actually we fixed precedence, but not the "only number or identifier as base" part. 
I'll address this.

PR #190, opened 2018-12-03
fixing issue with exponent order of operation

dhruvluthra:
In reference to:  https://github.com/Zokrates/ZoKrates/issues/175

It seems the main reason the order of operations was not being handled for exponents is because exponentiation was being treated the same as the add and subtract operations. 

Here, we move the handling of exponents from `parse_expr1` into `parse_term1`, which handles them a little bit differently. 

Still trying to understand fully why this fixes the problem.


JacobEberhardt:
Solved with the new parser

PR #199, opened 2018-12-08
add doc update to circleci job for base branch

stefandeml:


scripts/publish_book.sh
```diff
@@ -0,0 +1,14 @@
...
+apt-get update
+apt-get -qq install git python-minimal
+python --version
+cargo install mdbook
+cd zokrates_book && mdbook build
+git config --global user.email "stefan.deml+zokratesbot@decentriq.ch"
+git clone https://github.com/Zokrates/zokrates.github.io.git
+git clone https://github.com/davisp/ghp-import.git
+cd zokrates.github.io
+../ghp-import/ghp_import.py -n -p -f -m "Documentation upload" -b "master" -r https://zokratesbot:"$GH_TOKEN"@github.com/Zokrates/zokrates.github.io.git ../book
```
Schaeff: Shall we include the version number in the commit message?

scripts/publish_book.sh
```diff
@@ -0,0 +1,14 @@
+# Exit if any subcommand fails
+set -e
+
+apt-get update
+apt-get -qq install git python-minimal
+python --version
+cargo install mdbook
+cd zokrates_book && mdbook build
+git config --global user.email "stefan.deml+zokratesbot@decentriq.ch"
```
Schaeff: Is this required if we use a token?

stefandeml: yes, otherwise commits show up with "Unknown" as user.

scripts/publish_book.sh
stefandeml: added tags to the commit msg (including a catch for now in case no tag exists)

scripts/publish_book.sh
```diff
@@ -0,0 +1,18 @@
+# Exit if any subcommand fails
+set -e
+
+if [ "$CIRCLE_BRANCH" == "develop" ]; then
```
Schaeff: @stefandeml think we agreed for this to be master? Otherwise looks good!

stefandeml: As far as I see $CIRCLE_BRANCH is the source branch, hence we have to use "develop", no?

scripts/publish_book.sh
```diff
@@ -0,0 +1,22 @@
...
+    apt-get -qq install git python-minimal
+    python --version
+    cargo install mdbook
+    cd zokrates_book && mdbook build
+    git config --global user.email "stefan.deml+zokratesbot@decentriq.ch"
+    git clone https://github.com/Zokrates/zokrates.github.io.git
+    git clone https://github.com/davisp/ghp-import.git
+    cd zokrates.github.io
+    { 
+        TAG=$(git describe --tags) && echo $TAG
```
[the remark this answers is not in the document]
stefandeml: Fixed to use cargo version number

scripts/publish_book.sh
Schaeff: No I don't think so: after we've merged from a release branch to master, CircleCI builds master which should trigger deployment. Or am I missing something?

Schaeff:
```suggestion
if [ "$CIRCLE_BRANCH" == "master" ]; then
```

Issue #135, opened 2018-10-09
Add user-defined directives


Schaeff:
At a high level, it may be interesting for some users to write inline directives.
There is a security risk as the user has the responsibility to ensure the variables are constrained correctly, which we could make clear with `unsafe` blocks like Rust does

Thanks @jbaylina for the suggestion!

PR #203, opened 2018-12-12
Implement support for WASM helpers

gballet:
This commit provides the basic features to execute code. A replacement of `Identity` is provided for tessting.

Some points still need to be discussed:

 - [x]  ~~At the moment one can not return a pointer greater than `MAX_INT`. A potential idea would be to return a `i64` instead of a `i32`.~~ -> it has been decided to `panic` upon encountering an error.
 - [x] Decide how to activate the WASM stuff (see commented out `#[cfg(feature = "wasm")]` in code)
 - [x] Find the most practical way to implement `ConditionEq` and write it.

Moved to a subsequent PR:
 - Decide on the best interface for using WASM files, right now there is no caching.
 - Right now the amount of inputs and outputs is fixed, we need to discuss and find out if we want a function to determine how many outputs a given amount of inputs will provide.

EDIT: source code of the WASM binaries: https://gist.github.com/gballet/f14d11053d8f846bfbb3687581b0eecb


zokrates_core/src/helpers/wasm.rs
```diff
@@ -0,0 +1,385 @@
...
+		let outputs = id.execute(&input).expect("Identity call failed");
+		assert_eq!(outputs, input);
+	}
+
+	#[test]
+	fn check_memory_boundaries() {
+		/* Check that input writes are boundary-checked */
+		let id = WasmHelper::new_code("0061736d010000000105016000017f030302000005030100010615047f0041010b7f0041010b7f0041200b7f0141000b074b06066d656d6f727902000e6765745f696e707574735f6f6666000105736f6c766500000a6d696e5f696e7075747303000b6d696e5f6f75747075747303010a6669656c645f73697a6503020a2e0225000340412023036a410023036a280200360200230341016a240323032302470d000b41200b0600418080040b0b4b020041000b20ffffffff000000000000000000000000ffffffff0000000000000000000000000041200b20deadbeef000000000000000000000000deadbeef000000000000000000000000");
+		let input = vec![FieldPrime::from(65536)];
+		let outputs = id.execute(&input);
```
Schaeff: I would expect this to pass, is this testing an issue with this particular module?

zokrates_core/src/imports.rs
```diff
@@ -202,7 +203,14 @@ impl Importer {
     				},
     				s => return Err(CompileError::ImportError(Error::new(format!("Packing helper {} not found", s))))
     			}
-    		} else {
+			} /*else if import.source.starts_with("WASM/") {
```
Schaeff: Just to clarify: the imports in ZoKrates are currently for ZoKrates functions. We don't have inline solvers (yet? #135), maybe we could use the imports to import those once we do...

This work is already useful without exposing it to the user (ie for ZoKrates-generated code) so this part is not crucial at this point IMHO.

zokrates_core/src/helpers/wasm.rs
```diff
@@ -0,0 +1,385 @@
...
+		let result = panic::catch_unwind(|| { 
+			let x = WasmHelper::new_code("invalid bytecode");
+			x.get_signature();
+		});
+		assert!(result.is_err());
+	}
+
+	#[test]
+	fn check_signature_exports() {
+		/* Code is the same as identity, without the `min_inputs` export */
```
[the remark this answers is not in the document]
Schaeff: Use wasm builder to start from single bytecode

zokrates_core/src/imports.rs
Schaeff: to remove

zokrates_core/src/helpers/wasm.rs
```diff
@@ -0,0 +1,385 @@
...
+
+	pub fn new_file<U: Into<String>>(u: U) -> Self {
```
Schaeff: Maybe implement `From<PathBuf>` for WasmHelper

zokrates_core/src/helpers/wasm.rs
```diff
@@ -0,0 +1,385 @@
...
+		WasmHelper::File(u.into())
```
Schaeff: Maybe implement `From<&str>`

zokrates_core/src/helpers/wasm.rs
gballet: It's about checking that if the module is buggy and returns an invalid offset, the correct error is reported.

zokrates_core/src/helpers/wasm.rs
```diff
@@ -0,0 +1,385 @@
...
+	fn execute(&self, inputs: &Vec<T>) -> Result<Vec<T>, String> {
+		let loaded_module = match self {
+			WasmHelper::File(fname) => load_from_file(&fname[..]),
+			WasmHelper::Code(code_hex) => {
+				let code = FromHex::from_hex(&code_hex[..]).map_err(|e| e.to_string())?;
+				Module::from_buffer(&code).map_err(|e| e.to_string())?
+			},
+		};
+
+		let modinst = ModuleInstance::new(&loaded_module, &ImportsBuilder::default())
```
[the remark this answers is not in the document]
gballet: Still thinking about the best way to handle this one

zokrates_core/src/helpers/wasm.rs
gballet: I currently have `From<&str>` and `From<&mut file>` which I'm told by the compiler could be turned into one thing if I enable the proper nightly feature. Still investigating this one.

Schaeff: So we're trying to keep zokrates_core free of fs operations, so I anticipate the constructor would take the bytecode here, it would have been fetched in a previous layer.

zokrates_core/src/lib.rs
```diff
@@ -9,7 +9,11 @@ extern crate serde_json;
 extern crate serde_derive;
 extern crate bimap;
 extern crate bincode;
+extern crate parity_wasm;
```
Schaeff: How about using feature flags here?

PR #207, opened 2018-12-21
Simplify flattening

Schaeff:
The flattening step has grown in size and complexity. Make it smaller and more readable.

- Refactor to have a more testable and functional style. Only flattening expressions can yield new statements into a mutable vector. This vector is returned by flatten_statement so that it is obvious what new statements were created
- Function call resolution doesn’t use a for loop anymore: the flattener is responsible for returning the correct function, the return statement is then extracted because it has a special meaning when resolving calls (currently return)
- Make functions get the `flat_functions` from the `Flattener` state directly instead of passing it as argument. There is a trade-off here as we’re passing a mutable reference to the flattener, so `flatten_expression` for example could add new functions to the state, which we ideally wouldn't want to be even possible. However, the gain in readability is higher, and we wrap anything mutating `self` in methods so that it's clear where the list of functions is mutated.
- Disallow reusing Flatteners across different programs. Unless we explicitly make sure it is safe to reuse flatteners (the flattener state should not leak from one compilation to another) we cannot assume it is. Make `flatten_program` move self instead of borrowing.
- Remove `arguments_flattened`. They are not needed anywhere unless I missed something @JacobEberhardt.
- Adjust visibility: only make the program level flattening (and semantic checking) public.
- Adjust function call flattening to prevent cloning arguments: they can be moved instead of borrowed.
- Use zip whenever possible to avoid awkward `.enumerate`
- Remove `variable` map from flattener, it's not needed anywhere unless I missed something @JacobEberhardt


Schaeff:
This was split up and merged. Closing.

PR #212, opened 2019-01-03
Update status badges to circleci

Schaeff:
See effect [here](https://github.com/Zokrates/ZoKrates/tree/doc/update-status-badges/README.md)

TODO:
- [x] Not sure why master is showing as NO BUILDS where it actually has a successful build


stefandeml:
should we add a description for the two badges?
like Branch: master <badge>, develop <badge>

PR #216, opened 2019-01-10
Derive serialization for FieldPrime

Schaeff:
We currently implement serialization manually. This is causing problems when serializing to json (to bin with bincode was fine).

`BigInt` is compatible with `serde` already given the right flags, use that.


zokrates_field/src/field.rs
```diff
@@ -323,48 +324,48 @@ impl<'a> Pow<&'a FieldPrime> for FieldPrime {
...
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        // serializer.serialize_bytes(&(*self.value.to_biguint().to_bytes_le().as_slice()))
-        serializer.serialize_bytes(&(*self.into_byte_vector().as_slice()))
-    }
-}
+// // custom serde serialization
+// impl Serialize for FieldPrime {
```
stefandeml: @Schaeff is there a reason why you want to keep the old implementation?
Serialisation looks very similar. Both use little endian byte representation.

stefandeml:
Feel free to merge!

PR #220, opened 2019-01-14
Update G2 library

Schaeff:
As advised by @musalbas, update G2 to add more checks and allow more inputs.

Commit [here](https://github.com/musalbas/solidity-BN256G2/commit/aa02c35d2883a1a01844c543e552e915cfbaed9a)

~Note that this makes GM17 a bit more expensive as G2 addition went up to 50k gas from \~32k~
Edit: was always ~50k


musalbas:
Note the gas cost in the previous version was also 50k, I just hadn't updated the readme in  while

Schaeff:
Thanks @musalbas, updating

PR #223, opened 2019-01-16
Change fs_resolver for stdlib

Schaeff:
Change the behaviour of the importer to interpret paths starting with `.` as relative to the current file, and others as relative to the `stdlib` root, in a similar way python does.

Example:
```
// in src/foo.code
import "./path/to/bar.code" // resolves to src/path/to/bar.code
import "hashes/sha256.code" // resolves to $ZOKRATES_HOME/hashes/sha256.code
```

Introduce `$ZOKRATES_HOME` to locate the stdlib, with the following presets:
- In the dev Dockerfile, `$ZOKRATES_HOME=stdlib`
- In the prod Dockerfile, `$ZOKRATES_HOME=$HOME/.zokrates`. The stdlib is distributed by copying it from the source into `$HOME/.zokrates`
- Outside of Docker, `$ZOKRATES_HOME` has to be set manually

I wonder if there's a clearer/better setup than the above. Feedback welcome.

### DIFF
--- a/zokrates_fs_resolver/src/lib.rs
+++ b/zokrates_fs_resolver/src/lib.rs
@@ -20,23 +20,25 @@ fn resolve_with_location(
     location: &String,
     source: &String,
 ) -> Result<(BufReader<File>, String, String), io::Error> {
-    let path = PathBuf::from(location).join(PathBuf::from(source));
+    let source = PathBuf::from(source);
+
+    println!(
+        "source :{:?} starts_with: {:?}",
+        source,
+        source.starts_with(".")
+    );
+
+    let base = match source.starts_with(".") {
+        false => PathBuf::from(std::env::var(ZOKRATES_HOME).unwrap_or("".to_string())),
+        true => PathBuf::from(location),
+    };
+
+    let path = base.join(PathBuf::from(source));
     let (next_location, alias) = generate_next_parameters(&path)?;
 
-    match File::open(path) {
-        Ok(f) => Ok((BufReader::new(f), next_location, alias)),
-        Err(e) => match e.kind() {
-            io::ErrorKind::NotFound => match std::env::var(ZOKRATES_HOME) {
-                Ok(v) => {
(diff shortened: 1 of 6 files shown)


zokrates_fs_resolver/src/lib.rs
```diff
@@ -41,7 +54,7 @@ mod tests {
     #[test]
     fn valid_path_with_location() {
         let (_, next_location, alias) =
-            resolve(&Some(String::from("./src")), &String::from("lib.rs")).unwrap();
+            resolve(&Some(String::from("./src")), &String::from("./lib.rs")).unwrap();
```
stefandeml: Shouldn't we change the source of the other tests as well. No it is kind of trivial that they fail, right?

zokrates_fs_resolver/src/lib.rs
```diff
@@ -18,7 +20,18 @@ fn resolve_with_location(
     location: &String,
     source: &String,
 ) -> Result<(BufReader<File>, String, String), io::Error> {
-    let path = PathBuf::from(location).join(PathBuf::from(source));
+    let source = PathBuf::from(source);
+
+    // paths starting with `./` or `../` are interpreted relative to the current file
+    // other paths `abc/def.code` are interpreted relative to $ZOKRATES_HOME
+    let base = match source.components().next() {
+        Some(Component::CurDir) | Some(Component::ParentDir) => PathBuf::from(location),
```
stefandeml: I'm not sure if this expected behavior, but with this implementation one can still import from  a the grandparent directory via this hack: `import "./../../test.code"` instead of using the not supported `.../test.code`.

Schaeff: I think it's fine to go anywhere relatively.   As you said it would be a headache to prevent that and I don't really see a reason why. You could always symlink etc... File permissions should cover that not us. The implem here is just how paths are defined in rust: `.` and `..` are first class citizens.

PR #242, opened 2019-01-29
Display field elements in a more compact way

### LINKED ISSUE #232: Show field elements x>p/2 as 'p-'(p-x) instead of x in Display Trait for Field Elements
Currently, large field elements are simply displayed as integers.
However, representing large values (`p>p/2`) as 
`'p-'p-x`
is often easier to read and interpret.

Example with `p=197`:
`x=p-1=196`
Currently, we display x as: 
`196`
I propose to display x as:
`p-1` calculated as `'p-'197-196`

Schaeff:
-1 is currently displayed as p-1, which is quite a big number.
closes #232


JacobEberhardt:
The current impl. is different than my proposal. It uses negative numbers instead of "p-" notation. I consider the "p-" display option to be more accurate, as the negative number notation is "pre-field-wrap" and does not really represent a field element.

JacobEberhardt:
In my proposal, a field element (the data type the display function is implemented for) is explicitly displayed as such, i.e., [0,p-1] 

The current implementation displays a field element x as an element in an equivalence class (-p+x). The dependence on p is implicit here, unless we added (mod p) to the statements.

I would argue that the first option is mathematically more accurate. However, I'd be happy to discuss if the inaccuracy would improve readability/usability.

JacobEberhardt:
We decided the following:

- **We only make display changes to the intermediate representation**, which is very close to the R1CS.
- The following rules to display field elements are used:
  - field element x with values [0,(p-1)/2] as x
  - field element x with values  [(p-1)/2+1,p-1] as -p+x
- Add documentation on this. This should be part of a section on the intermediate representation in order to not be confused with the field element representation used elsewhere.

zokrates_book/src/reference/ir.md
```diff
@@ -0,0 +1,17 @@
+# ZIR
+
+ZIR is the intermediate representation ZoKrates uses to represent programs. It is close to R1CS but still encapsulates witness generation.
+
+**Note that ZIR is still in development and can change without notice.**
+
+When generating R1CS constraints, very large numbers are often used, which can make reading ZIR hard for humans.
+To mitigate this, ZIR uses an abuse of language (or an isomorphism) when displaying field elements: they are shown as members of the interval `[-(p-1)/2, (p-1)/2]`.
```
JacobEberhardt: I think abuse of language is a bit strong maybe :D 
I propose: Displays field elements by using the following isomorphism:
field element x with values [0,(p-1)/2] as x
field element x with values [(p-1)/2+1,p-1] as -p+x

PR #251, opened 2019-02-04
Optimise linear combinations

Schaeff:
We currently optimise based on the `flat_absy` abstraction, mostly simplifying things like
```
b = a
c = b
```
Move the optimiser to work on ZIR and extend support for linear combinations:
```
c = a + 2b
return c + a
```
should simplify to
```
return 2a + 2b
```

Relative gains vs 57ed8adb1 (lower is better)

photo5969862107895803573

@stefandeml https://github.com/rust-unofficial/patterns/blob/master/patterns/fold.md this is useful for context!


zokrates_core/src/optimizer/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+use zokrates_field::field::Field;
+
+pub trait Optimize {
+    fn optimize(self) -> Self;
+}
+
+impl<T: Field> Optimize for Prog<T> {
+    fn optimize(self) -> Self {
+        // // only keep the main function
+        // let r = DeadCodeOptimizer::optimize(self);
```
Schaeff: remove

PR #262, opened 2019-02-08
fix loop counter persistence bug in identity WASM

gballet:
The identity WASM plugin contains a loop counter to go over the bytes of the input field. Not being cleared at the end of the loop, it would return incorrect results for subsequent calls of the `solve` function.

This PR reset the loop counter at the end of `solve` and add two tests to make sure that both `WasmHelper::IDENTITY_WASM` and `WasmHelper::BITS_WASM` produce correct outputs after being called multiple times.


gballet:
wrong branch, closing

PR #267, opened 2019-02-11
Simplify type casting

Schaeff:
When designing type casting, we introduced an `Identity` helper `a = Identity(b)` to make any type casting fit into the model of a directive call, then constraints. For bool to field (and back), the solver is the identity function, while for u8 it may be a bit decomposition for example.

What looked mostly like a way to keep things clear is actually now preventing the optimiser to reduce such statements: the optimiser does not operate on solvers as they can be non-deterministic.

Instead, use simple in-circuit bindings `a = b` which the optimiser can operate on.

Also, only bool to field is being used now, so remove all code for other cases.


Schaeff:
Closing for now because of unexpected issue in the implementation

PR #277, opened 2019-02-25
Remove interactive mode and accept inputs from stdin

Schaeff:
We have support for `--interactive` in witness generation where private inputs are passed interactively.
Meanwhile, #105 asks for inputs to be passed from a file, so not with the `-a` argument. To cover all use cases, using stdin as the input sounds ideal.

The combination of the two is awkward, as with `--interactive` turned on, one would have to first pass public inputs through stdin, then private inputs also through stdin.

Also, `--interactive` is disabled by default and I can't see a clear need for it.

This PR removes `--interactive` and enables a space-separated list of values to be passed through stdin:
```
echo "0 1 2" | zokrates compute-witness
```

Todo:
- [x] Doc update


stefandeml:
I guess we all agree that the interactive mode hasn't been used a lot and support for passing arguments via `stdin` is a big improvement for the overall user experience.

PR #279, opened 2019-02-26
Make a crate out of stdlib

Schaeff:
To improve logic testing of stdlib functions, wrap stdlib in a crate and add integration tests there.

cc @stefandeml


stefandeml:
Thanks.

I'm wondering how to deal best with arrays as inputs.
Right now we need to flatten the inputs?

stefandeml:
I guess we can't move the sha tests and use the new testing framework as libsnark is not enabled.

stefandeml:
What adding support for a folder structure in the test folder?
The current macro can handle the '/' token.

Schaeff:
> What adding support for a folder structure in the test folder?
> The current macro can not handle the `/` token.

This should be supported now, I moved from macro-expension-time to build-time test generation, so all tests in the bench folder should be turned into tests now.

> I guess we can't move the sha tests and use the new testing framework as libsnark is not enabled.
Right let's keep them where they are for now.

PR #296, opened 2019-03-15
Reduce Flattener API

Schaeff:
As we expose the constructor of the Flattener, this would make it possible to reuse it, which isn't great as it's not stateless. Remove this possibility by creating the flattener in a static call.

Also stop exposing instance-level flattening functions.

### DIFF
--- a/zokrates_core/src/flatten/mod.rs
+++ b/zokrates_core/src/flatten/mod.rs
@@ -1029,7 +1029,7 @@ impl Flattener {
         }
     }
 
-    pub fn flatten_statement<T: Field>(
+    fn flatten_statement<T: Field>(
         &mut self,
         functions_flattened: &Vec<FlatFunction<T>>,
         arguments_flattened: &Vec<FlatParameter>,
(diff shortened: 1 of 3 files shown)


zokrates_core/src/flatten/mod.rs
```diff
@@ -30,12 +30,16 @@ pub struct Flattener {
     bijection: BiMap<String, FlatVariable>,
 }
 impl Flattener {
+    pub fn flatten<T: Field>(p: TypedProg<T>) -> FlatProg<T> {
+        Flattener::new(T::get_required_bits()).flatten_program(p)
```
JacobEberhardt: Is this really a better design? Doesn't it assume that the flatten function is only called once?

Schaeff: It does. If we want to reuse Flattener instances we would need to reset their state. I feel like creating a new one each time is cleaner and cheap.

JacobEberhardt: Okay!

PR #301, opened 2019-03-18
Add cryptographic primitives to ZoKrates

stefandeml:
This PR is adding support for embedded elliptic curve computations (based on babyJubjub).
Scope:

-  twisted edwards ECC (Addition, Doubling, scalar multiplication, group-order check, on-curve check)
-  EdDSA signature verification using SHA256


zokrates_book/src/concepts/stdlib.md
```diff
@@ -26,14 +28,32 @@ There also is support for 2 round (1024bit input) and and 3 round (1536bit input
...
+
+#### Proof of private-key ownership
+
+```zokrates
+import "ecc/proofOfOwnership.code"
+```
+
+Verifies match of a given public/private keypair. Checks if the following equation holds for the provided keypair:  
+`pk = sk*G`  
+where `G` is the chosen base point of the subgroup and `*` denotes scalar multiplication.
```
Schaeff:
```suggestion
where `G` is the chosen base point of the subgroup and `*` denotes scalar multiplication in the subgroup.
```

zokrates_book/src/concepts/stdlib.md
```diff
@@ -26,14 +28,32 @@ There also is support for 2 round (1024bit input) and and 3 round (1536bit input
...
+
+#### Signature verification
+
+```zokrates
+import "signatures/verifyEddsa.code"
+```
 
-Some components of the standard library cannot yet be efficiently represented in the ZoKrates DSL language. Those functions are injected at compile-time and are available by default.
+Verifies an EdDSA Signature. Checks the correctness of a given EdDSA Signature (`R,S`) for the provided public key(`A`) and message (`M0` and `M1`).
```
Schaeff:
```suggestion
Verifies an EdDSA Signature. Checks the correctness of a given EdDSA Signature `(R,S)` for the provided public key `A` and message `(M0, M1)`.
```

zokrates_stdlib/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "zokrates_stdlib"
 version = "0.1.0"
-authors = ["schaeff <thibaut@schaeff.fr>"]
+authors = ["schaeff <thibaut@schaeff.fr>", "Stefan Deml <stefandeml@gmail.com>"]
```
Schaeff:
```suggestion
authors = ["Stefan Deml <stefandeml@gmail.com>", "schaeff <thibaut@schaeff.fr>"]
```
😉

zokrates_stdlib/stdlib/ecc/babyjubjubParams.code
```diff
@@ -0,0 +1,22 @@
+// Parameters are based on: https://github.com/HarryR/ethsnarks/tree/master/src/jubjub
```
Schaeff: Let's put a specific commit here?

zokrates_stdlib/stdlib/ecc/babyjubjubParams.code
```diff
@@ -0,0 +1,22 @@
...
+// Note: parameters will be update soon to be more compatible with zCash's implementation
```
Schaeff:
```suggestion
// Note: parameters will be updated soon to be more compatible with zCash's implementation
```

zokrates_stdlib/stdlib/ecc/edwardsAdd.code
```diff
@@ -0,0 +1,18 @@
+import "ecc/babyjubjubParams.code" as context
+// Function adds two points on a twisted Edwards curve
```
Schaeff:
```suggestion
// Add two points on a twisted Edwards curve
```

zokrates_stdlib/stdlib/ecc/edwardsNegate.code
```diff
@@ -0,0 +1,10 @@
+import "ecc/babyjubjubParams.code" as context
+// Function negates a given edwards point.
```
Schaeff:
```suggestion
// Negate a point on an Edwards curve
```

zokrates_stdlib/stdlib/ecc/edwardsOnCurve.code
```diff
@@ -0,0 +1,16 @@
+// Function checks if a provided point is on a twisted Edwards curve
```
Schaeff:
```suggestion
// Check if a point is on a twisted Edwards curve
```

zokrates_stdlib/stdlib/ecc/edwardsOrderCheck.code
```diff
@@ -0,0 +1,19 @@
+import "ecc/edwardsAdd.code" as add
+import "ecc/edwardsScalarMult.code" as multiply
+import "utils/pack/unpack256.code" as unpack256
+// Verifies that the point is not one of the low-order points.
+// If any of the points is multiplied by the cofactor, the resulting point
+// will be infinity. 
+// Returns true if the point is not one of the low-order points, false otherwise.
```
Schaeff:
```suggestion
// Returns 1 if the point is not one of the low-order points, 0 otherwise.
```

zokrates_stdlib/stdlib/ecc/edwardsScalarMult.code
```diff
@@ -0,0 +1,26 @@
...
+// https://github.com/zcash-hackworks/sapling-crypto/blob/master/src/jubjub/fs.rs#L555
+def main(field[256] exponent, field[2] pt, field[10] context) -> (field[2]):
+
+    field[2] infinity = [context[2], context[3]] 
+
+    field[2] doubledP = pt
+    field[2] accumulatedP = infinity 
+
+    field j = 255 
+    accumulatedP = if exponent[j] == 1 then doubledP else accumulatedP fi
```
Schaeff: Is there an issue with making everything fit in the same loop here?

zokrates_stdlib/stdlib/ecc/proofOfOwnership.code
```diff
@@ -0,0 +1,28 @@
+import "ecc/edwardsAdd.code" as add
+import "ecc/edwardsScalarMult.code" as multiply
+import "utils/pack/unpack256.code" as unpack256
+
+/// Verifies match of a given public/private keypair.
+///
+///    Checks if the following equation holds for the provided keypair:
+///    pk = sk*G
+///    where G is the chosen base point of the subgroup
+///    and * denotes scalar multiplication
```
Schaeff:
```suggestion
///    and * denotes scalar multiplication in the subgroup
```

zokrates_stdlib/stdlib/ecc/proofOfOwnership.code
```diff
@@ -0,0 +1,28 @@
...
+///
+/// Arguments:
+///    pk: Curve point. Public key.
+///    sk: Field element. Private key.
+///    context: Curve parameters (including generator G) used to create keypair.
+///
+/// Returns:
+///     Return true for pk/sk being a valid keypair, false otherwise.
```
Schaeff:
```suggestion
///     Return 1 for pk/sk being a valid keypair, 0 otherwise.
```

zokrates_stdlib/stdlib/hashes/sha256/1024bitPadded.code
```diff
@@ -0,0 +1,15 @@
+import "./1536bit.code" as sha256
+// A function that takes 2 field[256] arrays as inputs
```
Schaeff:
```suggestion
// Take two field[256] arrays as input 
```

zokrates_stdlib/stdlib/signatures/verifyEddsa.code
```diff
@@ -0,0 +1,47 @@
...
+import "ecc/edwardsScalarMult.code" as scalarMult 
+import "ecc/edwardsAdd.code" as add
+import "utils/pack/unpack256.code" as unpack256
+import "ecc/edwardsOnCurve.code" as onCurve
+import "ecc/edwardsOrderCheck.code" as orderCheck
+
+/// Verifies an EdDSA Signature.
+///
+/// Checks the correctness of a given EdDSA Signature (R,S) for the provided
+/// public key(A) and message (M0 and M1).
```
Schaeff:
```suggestion
/// public key A and message (M0, M1).
```

zokrates_stdlib/stdlib/signatures/verifyEddsa.code
```diff
@@ -0,0 +1,47 @@
...
+/// Arguments:
+///    R: Curve point. Hidden version of the per-message nonce.
+///    S: Field element. Signature to be verified.
+///    A: Curve point. Public part of the key used to create S.
+///    M0: 256bit array. First 256bits of the message used to create S  .
+///    M1: 256bit array. Trailing 256bits of the message used to create S  .
+///    context: Curve parameters used to create S.
+///
+/// Returns:
+///     Return true for S being a valid EdDSA Signature, false otherwise.
```
Schaeff:
```suggestion
///     Return 1 for S being a valid EdDSA Signature, 0 otherwise.
```

zokrates_stdlib/stdlib/ecc/edwardsScalarMult.code
stefandeml: one could make it fit into the same loop that way
```
    field[2] doubledP = pt
    field[2] accumulatedP = infinity
    
    for field i in 0..256 do
        field j = 255 - i 
        candidateP = add(accumulatedP, doubledP, context)
        accumulatedP = if exponent[j] == 1 then candidateP else accumulatedP fi
        doubledP = add(doubledP, doubledP, context)
    endfor

    1 == assertOnCurve(accumulatedP, context)
```

def. looks nicer but means we have one additional point doubling that's not needed.

zokrates_stdlib/stdlib/hashes/sha256/1024bitPadded.code
```diff
@@ -0,0 +1,15 @@
+import "./1536bit.code" as sha256
+// Take two field[256] arrays as input 
+// and returns their sha256 full round output as an array of 256 field elements.
+def main(field[256] a, field[256] b, field[256] c, field[256] d) -> (field[256]):
```
HarryR: Is it possible to have a generic version of the SHA256 function?

See: https://github.com/HarryR/ethsnarks/blob/master/src/utils.cpp#L80 - This is a `bits2blocks_padded` function, which converts an array of bits into a sequence of blocks with SHA256-compliant padding. The output of the function can be passed into a SHA256 chain.

Instead of having specific versions of the SHA2-256 function for different lengths, it would be better to have a single function which handles an arbitrary number of bits as input then adds padding as is necessary.

I think it would be a good exercise to see how this language can handle something like that.

Schaeff: Absolutely, that's the goal. Not in this PR though!

One way that's feasible even in the short term is to allow for loop bounds to be any expression, and only make sure that these expressions are a compile time constant. Then you could write

`sha256(x, 42)` with 42 the number of rounds
but not `sha256(x, n)` where n is an input to the program

zokrates_stdlib/stdlib/ecc/edwardsAdd.code
```diff
@@ -0,0 +1,18 @@
...
+
+    field a = context[0]
+    field d = context[1] 
+
+    field u1 = pt1[0]
+    field v1 = pt1[1]
+    field u2 = pt2[0]
+    field v2 = pt2[1]
+
+    field uOut = (u1*v2 + v1*u2) / (1 + d*u1*u2*v1*v2)
```
kobigurk: This can be optimized by quite a bit.
It costs 2 constraints for the return arguments, and the rest 21 for the addition.

It can be optimized so the addition (in zokrates code) so the addition costs 14 constraints by reusing some products:

    field a = context[0]
    field d = context[1]

    field u1 = pt1[0]
    field v1 = pt1[1]
    field u2 = pt2[0]
    field v2 = pt2[1]

    field beta = u1*v2
    field gamma = v1*u2
    field delta =  (u1+v1)*(u2+v2)
    field tau = beta*gamma
    field uOut = (beta+gamma) / (1+d*tau)
    field vOut = (delta - beta - gamma) / (1-d*tau)
    return [uOut, vOut]

In pure constraints code this would have been 6 constraints, so there's still room for improvement (maybe the optimizer touches this?)

zokrates_stdlib/stdlib/ecc/edwardsOrderCheck.code
```diff
@@ -0,0 +1,19 @@
...
+// Verifies that the point is not one of the low-order points.
+// If any of the points is multiplied by the cofactor, the resulting point
+// will be infinity. 
+// Returns 1 if the point is not one of the low-order points, 0 otherwise.
+// Curve parameters are defined with the last argument
+// https://github.com/zcash-hackworks/sapling-crypto/blob/master/src/jubjub/edwards.rs#L166
+def main(field[2] pt, field[10] context) -> (field):
+
+    field cofactor = context[7] 
+    field[256] cofactorExponent = unpack256(cofactor)
```
kobigurk: Making the cofactor this generic costs a lot. Cofactors are usually small, and so multiplying by 4 and 8 would save many constraints.

zokrates_stdlib/stdlib/signatures/verifyEddsa.code
```diff
@@ -0,0 +1,47 @@
...
+///
+/// Returns:
+///     Return 1 for S being a valid EdDSA Signature, 0 otherwise.
+def main(private field[2] R, private field S, field[2] A, field[256] M0, field[256] M1, field[10] context) -> (field):
+
+    field[2] G = [context[4], context[5]]
+
+    // Check if R is on curve and in prime-order sub-group. A is public input and can be checked offline 
+    field isOnCurve = onCurve(R, context) // throws if R is not on curve
+    field isPrimeOrder = orderCheck(R, context)
```
kobigurk: This doesn't check it's in the prime order subgroup, it checks that it's not in the small subgroup. It could be in a coset, i.e., P = a*G + b*H where G is the generator of the prime subgroup and H is the generator of the small subgroup, and the check would still pass.

The check itself is fine, just not the comment :)

Additionally, verification in edDSA multiplies by the cofactor in the verification equation (why I think you tried to avoid by checking you're in the prime subgroup).

It might not be completely necessary, but it deviates from https://tools.ietf.org/html/rfc8032#page-43 and librustzcash, for example.

zokrates_stdlib/stdlib/ecc/edwardsOrderCheck.code
HarryR: Agreed, in an ideal world the cofactor would take advantage of constant reduction.

For Twisted Edwards curves the cofactor will be `2^c`, so a chain of doubling operations (say, three of them) should be sufficient for a cofactor of 8.

zokrates_stdlib/stdlib/signatures/verifyEddsa.code
stefandeml: Thanks a lot for pointing that out. I will update the comment accordingly.

Regarding the cofactorless verification this implementation follows:
https://eprint.iacr.org/2015/677.pdf  (see section "Cofactorless  verification")
We've decided to go that way to be compatible with the work in
- ethsnarks: https://github.com/HarryR/ethsnarks/blob/master/ethsnarks/eddsa.py#L193
- thematter: https://github.com/matter-labs/sapling-crypto/blob/master/src/circuit/baby_eddsa.rs#L121

zokrates_stdlib/stdlib/ecc/edwardsOrderCheck.code
stefandeml: Yes, as we will only support Edwards curves we can easily go for this optimisation and save these constraints.

@Schaeff ideally this could also be caught by the optimiser as the cofactor is a constant and all additional iterations in the `scalarMul` loop only create unused variables (see https://github.com/Zokrates/ZoKrates/issues/303)

stefandeml: @Schaeff  for reference:

> Unnpack&scalarMul, point hardcoded = 14061 constraints
> Unnpack&scalarMul, point as argument = 14096 constraints
> Double double double, point as argument = 72 constraints

stefandeml: We will hardcode the check for now using the double^3 approach, but will leave a not there to make it generic as soon as the optimiser catches these unused variables / assignments.

zokrates_stdlib/stdlib/ecc/edwardsAdd.code
stefandeml: Thanks for pointing that out. 

For now we want to push most of the optimisations to the optimiser. 
I opened an issue https://github.com/Zokrates/ZoKrates/issues/304 for the re-use of intermediary variables.
I also opened an issue to improve constant propagation https://github.com/Zokrates/ZoKrates/issues/307

On additional note: We could also use Montgomery arithmetics in some cases to reduce the number of constraints for incomplete addition to 3.

stefandeml:
thanks a lot for the comments @kobigurk @HarryR - very much appreciated.

Mahsa-Bastankhah:
Hi.does this new update supports encryption? in other word can I create proof that a cipher text is correct encryption of a secret input?

rbkhmrcr:
@Mahsa-Bastankhah if you're using elgamal or another elliptic curve based encryption protocol then this is enough to allow you to implement a scheme where you prove that, for example, something is a correctly formed elgamal encryption of a known secret input, yep!

PR #305, opened 2019-03-20
Adds the ability to export a R1CS.json in the setup phase.

alexpArtos:


alexpArtos:
Hi @HarryR . I agree with you and we debated that ourselves. 
The reason is that we could not find a place where common functionality for both backends would be placed (except for a single writeToFile function). 

The current version of ZoKrates duplicates significant parts of the code in gm17.cpp and pghr13.cpp, namely `createConstraintSystem` and `exportInput`.

With a philosophy of minimal intrusion, we decided to keep the same approach. Also, @Schaeff mentioned in Gitter they are doing a refactoring around these parts so we didn't want to mess with the structure too much.

PR #319, opened 2019-04-10
Create binaries with libsnark disabled on new tags

Schaeff:
We tell users to use Docker. This is a pain.
Now that we have a pure rust implementation on groth16, distribute binaries for common platforms.
- Disable libsnark by default, still controlled by `$WITH_LIBSNARK`
- Build releases on Travis
- Offer a one line script to install ZoKrates at zokrat.es


zokrates_book/src/gettingstarted.md
```diff
@@ -17,10 +25,9 @@ From there on, you can use the `zokrates` CLI.
 You can build the container yourself from [source](https://github.com/JacobEberhardt/ZoKrates/) with the following commands:
```
stefandeml: wrong link

PR #322, opened 2019-04-14
Replace BTree with Vec in LinComb

### LINKED ISSUE #312: Switch IR implementation of LinComb from map to vec
We represent a linear combination as a Map where the key is a variable and the value is the coefficient for that variable. This is intuitive, but operations are expensive as we need to reduce each variable to a single record, which requires iterating on `a` and `b` to get `a + b`

If we used `Vec<(Variable, Coefficient)>`, LinComb arithmetics are reduced to array concatenation. We basically allow `a + a + a` where with map we would reduce to `3a`
.

We could always cheaply reduce to Map from that non-reduced representation if needed.

eupn:
Resolves #312.

Replaced `BTreeMap` with `Vec` of pairs `(var, value)`. Unit tests passed, although the order of sum items that was dependent on the order of keys in`BTreeMap` is changed.

?r @Schaeff


Schaeff:
Awesome, let me have a look at it! It's true that the BTreeMap representation has the advantage of being deterministic so great for testing, let's see what we can do there.

zokrates_core/src/ir/expression.rs
```diff
@@ -105,14 +105,27 @@ impl<T: Field> Add<LinComb<T>> for LinComb<T> {
 
     fn add(self, other: LinComb<T>) -> LinComb<T> {
```
Schaeff: Any hints why just concatenating `self.0` and `other.0` would not work?

zokrates_core/src/ir/expression.rs
```diff
@@ -69,8 +69,9 @@ impl<T: Field> LinComb<T> {
 
     pub fn try_summand(&self) -> Option<(&FlatVariable, &T)> {
         if self.0.len() == 1 {
-            return self.0.iter().next();
+            return self.0.first().map(|(id, k)| (id, k))
```
Schaeff: Am I right to guess that `map` was added to turn `&(_, _)` into `(&_, &_)` ?
Maybe `Option<&(FlatVariable, T)>` is fine as a return type for `try_summand` in which case this would not be needed?
In any case, would be cool to clarify here :)

zokrates_core/src/ir/expression.rs
```diff
@@ -105,14 +105,27 @@ impl<T: Field> Add<LinComb<T>> for LinComb<T> {
 
     fn add(self, other: LinComb<T>) -> LinComb<T> {
         let mut res = self.0;
+        let mut to_remove = BTreeSet::new();
+
         for (k, v) in other.0 {
-            let new_val = v + res.get(&k).unwrap_or(&T::zero());
-            if new_val == T::zero() {
-                res.remove(&k)
+            let var;
```
Schaeff: We try not to declare without allocating when possible. I think it's possible here?

zokrates_core/src/ir/expression.rs
```diff
@@ -122,14 +135,27 @@ impl<T: Field> Sub<LinComb<T>> for LinComb<T> {
...
+            if new_val != T::zero() {
+                res.push((k, new_val));
             } else {
-                res.insert(k, new_val)
-            };
+                to_remove.insert(k);
+            }
         }
+
+        res.retain(|(id, _)| !to_remove.contains(id));
```
Schaeff: So you're filtering out terms that have coefficient 0 at the cost of iterating on the full lincomb. My intuition is that this is close in complexity to reducing to a hashmap, which is what we're trying to avoid. Maybe you have some insights here?

zokrates_core/src/ir/expression.rs
```diff
@@ -184,23 +210,27 @@ mod tests {
             let a: LinComb<FieldPrime> = FlatVariable::new(42).into();
             let b: LinComb<FieldPrime> = FlatVariable::new(42).into();
             let c = a + b.clone();
-            let mut expected_map = BTreeMap::new();
-            expected_map.insert(FlatVariable::new(42), FieldPrime::from(2));
-            assert_eq!(c, LinComb(expected_map));
+
+            let expected_vec = vec![
+                (FlatVariable::new(42), FieldPrime::from(1)),
+                (FlatVariable::new(42), FieldPrime::from(2)),
```
Schaeff: Shouldn't this be `[(_42, 1), (_42, 1)]` as the unreduced original `(_42 => 2)`

zokrates_core/src/ir/expression.rs
```diff
@@ -184,23 +210,27 @@ mod tests {
...
+            ];
+            assert_eq!(c, LinComb(expected_vec));
         }
         #[test]
         fn sub() {
             let a: LinComb<FieldPrime> = FlatVariable::new(42).into();
             let b: LinComb<FieldPrime> = FlatVariable::new(42).into();
             let c = a - b.clone();
-            assert_eq!(c, LinComb::zero());
+            assert_eq!(LinComb::zero(), c);
```
Schaeff: Here I believe we should expect `[(_42, 1), (_42, -1)]`

zokrates_core/src/ir/expression.rs
```diff
@@ -184,23 +210,27 @@ mod tests {
...
         }
 
         #[test]
         fn display() {
             let a: LinComb<FieldPrime> =
                 LinComb::from(FlatVariable::new(42)) + LinComb::summand(3, FlatVariable::new(21));
-            assert_eq!(&a.to_string(), "3 * _21 + 1 * _42");
+            assert_eq!(&a.to_string(), "1 * _42 + 3 * _21");
```
Schaeff: It would make sense for `fmt::Display` to reduce to canonical before printing, so that this test ideally would not change?

zokrates_core/src/ir/expression.rs
```diff
@@ -236,7 +266,7 @@ mod tests {
                     + LinComb::summand(4, FlatVariable::new(33)),
                 right: LinComb::summand(1, FlatVariable::new(21)),
             };
-            assert_eq!(&a.to_string(), "(4 * _33 + 3 * _42) * (1 * _21)");
+            assert_eq!(&a.to_string(), "(3 * _42 + 4 * _33) * (1 * _21)");
```
Schaeff: Same here

zokrates_core/src/ir/expression.rs
eupn: Yes, `map` is added to deref. references inside a tuple. It would be more clear to change return type instead, I'll try to do so.

zokrates_core/src/ir/expression.rs
eupn: Can you please give me an intuition on how to perform `add` and `sub` on a linear combinations that are represented internally as a `Vec` of pairs `(var, value)`? It seems that I missing something whilst trying to mimic the `BTreeMap`-based code.

zokrates_core/src/ir/expression.rs
eupn: Yes, maybe I should build a resulting lincomb while iterating over terms and just not include into it those of them that are resulting to zero.

zokrates_core/src/ir/expression.rs
eupn: Can you please clarify what a canonical representation is? I probably should implement a method `as_canonical() -> Self` for a lincomb and use that in `fmt::Display` and in communication with backend.

zokrates_core/src/ir/expression.rs
eupn: It seems that concatenation is working for addition, but I not sure that it would for a substraction.

zokrates_core/src/ir/expression.rs
eupn: Yes it should, good catch!

zokrates_core/src/ir/expression.rs
Schaeff: Yes maybe we need to define some types here for it to be clearer. By canonical I mean BTreeMap based, where you have a single, deterministic way to represent a given linear combination (BTreeMap adds the ordered aspect which you don't get from HashMap)

zokrates_core/src/ir/expression.rs
Schaeff: I believe for subtraction you could negate the coefficient of the second lincomb
`self.0.into_iter().chain(other.0.into_iter().map(|(var, coeff) (-var, coeff))).collect()` or something

zokrates_core/src/ir/expression.rs
eupn: Can we just sort the underlying `Vec` by variable ID?

Something, like:
```rust
pub fn as_canonical(&self) -> Self {
  let mut sorted = self.0.clone();
  sorted.sort_by_key(|(id, _)| id);

  LinComb(sorted)
}
```

Schaeff: Sorting is one thing, but we mostly need each variable to appear only once in each linear combination, which the vector does not enforce. Hence the BTreeMap imo.

zokrates_core/src/ir/expression.rs
Schaeff: You could filter out `(_42, 0)` which I'm not sure where it would come from, but I don't think you should filter out `[(_42, 1), ...,  (_42, -1)]` as it's the point of this PR to not reduce at each operation and instead only reduce when needed.

Schaeff:
@eupn I had a quick go here https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=07c263548f34a377c3497a8b94d62a32
maybe that can help!

eupn:
> @eupn I had a quick go here https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=07c263548f34a377c3497a8b94d62a32
> maybe that can help!

Got it, thanks a lot!

Schaeff:
Hey @eupn I realise we don't tell contributors to do it but we use https://github.com/rust-lang/rustfmt. 
I'll add this somewhere so it's clear!

Schaeff:
Hey @eupn I think we're getting there! just a bit of context in the last change: the breaking test on 01c4d18 was due to the fact that that equality on programs (which is being tested) is derived automatically, so two lincomb that are "equivalent" (ie have the same canonical representation) are not "equal" (as they can have different representation in a vector of pairs). Therefore instead of making the optimizer canonical, I changed the definition of equality between lincomb to canonical equivalence.

Let me know if you have any thoughts or I missed something!

Schaeff:
@eupn and on this last change, I made sure `(2*~one + 3*~one) * (...)` is recognised as linear: without it, it would not try to reduce to `5*~one` as the lincomb is of size > 1

PR #326, opened 2019-04-16
Use buffered io in CLI and tests

Schaeff:
Quick tests show that some programs compile 4x faster, and reading programs and witnesses is much faster too.


Schaeff:
@stefandeml  please ignore Travis, it ran because I opened against master.

Issue #347, opened 2019-05-10
contents of proof.json contains invalid json for gm17 and pghr13 proving scheme


petscheit:
When using the gm17 and pghr13 proving schemes to generate a proof, the content of proof.json doesn't have a consistent format. The array with input as a key doesn't store the data in correct hex format ("0x1bba") but in an invalid format (1bba). All other entries of this file are formated in valid hex. When generating a proof with g16 scheme, this doesn't happen and all entries are formated correctly.

PR #352, opened 2019-05-15
Fixes issue #347 and removes duplicate examples

petscheit:
-Issue #347 is fixed, input section of proof.jsnn now contains valid hex strings for all proving-schemes
-Since there where loads of duplicates and very basic examples a lot of them are removed here


Schaeff:
Thanks for this! Looks good, however would you mind submitting two separate PRs for these two things? The JSON one I think should be easy to merge, the examples one I'd like to discuss a bit more.

PR #354, opened 2019-05-22
fixes issue #347

petscheit:
fixes issue described in issue 347


Schaeff:
Sweet, however this still includes example removal, maybe `git checkout develop zokrates_cli/examples` would do

petscheit:
Ok im super confused how this happen. Sorry about that, my git has become a bit rusty

PR #355, opened 2019-05-22
Example removal

petscheit:
removed examples that are either very simple or have been covered in the docs or another example. All up for discussion.

(sorry for the reverted commit, my branched got all tangled up, should be fine like this though)


Schaeff:
I understand what we're trying to solve here (the example folder being too big and messy) deleting this would remove a bunch of compilation tests that we're doing now. These tests are not very powerful as they only check whether the source can be parsed/compiled, but doesn't check the behavior at all.
If we want to reduce the size of the examples folder, I think we should define what examples we want to have in there, and migrate the rest to integration tests.

Issue #201, opened 2018-12-11
Use ABIv2 in the verifier


Schaeff:
Our current solidity verifier is quite verbose, partly due to the lack of support for structs as params in the current ABI.

Explore how we could use ABIv2 to make the contract clearer.

Schaeff:
The only change needed is
```
verifyTx(
   Proof memory p,
...
```

Schaeff:
We have this now thanks to @petscheit

PR #359, opened 2019-05-29
Updates ABI Encoder #201

petscheit:
Updates the ABIEncoder to v2, which enables us to pass structs as parameters to functions. Simplifies the Verifier.sol a bit and is also needed for application tests (takes a lot of complexity out of the testing script because I can simply pass the proof object from the proof.json instead of parsing every value separately when calling verifyTx)


Schaeff:
Would be super awesome to have this!
A few thoughts as I've tried this in the past.
- Are you able to test a contract on Remix with this? I tried in the past an there was a missing piece, can't remember if it was on the Remix encoding side or if ABIV2 didn't support a certain type of recursive struct. Maybe the issue was with the verifying key, is this why you didn't change it? If not why?
- Could you share a gas cost comparison?
- ABIV2 is likely to stay experimental for a little while now, maybe it would make sense to have two versions as I agree it's great for testing and UX but people may still want the "stable" one

petscheit:
Regarding Remix:
-currently structs containing structs can't be passed with remix, as far as I understand. Wasn't able to make that work, so thats actually quite a downside. 

Gas comparison:
image

I'm currently having some issue with the gm17 contract, for some reason the compiler wont link the imported libraries properly so can't test the gas usage for that one atm, still investigating. Overall it seems to have a positive effect on deployment and execution cost. I used the root code for the comparison, with the values given in the getting started. 

Regarding having two versions of the contract, I agree. Should I implement that?

Schaeff:
@petscheit I checked with @yann300 from Remix and passing structs works fine *with the new interface* which is being rolled out. So with this and ABIv2 being still experimental, I guess for this PR it would make sense to have a `--abiv2` flag in `zokrates export-verifier` which is off by default (until things get stable) and which creates an ABIv2 contract for testing. @JacobEberhardt thoughts welcome!

JacobEberhardt:
I agree. This ABIv2 contract should be covered by the tests @petscheit is currently adding to the CI pipeline as well.

petscheit:
OK, sounds good to me too. I"ll get going on this now and implement it into the testing pipeline aswell. @Schaeff Should I change the proof.json format so it works for the different ABI's? I'm thinking the proof.json could have two sections, one for abiv1 and one for abiv2 and they contain the data needed for the proof in copy/paste friendly format. Would also make the circleci integration easier

JacobEberhardt:
If i may chime in here: The goal is to have the proof.json as an independent minimal format and then offer a cli-option to convert into copy-past friendly representations for various target platforms/IDEs.

Schaeff:
Hey @petscheit , closing this as we merged #414 please shout if I got this wrong!

PR #363, opened 2019-06-04
Remove strings from absy

Schaeff:
With the pest parser, the AST nodes bear references to the source string. Carry these over to absy instead of creating owned strings.

### DIFF
--- a/zokrates_core/src/absy/mod.rs
+++ b/zokrates_core/src/absy/mod.rs
@@ -152,29 +152,6 @@ impl<'ast, T: Field> fmt::Display for Assignee<'ast, T> {
     }
 }
 
-// impl<'ast, T: Field> From<ExpressionNode<'ast, T>> for AssigneeNode<T> {
-//     fn from(e: ExpressionNode<'ast, T>) -> Self {
-//         match e.value {
-//             Expression::Select(box e1, box e2) => match e1 {
-//                 ExpressionNode {
-//                     value: Expression::Identifier(id),
-//                     start,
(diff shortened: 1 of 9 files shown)


zokrates_core/src/semantics.rs
```diff
@@ -266,7 +274,7 @@ impl Checker {
         }
 
         Ok(TypedFunction {
-            id: funct.id.clone(),
```
Schaeff: we only create the owned strings for the typed_absy

zokrates_core/src/absy/variable.rs
```diff
@@ -2,37 +2,39 @@ use crate::absy::Node;
...
 use std::fmt;
 
-#[derive(Serialize, Deserialize, Clone, PartialEq, Hash, Eq)]
-pub struct Variable {
-    pub id: String,
+use crate::absy::Identifier;
+
+#[derive(Clone, PartialEq, Hash, Eq)]
+pub struct Variable<'ast> {
+    pub id: Identifier<'ast>,
```
Schaeff: Variables do not contain owned strings anymore, but references to the source string

zokrates_core/src/absy/mod.rs
```diff
@@ -20,15 +20,17 @@ use crate::imports::ImportNode;
 use std::fmt;
 use zokrates_field::field::Field;
 
+pub type Identifier<'ast> = &'ast str;
```
Schaeff: An identifier is a reference

PR #366, opened 2019-06-05
Add print-proof command to cli

petscheit:
-cli now contains the print-proof command which will output a pastable version of the proof
-formats to pick from are json or remix compatible
-also will be used by testing pipeline in the future


Schaeff:
The reason the build fails was solved on develop, merging develop back should solve it.

PR #368, opened 2019-06-06
Remove strings from typed_absy

Schaeff:
- removes strings in typed_absy
- adapt semantic check to keep working (less use of references, cloning)
- adapt flattening to work with non-string-based identifiers, especially for arrays as we used to use "foo_c42"  for the 43th item of `foo`. Now we instead keep track of all flat variables for a given typed variable and access the 43th item of that.

### DIFF
--- a/zokrates_core/src/flatten/mod.rs
+++ b/zokrates_core/src/flatten/mod.rs
@@ -1336,8 +1336,8 @@ mod tests {
         let mut statements_flattened = vec![];
         let statement = TypedStatement::MultipleDefinition(
             vec![
-                Variable::field_element("a".to_string()),
-                Variable::field_element("b".to_string()),
+                Variable::field_element("a".into()),
+                Variable::field_element("b".into()),
             ],
             TypedExpressionList::FunctionCall(
                 "foo".to_string(),
@@ -1385,8 +1385,8 @@ mod tests {
         }];
         let statement = TypedStatement::MultipleDefinition(
             vec![
-                Variable::field_element("a".to_string()),
-                Variable::field_element("b".to_string()),
+                Variable::field_element("a".into()),
+                Variable::field_element("b".into()),
             ],
             TypedExpressionList::FunctionCall(
                 "dup".to_string(),
@@ -1437,7 +1437,7 @@ mod tests {
         }];
         let mut statements_flattened = vec![];
         let statement = TypedStatement::Definition(
-            TypedAssignee::Identifier(Variable::field_element("a")),
+            TypedAssignee::Identifier(Variable::field_element("a".into())),
             TypedExpression::FieldElement(FieldElementExpression::FunctionCall(
                 "foo".to_string(),
                 vec![],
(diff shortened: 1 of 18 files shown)


zokrates_core/src/flatten/mod.rs
```diff
@@ -11,19 +11,18 @@ use crate::typed_absy::*;
...
 
 /// Flattener, computes flattened program.
 #[derive(Debug)]
-pub struct Flattener {
+pub struct Flattener<'ast> {
     /// Index of the next introduced variable while processing the program.
     next_var_idx: usize,
     ///
-    bijection: BiMap<String, FlatVariable>,
+    layout: HashMap<Identifier<'ast>, Vec<FlatVariable>>,
```
Schaeff: Any Identifier in the TypedProgram maps to a series of FlatVariable (one for Field, n for FieldArray(n))

zokrates_core/src/flatten/mod.rs
```diff
@@ -725,8 +722,8 @@ impl Flattener {
                         FieldElementArrayExpression::Identifier(size, id) => {
                             assert!(n < T::from(size));
                             FlatExpression::Identifier(
-                                self.get_latest_var_substitution(&format!("{}_c{}", id, n))
-                                    .clone(),
+                                self.layout.get(&id).unwrap().clone()
+                                    [n.to_dec_string().parse::<usize>().unwrap()],
```
Schaeff: get flat variables for id, take the nth. Could be made cleaner

zokrates_core/src/flatten/mod.rs
```diff
@@ -1177,35 +1185,12 @@ impl Flattener {
...
-                                }
-                            }
-                        }
+                        let rhs = rhs_flattened.expressions.into_iter();
+
+                        let vars = vars.into_iter().flat_map(|v| self.use_variable(&v));
 
-                        // we should have exhausted the return values
-                        assert_eq!(iterator.next(), None);
+                        statements_flattened
```
Schaeff: use `zip` instead of low level iterator magic

zokrates_core/src/flatten/mod.rs
```diff
@@ -1299,30 +1265,31 @@ impl Flattener {
     /// # Arguments
     ///
     /// * `name` - a String that holds the name of the variable
-    fn use_variable(&mut self, name: &String) -> FlatVariable {
-        // issue the variable we'll use
-        let var = self.issue_new_variable();
+    fn use_variable(&mut self, variable: &Variable<'ast>) -> Vec<FlatVariable> {
```
Schaeff: We issue flat variables for typed variables, whatever the type

zokrates_core/src/static_analysis/inline.rs
```diff
@@ -5,14 +5,14 @@ use crate::types::{Signature, Type};
...
 use zokrates_field::field::Field;
 
-pub struct Inliner<T: Field> {
-    functions: Vec<TypedFunction<T>>,
-    statements_buffer: Vec<TypedStatement<T>>,
-    context: Vec<(String, usize)>,
+pub struct Inliner<'ast, T: Field> {
+    functions: Vec<TypedFunction<'ast, T>>,
+    statements_buffer: Vec<TypedStatement<'ast, T>>,
+    context: Vec<(&'ast str, Signature, usize)>,
```
Schaeff: the context is function name + signature + number of call

zokrates_core/src/static_analysis/mod.rs
```diff
@@ -33,6 +33,7 @@ impl<T: Field> Analyse for TypedProg<T> {
         let r = Propagator::propagate(r);
         // apply inlining strategy
         let r = Inliner::inline(r);
+        println!("{}", r);
```
JacobEberhardt: should this be here?

PR #383, opened 2019-06-14
Add array initializer sugar

Schaeff:
Introduce `[42; 3]` as sugar for `[42, 42, 42]`
The count has to be a parse-time constant (no constant propagation!)


zokrates_cli/examples/book/array.code
```diff
@@ -1,4 +1,5 @@
 def main() -> (field):
```
JacobEberhardt: The documentation could benefit from more extensive array examples.

PR #386, opened 2019-06-15
add Abiv2 option to cli & changes proof format to uniform format

petscheit:
-adds abiv2 flag to export-verifier command -> will export contracts on v1 or v2 (default v1)
-normalizes format of proof.json so all schemes have a matching one
-removes testing functionality from print-proof command


petscheit:
I changed the format of the proof.json so the keys of the proof are in caps (so they match the contract format, which makes print-proof a lot simpler), now the tests fail because in the g16 tests, I had to capitalize the keys. How should this be handled?

Schaeff:
If possible I'd rather not change the format of the json proof as it's a breaking change. I think we can afford the extra complexity of caps conversion. Would that work?

petscheit:
I mean currently the proofs have different formats. E.g. in gm17 and pghr13 one key is input, while in g16 its called inputs. Similar story for the proof keys, in gm17 the A key is capitalized, in g16 the a key  is not. Is it supposed to be like that?

Schaeff:
I see, good point. We had a think and it would make sense to have everything written consistently with no caps, and with `_` separators (snake_case)
gammaABC -> gamma_abc
IC -> ic

And inputs should be `inputs` everywhere.
That's quite a few places to change things for all three protocols, is that something you could do? I guess your testing work would guarantee that we don't miss one of them.

petscheit:
Ok, that sounds like a good solution. I'll get on it now and change the variables in the contracts and in the proof.json to that format. And I also believe the testing should catch any errros

PR #395, opened 2019-06-19
Update G2 Solidity library

musalbas:
This reduces the gas cost for addition by about 20,000 by using the modexp precompile for _modInv using Fermat's little theorem instead of Extended Euclidean algorithm, see https://github.com/musalbas/solidity-BN256G2/issues/2 and https://en.wikipedia.org/wiki/Modular_multiplicative_inverse#Using_Euler's_theorem


Schaeff:
Hey @musalbas there was a strange conflict and I ended up merging manually, your commits are [here](https://github.com/Zokrates/ZoKrates/commit/4a802a8d40b2522a00b58bc64d7b649b97580d56)

Thanks for this :)

PR #399, opened 2019-06-21
Add boolean constants

Schaeff:
Boolean constants `true` and `false` are supported in typed_absy and flattening, but not in the parser.
Add them to the grammar, pest ast, and absy.

### DIFF
--- a/zokrates_cli/examples/boolean_literal.code
+++ b/zokrates_cli/examples/boolean_literal.code
@@ -1,0 +1,2 @@
+def main(bool a) -> (bool):
+	return (false || true) && false
--- a/zokrates_core/src/absy/from_ast.rs
+++ b/zokrates_core/src/absy/from_ast.rs
@@ -221,12 +221,12 @@ impl<'ast, T: Field> From<pest::IterationStatement<'ast>> for absy::StatementNod
             .collect();
 
         let from = match from.value {
-            absy::Expression::Number(n) => n,
+            absy::Expression::FieldConstant(n) => n,
             e => unimplemented!("For loop bounds should be constants, found {}", e),
         };
 
         let to = match to.value {
-            absy::Expression::Number(n) => n,
+            absy::Expression::FieldConstant(n) => n,
             e => unimplemented!("For loop bounds should be constants, found {}", e),
         };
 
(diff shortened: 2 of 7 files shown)


zokrates_parser/src/zokrates.pest
```diff
@@ -77,7 +77,9 @@ unary_expression = { op_unary ~ term }
 
 assignee = { identifier ~ ("[" ~ range_or_expression ~ "]")* }
 identifier = @{ ((!keyword ~ ASCII_ALPHA) | (keyword ~ (ASCII_ALPHANUMERIC | "_"))) ~ (ASCII_ALPHANUMERIC | "_")* }
-constant = @{ "0" | ASCII_NONZERO_DIGIT ~ ASCII_DIGIT* }
+constant = { decimal_number | boolean_constant }
+decimal_number = @{ "0" | ASCII_NONZERO_DIGIT ~ ASCII_DIGIT* }
+boolean_constant = { "true" | "false" }
```
JacobEberhardt: In the grammar, these should probably be called `boolean_literals`. Unimportant detail, though.

zokrates_pest_ast/src/lib.rs
```diff
@@ -552,7 +552,32 @@ mod ast {
 
     #[derive(Debug, FromPest, PartialEq, Clone)]
     #[pest_ast(rule(Rule::constant))]
-    pub struct ConstantExpression<'ast> {
+    pub enum ConstantExpression<'ast> {
+        DecimalNumber(DecimalNumberExpression<'ast>),
+        BooleanConstant(BooleanConstantExpression<'ast>),
```
JacobEberhardt: Same here. Terminology-wise `DecimalNumber` and `BooleanConstant` seem to be different levels of abstraction. 
We could move to `FieldLiteral` and `BooleanLiteral`. Not sure it's worth the effort.

zokrates_cli/examples/boolean_literal.code
```diff
@@ -0,0 +1,2 @@
+def main(bool a) -> (bool):
+	return (false || true) && false
```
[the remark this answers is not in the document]
Schaeff: We use examples as test cases for the parser https://github.com/Zokrates/ZoKrates/blob/master/zokrates_parser/src/lib.rs#L26
Maybe you mean a test in `from_ast`? Or an e2e?

zokrates_core/src/absy/from_ast.rs
```diff
@@ -535,7 +549,7 @@ mod tests {
 
     #[test]
```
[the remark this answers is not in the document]
Schaeff: Agreed!

zokrates_parser/src/zokrates.pest
Schaeff: Agreed

zokrates_cli/examples/boolean_literal.code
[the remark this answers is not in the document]
JacobEberhardt: e2e, yes, sorry. However, considering that it's a parser level thing, this would not add any value, so good as is.

zokrates_pest_ast/src/lib.rs
Schaeff: I'll change to `BooleanLiteral` but keep `DecimalNumber` as at this point we could have different types of constants for the same type (say decimal and hex)

PR #402, opened 2019-06-22
[WIP] Implement boolean arrays

Schaeff:
Refactor to enable arrays of anything, not just field elements. We implement boolean arrays here.


Schaeff:
see #434

PR #407, opened 2019-06-24
Implement github resolver

### LINKED ISSUE #111: Implement a github import resolver
Once #110 is merged, it will be really easy to implement other import resolvers. That means that:
```
import github.com/my/rep/hash.code as my_hash
```
becomes possible.

Create a new crate similar to `zokrates_fs_resolver` which resolves Github files.

eupn:
Closes #111.

Github import syntax is the following:
```zokrates
import "github.com/<root>/<repo>/<branch>/<path/to/file>"
```

Where:
- `<root>` is the user or organization name
- `<repo>` is the repository name
- `<branch>` is the branch/snapshot name, e.g. `master`
- `<path/to/file>` is the absolute path to file in the specified branch of the repository

Example that uses github import:
```
./zokrates compile --input zokrates_cli/examples/imports/import_github.code
```


zokrates_github_resolver/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+) -> Result<PathBuf, io::Error> {
+    let url = format!(
+        "https://raw.githubusercontent.com/{root}/{repo}/{branch}/{path}",
+        root = root,
+        repo = repo,
+        branch = branch,
+        path = path
+    );
+
+    Ok(download_url(&url)?)
```
Schaeff: I think this is equivalent to `download_url(&url)`

zokrates_github_resolver/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+/// This importer needs to be provided with location since relative paths could be used inside the
+/// files that are imported from github.
+pub fn resolve(
+    location: &Option<String>,
+    path: &String,
+) -> Result<(BufReader<File>, String, String), io::Error> {
+    if let Some(location) = location {
+        let (root, repo, branch, path) = parse_input_path(&path)?;
+
+        let pb = download_from_github(&root, &repo, &branch, &path).unwrap();
```
Schaeff: As you do error handling in `download_from_github`, it would be nice to return the error here instead of panicking

zokrates_github_resolver/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/// Parses github import syntax: "github:<root>/<repo>/<branch>/<path/to/file>"
+/// Where:
+/// - <root> is the user or organization name
+/// - <repo> is the repository name
+/// - <branch> is the branch/snapshot name, e.g. `master`
+/// - <path/to/file> is the absolute path to file in the specified branch of the repository
+fn parse_input_path(path: &str) -> Result<(String, String, String, String), io::Error> {
+    let path = path.replacen("github:", "", 1);
```
Schaeff: You would use `GITHUB_IMPORT_PREFIX`

zokrates_github_resolver/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+        ));
+    }
+
+    let components = path.split("/").collect::<Vec<_>>();
+
+    // Check that root, repo, branch & path are specified
+    if components.len() < 4 {
+        return Err(io::Error::new(
+            io::ErrorKind::Other,
+            "Invalid github import syntax. Should be: github:<root>/<repo>/<branch>/<path>",
```
Schaeff: `GITHUB_IMPORT_PREFIX`

zokrates_github_resolver/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+    Ok(pb)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    pub fn import_simple() {
+        let res = parse_input_path("github:eupn/test/master/examples/imports.code").unwrap();
```
Schaeff: I think we should have the tests point to the ZoKrates repo here.

We also need to test `resolve` which requires mocking github for determinism and to avoid requiring an internet connexion to run the tests. Maybe we could use [mockito](https://docs.rs/mockito/0.17.1/mockito/) 

Any thoughts?

eupn: Redirected everything to ZoKrates repo. I can try to gate some kind of `reqwest` mock via `#[cfg(test)]` though.

Schaeff: That would be very nice!

eupn:
@Schaeff I have a suggestion to change the `GITHUB_IMPORT_PREFIX` from `github:` to `github.com/` as proposed in #111. Is it better or the current prefix is good enough?

Schaeff:
@eupn agreed, that would make sense

eupn:
@Schaeff ready to go!

zokrates_github_resolver/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+
+        Ok((br, location.to_owned(), alias))
+    } else {
+        Err(io::Error::new(io::ErrorKind::Other, "No location provided"))
+    }
+}
+
+/// Initializes github.com mocks for `import_github` example to run from `examples` tests.
+/// Note that returned mock objects should be alive prior to github requests.
+fn init_github_mock(location: &str) -> (Mock, Mock) {
```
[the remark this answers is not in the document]
eupn: I tried it first, but then I figured out that it only works for tests that are running in `zokrates_github_resolver` crate. But when you run tests for `zokrates_cli` crate, said resolver crate is built without `#[cfg(test)]` flag. You can try it yourself. That’s the reason why I omitted this flag from definitions and use mockito crate in non-dev dependencies. Don’t know how to circumvent this issue though.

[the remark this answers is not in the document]
Schaeff: I had a thought about it. I think mocks should be generated by each test. So if we were testing anything outside `zokrates_github_resolver`, we would have to pass a mock server to the resolver. For now, I reckon we should test only in `zokrates_github_resolver`. I made a PR on your branch with an idea of what it could look like.

Issue #394, opened 2019-06-19
Exponentiation with variable exponent


JacobEberhardt:
We could add this to the language directly or as a stdlib function. It requires bit decomposition of the exponent, which is expensive.
If we found an elegant way of warning the user regarding the expected expense, it could also be nice to have it directly as DSL-operation

PR #411, opened 2019-06-25
Exponentiation by squaring

### LINKED ISSUE #391: Use exponentiation-by-squaring for power flattening
We currently generate O(n) constraints to compute x**n (n constant)

```
// x**7
x_0 = x * x
x_1 = x_0 * x
x_2 = x_1 * x
x_3 = x_2 * x
x_4 = x_3 * x
x_5 = x_4 * x
```
Instead, we could do it in O(log(n)) with exponentiation by squaring
```
// x**7 = x * (x**2) * (x**4)
// first build required powers
x_0 = x * x // x**2
x_1 = x_0 * x_0 // x**4
// then multiply them
x_2 = x * x_0 // x**3
x_3 = x_2 * x_1 // x**7
```

https://en.wikipedia.org/wiki/Exponentiation_by_squaring

Schaeff:
Closes #391

### DIFF
--- a/zokrates_core/src/flatten/mod.rs
+++ b/zokrates_core/src/flatten/mod.rs
@@ -658,39 +658,67 @@ impl<'ast> Flattener<'ast> {
                         // TODO change that
                         assert!(base_flattened.is_linear());
 
-                        match e {
-                            e if *e == T::zero() => FlatExpression::Number(T::one()),
-                            // flatten(base ** 1) == flatten(base)
-                            e if *e == T::one() => base_flattened,
-                            // flatten(base ** 2) == flatten(base) * flatten(base)
-                            // in this case, no need to define an intermediate variable
-                            // as if a is linear, a ** 2 quadratic
-                            e if *e == T::from(2) => {
-                                FlatExpression::Mult(box base_flattened.clone(), box base_flattened)
-                            }
-                            // flatten(base ** n) = flatten(base) * flatten(base ** (n-1))
-                            e => {
-                                // flatten(base ** (n-1))
-                                let tmp_expression = self.flatten_field_expression(
-                                    functions_flattened,
-                                    statements_flattened,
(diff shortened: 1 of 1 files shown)


stefandeml:
```
def main(field a) -> (field):                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
  //field a = 2                                                                                                                                                                                                                                                                        
  field b = a**8                                                                                                                                                                                                                                                                       
  field c = a**6                                                                                                                                                                                                                                                                       
  return c 
```
yields the following flattened code:

```

Compiled program:
def main(_0) -> (1):
        (1 * _0) * (1 * _0) == 1 * _1
        (1 * _1) * (1 * _1) == 1 * _2
        (1 * _2) * (1 * _2) == 1 * _3
        (1 * _0) * (1 * _0) == 1 * _6
        (1 * _6) * (1 * _6) == 1 * _7
        (1 * _6) * (1 * _7) == 1 * _9
        (1 * ~one) * (1 * _9) == 1 * ~out_0
         return ~out_0
```

Schaeff:
Thanks for the comments @stefandeml !
- I think this links to two things: adding error management to the static analyser: right now we consider that anything that passes semantic checks won't fail, but for exponents we only know after constant propagation if there was a user error, so we should probably return an Err and not panic. We could add an issue for this. The second one is documentation: how should we document the operations allowed for `field`? I'm thinking this could be done in https://zokrates.github.io/concepts/types.html#field for now, any thoughts on that? We could open an issue there too.
- It could make sense, to be discussed in a different issue? Feel free to open if you think it's worth it.
- I'd love to explore that in a different issue too, I guess this starts require more global knowledge, keeping track of what powers of the base were already computed

JacobEberhardt:
I already opened an issue addressing your second point @stefandeml:
https://github.com/Zokrates/ZoKrates/issues/394
Feel free to discuss there.

PR #415, opened 2019-06-29
Update print-proof cli function

petscheit:
A new remix version was rolled out a couple of weeks ago, now tupels can be passed to functions in remix (abiv2 contracts). Adds a new argument to select proper format.


Schaeff:
Nice. Would it make sense to have this flag default to `v1` as `v2`, though supported by the new remix , is still experimental?

petscheit:
I wasn't sure about the default value. Had a default set to v1 in the beginning, then thought that might be confusing. On the other hand it would be consistent with exporting the verifier, since the default there is also v1. Adding the experimental argument, probably make more sense to have it as a default.

PR #416, opened 2019-06-30
ZkInterface adapter

naure:
This is a [ZkInterface](https://github.com/QED-it/zkinterface) adapter to use ZoKrates programs with other proving systems.

1. Using ZoKrates as a library:

See [live demo](https://qed-it.github.io/zkinterface-wasm-demo/).

See [example code](https://github.com/QED-it/zkinterface-wasm/blob/master/zkif-wasm-bundles/zkif-wasm-zokrates/src/frontend.rs).

2. Using with the CLI --proving-scheme:

```
zokrates setup          -s zkinterface -p r1cs.zkif
zokrates generate-proof -s zkinterface -j witness.zkif
```

Then, the .zkif files can be loaded by compatible proving systems.

Other points:

* Support for Curve25519.
To use Dalek's Bulletproof, ZoKrates needs to work with the scalar field of Curve25519.
Thibaut suggested to use a trait and a runtime parameter. However for now
we went with the minimal change which is a compile-time cargo feature `curve25519`.
In ZkInterface, there is a property `field_maximum` which allows the receiving
proving system to check that a compatible field was used.

* There is a test_zkinterface().
In there it is visible exactly what data is generated from an example program.
It involves the example value `-1` as a way to verify the correct usage of the scalar field.

* This is a joint work with @tronch0 and the team at qedit.


Schaeff:
Hey @naure, thanks a lot for this, it looks good! Here is our plan:
- We're implementing generic fields in ZoKrates, see #425 for an MVP. Still some work to do on that, especially on the cpp side.
- We need to spec out how people would use these features. In particular, we may want to change the target to something like `--curve curve25519 --proof-system bulletproof` rather than the interface. Thoughts on this welcome!

We're hoping to have the design done by the end of the month, and implementation depending on priorities later, as this is quite a substantial change of API in ZoKrates.

Schaeff:
Hi @naure [this](https://notes.ethereum.org/oLOhcn4oT32B55xGfQ9ymQ?both) is our draft for the backend API. We have a binary format to persist curve information to the programs so you can't "interpret" a bn128 program as a bls12 one.
Thoughts welcome here!

PR #417, opened 2019-07-01
Put resolvers behind feature in zokrates_cli

Schaeff:
Make it possible to build `zokrates_cli` with fs resolver only.

### DIFF
--- a/zokrates_cli/Cargo.toml
+++ b/zokrates_cli/Cargo.toml
@@ -6,9 +6,11 @@ repository = "https://github.com/JacobEberhardt/ZoKrates.git"
 edition = "2018"
 
 [features]
-default = []
+default = ["fs", "github"]
 libsnark = ["zokrates_core/libsnark"]
 wasm = ["zokrates_core/wasm"]
+github = ["zokrates_github_resolver"]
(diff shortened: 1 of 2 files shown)


zokrates_cli/Cargo.toml
```diff
@@ -6,18 +6,20 @@ repository = "https://github.com/JacobEberhardt/ZoKrates.git"
...
 
 [dependencies]
 clap = "2.26.2"
 bincode = "0.8.0"
 regex = "0.2"
 zokrates_field = { version = "0.3", path = "../zokrates_field" }
 zokrates_core = { version = "0.3", path = "../zokrates_core" }
-zokrates_fs_resolver = { version = "0.4", path = "../zokrates_fs_resolver"}
-zokrates_github_resolver = { version = "0.1", path = "../zokrates_github_resolver"}
+zokrates_fs_resolver = { version = "0.4", path = "../zokrates_fs_resolver", optional = true}
```
JacobEberhardt: Shouldn't the fs resolver be manadatory?

JacobEberhardt: Ok to hide in a feature since it's only mandatory for cli.

PR #427, opened 2019-07-09
Any expression as base of exponentiation

JacobEberhardt:
Before the parser rewrite, we only allowed variables or constants as base for fixed-exponent exponentiations.
This restriction is no longer required.


zokrates_cli/tests/code/arithmetics.json
```diff
@@ -0,0 +1,4 @@
+[
```
Schaeff: should be renamed to `arithmetics.arguments.json`

Issue #389, opened 2019-06-17
Implement no-output circuits


JustinDrake:
Boolean functions should be implementable without any "output" variables, i.e. using only assertions. Currently ZoKrates requires having one output variable for the final `return`.

JacobEberhardt:
You can simply `return` without providing an expression.
This implies 0 outputs and no constraints.

Schaeff:
We need to add tests before we close this

Schaeff:
Done, closing!

PR #428, opened 2019-07-09
Added test for empty return

JacobEberhardt:
Adds a test case for using `return` without subsequent expression in order to achieve a no-output ZoKrates program.

Addresses #389


Schaeff:
no_return is given 1 arg but expects 2

PR #434, opened 2019-07-16
Arrays of anything

Schaeff:
extension of #402 to any array content


zokrates_core/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![feature(box_patterns, box_syntax)]
+#![feature(box_patterns, box_syntax, associated_type_bounds)]
```
Schaeff: Maybe there's a way to do without this. It's needed for the `Flatten` trait in `flatten.rs` so that we can have the bound `TryFrom<TypedExpression<'ast, T>, Error: std::fmt::Debug>`

zokrates_core/src/flatten/mod.rs
```diff
@@ -43,6 +106,258 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+                        box FlatExpression::from(*t0_id),
+                        box FlatExpression::from(t1_id),
+                    ),
+                )
+            },
+        ));
+        res.into_iter().map(|r| r.into()).collect()
+    }
+
+    /// Flatten a array selection expression
```
JacobEberhardt: "an", if you wanna be nitpicky ;)

zokrates_core/src/flatten/mod.rs
```diff
@@ -43,6 +106,258 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `symbols` - Available functions in in this context
```
JacobEberhardt: remove one "in"

zokrates_core/src/flatten/mod.rs
```diff
@@ -661,251 +1017,110 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
-    /// Flattens a field array expression
+    /// Flattens an array expression
     ///
     /// # Arguments
     ///
     /// * `symbols` - Available functions in in this context
     /// * `statements_flattened` - Vector where new flattened statements can be added.
-    /// * `expression` - `FieldElementArrayExpression` that will be flattened.
-    fn flatten_field_array_expression(
+    /// * `expression` - `ArrayExpression` that will be flattened.
```
JacobEberhardt: should be 'expr'

zokrates_core/src/lib.rs
stefandeml: At this point there is no real need to have the bound generic right?

Shouldn't this also work:
```
rait Flatten<'ast, T: Field>
    : TryFrom<TypedExpression<'ast, T>, Error=()> 
    + IfElse<'ast, T> 
    + Select<'ast, T> {
```
And then there is no need for `associated_type_bounds`

stefandeml:
Just to persist this here as well:

My main take on that PR for now is that
a) we need update the docs with the new features added to the DSL
b) make sure that we have more tests written in DSL for these features, like we do in the stdlib. However to me this should be done in a separate PR.

So after fixing the docs and potentially fixing some of the typos, this should be merged.

zokrates_core/src/lib.rs
Schaeff: Oh right, we do know the associated type here! Thanks :)

Downside was an experimental feature @JacobEberhardt

zokrates_core/src/flatten/mod.rs
Schaeff: wanna try the github "suggest change" feature? :p

JacobEberhardt:
In the current version, how we process indices of multidimensional array declarations is unlike in C/Java. See the following example:
```
def main() -> (field):
    // valid ZoKrates
    field[3][2] a = [[1, 2, 3],[4, 5, 6]]
    // Interpretation array of three elements packed in array of two elements

    // allowed access a[0..2][0..3] -> not [0..3][0..2] as you may expect from declaration

    // not working:
    // C/Java world: Array of two elements of array of 3 elements
    field[2][3] b = [[1, 2, 3],[4, 5, 6]]
    b[1] // should be [1, 2, 3]


    // allowed access [0..2][0..3]
    return a[1][2]
```
Beyond being "established", the C/Java version is also aligned with the intuition of declaring and manipulating matrices. Hence, I would propose to change the implementation to mirror that behaviour.

PR #435, opened 2019-07-17
.code extension set internally in fs_resolver

petscheit:
.code file imports are now resolved without file extension in import statement/cli command

All file extensions removed from imports in .code files + book entries have been changed


zokrates_cli/src/bin.rs
```diff
@@ -260,9 +260,11 @@ fn cli() -> Result<(), String> {
         ("compile", Some(sub_matches)) => {
             println!("Compiling {}\n", sub_matches.value_of("input").unwrap());
 
-            let path = PathBuf::from(sub_matches.value_of("input").unwrap());
+            let mut path = PathBuf::from(sub_matches.value_of("input").unwrap());
+            path.set_extension("code");
```
Schaeff: I think we should replicate python behavior here: when passing a file to the python CLI, you do give the extension afaik
```
python script      
python: can't open file 'script': [Errno 2] No such file or directory
```

Schaeff: so I don't think there should be magic happening in the cli, only the import statements. Thoughts welcome though!

petscheit: Ok. I mean currently you can pass the extension, but dont have to. Let me know how I should change it

Schaeff: If I read it correctly, if you pass an extension, it overrides it with `.code`. I think at the CLI level, the extension should always be passed, as it is now

zokrates_fs_resolver/src/lib.rs
```diff
@@ -20,11 +20,13 @@ fn resolve_with_location(
     location: &String,
     source: &String,
 ) -> Result<(BufReader<File>, String, String), io::Error> {
-    let source = PathBuf::from(source);
+    let mut source = PathBuf::from(source);
+
+    source.set_extension("code");
```
Schaeff: I think this should happen earlier, where the resolver is called: the fact that we interpret `foo` as `foo.code` is not specific to `fs_resolver` so the `source` argument should already have the extension here. The resolver only fetches the file.

zokrates_cli/src/bin.rs
petscheit: Ok, will revert those changes

zokrates_fs_resolver/src/lib.rs
petscheit: I moved the extension addition to `zokrates_core/src/imports.rs`, much nicer this way. Also restored the tests in the `fs_resolver` which is a lot cleaner. 
Two remaining questions:
1. For the imports of github files, do we want to keep the extensions there or rather not? Currently they are not needed, so the format is the same as importing a local file
2. Tests are now not working because the `github_resolver` test gets its testing file from the master branch, which still contains the file extension. Should I leave it like this or change the imports, so the file extension can be passed, but doesn't have to.

Issue #261, opened 2019-02-08
Bundle computation and proof generation in one CLI command


Schaeff:
Although in some cases witness generation would happen at a different place proving does, most users probably compute + prove by default, the witness being irrelevant here. This could be the default.

Thoughts welcome

JacobEberhardt:
Witness generation is a key aspect of development and debugging of ZoKrates code. As proving can become really time-consuming for larger circuits, having these operations separately i required in my opinion.
However, an additional command that does both could be added.

Schaeff:
So this could either 
- be a new command
- a flag on `compute-witness`. I'd support renaming `compute-witness` to `run` or `compute` to make it friendlier
-`prove` could be the current `generate-proof` and accept a program and inputs just like `compute-witness`
- other ideas?

PR #437, opened 2019-07-19
WIP: Adding compile and verify script/minor readme update

edisinovcic:
Adding just a script for getting started even easier, the place of the script can be changed. 

In which branch should I merge this one?


Schaeff:
Hi @edisinovcic thanks the suggestion!
If I understand correctly, this is specific to IntelliJ?
It's clear we could do a better job with the `examples` folder but I'm not sure a script that encapsulates the whole pipeline is what we need. Maybe you can share what you had trouble with and we can figure out what to improve?

edisinovcic:
No, it's vscode specific, it's just for debugging scripts, so it's not intended to be used inside images. It can be used for debugging bash scripts for example compile and validate that I've created for the whole workflow. 

Example folder is there but definitely, some other place should be better to store the script and files, for example inside zokrates_cli/examples. What do you think?

edisinovcic:
Also there is a bug with latest version of rustc and that's why integration tests are failing.

Schaeff:
> Also there is a bug with latest version of rustc and that's why integration tests are failing. The solution is to lower the version of the compiler used or to update the code. Problem is that there are only two version on dockerhub which was updated 19 hours ago.
> 
> https://hub.docker.com/r/rustlang/rust/tags

This was fixed on develop, can you make your change target that?

> No, it's vscode specific, it's just for debugging scripts, so it's not intended to be used inside images. It can be used for debugging bash scripts for example compile and validate that I've created for the whole workflow.

I'm skeptical about adding something specific to an IDE and I'm struggling to understand what problem this is solving.
In general I think that any changes that improve onboarding for users should be done in `zokrates_book` by default.

edisinovcic:
> This was fixed on develop, can you make your change target that?
Will do.
> I'm skeptical about adding something specific to an IDE and I'm struggling to understand what problem this is solving.
> In general I think that any changes that improve onboarding for users should be done in `zokrates_book` by default.
Okay, the idea was having script that does all the stuff and you send all the arguments in one line. Currently it's pretty good so it's maybe overengineered (no need for it).

Schaeff:
I see, but why would you want to do everything in one go? ie usually you compile once, do a setup once, and run and prove many times. We have an open issue to combine run and prove here #261 but not agreed yet.

edisinovcic:
True that. Then I will close this one.

PR #445, opened 2019-07-30
Add precedence tests

petscheit:
Adds a new test, that checks if the precedence rules of operators are met. I believe I covered all scenarios, but I can't tell for sure. Researched a bit, but wasn't able to find a method of testing this, where we can be sure that every edge case is covered. Open for any proposals regarding this


Schaeff:
This looks like it goes in the right direction! Let's try and get this in. Can you merge develop into this branch?

zokrates_cli/tests/code/precedence.expected.witness
```diff
@@ -0,0 +1 @@
+~out_0 1
```
Schaeff: There should be a `precedence.arguments.json` file with the arguments passed to the function, here `[]` I guess.

zokrates_cli/tests/code/precedence.code
```diff
@@ -1,16 +1,17 @@
-def main() -> (field):
-  9 == 1 + 2 * 2 ** 2         // Checks precedence of arithmetic operators (expecting transitiv behaviour)
+def main() -> ():
+  9 == 1 + 2 * 2 ** 2                                              // Checks precedence of arithmetic operators (expecting transitiv behaviour)
   9 == 2 ** 2 * 2 + 1
   7 == 2 ** 2 * 2 - 1
   3 == 1 + 2 / 2 ** 2
```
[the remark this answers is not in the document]
petscheit: Sorry about that, must have messed up while copying from my local testing file to this file.

zokrates_cli/tests/code/precedence.expected.witness
petscheit: I see, wasn't able to test the running testing integration as it wasn't working. Regarding the expected.witness: Left the file empty since the cli returns nothing for the witness when running this

Schaeff:
Hey @petscheit I believe this is still lacking a `precedence.arguments.json` file?

PR #448, opened 2019-08-07
Implement structures

Schaeff:
Implement structures in ZoKrates:
```
from "./bar.code" import Bar

struct Foo {
   Bar bar
   field[3] baz
}

def main(private Foo f) -> (Bar):
   return f.bar
```

## Changes
- New explicit syntax for imports: `from PATH import SYMBOL as ALIAS`, we have backward compatibility by aliasing `import PATH` to `from PATH import main as PATH.FILENAME`
- structs can be declared, defined, accessed, mutated


t.code
```diff
@@ -0,0 +1,16 @@
+from "./u.code" import Fooo
+
+struct Bar {
+	a: field,
+	a: field,
```
JacobEberhardt: Should be 'b'

zokrates_core/src/typed_absy/variable.rs
```diff
@@ -27,6 +26,10 @@ impl<'ast> Variable<'ast> {
         Self::with_id_and_type(id, Type::array(ty, size))
     }
 
+    pub fn struc(id: Identifier<'ast>, ty: Vec<(MemberId, Type)>) -> Variable<'ast> {
```
stefandeml:
```suggestion
    pub fn struct(id: Identifier<'ast>, ty: Vec<(MemberId, Type)>) -> Variable<'ast> {
```

zokrates_core/src/static_analysis/propagation.rs
```diff
@@ -302,10 +327,124 @@ impl<'ast, T: Field> Folder<'ast, T> for Propagator<'ast, T> {
...
+    fn fold_struct_expression_inner(
+        &mut self,
+        ty: &Vec<(MemberId, Type)>,
+        e: StructExpressionInner<'ast, T>,
+    ) -> StructExpressionInner<'ast, T> {
+        match e {
+            StructExpressionInner::Identifier(id) => {
+                match self
+                    .constants
+                    .get(&TypedAssignee::Identifier(Variable::struc(
```
stefandeml:
```suggestion
                    .get(&TypedAssignee::Identifier(Variable::struct(
```

zokrates_core/src/static_analysis/constrain_inputs.rs
```diff
@@ -0,0 +1,120 @@
+//! Add runtime boolean checks on user inputs
```
stefandeml: That means we now check for all Boolean inputs strictly.

Eg in that example code:
```go
struct A {
        bool b
}

def main(A a) -> (field):
  field dummy = 12
  return dummy
```

we will add boolean checks for the input `a`.

zokrates_core/src/static_analysis/propagation.rs
Schaeff: `struct` is reserved in Rust, cannot be used for a function name

zokrates_core/src/typed_absy/variable.rs
Schaeff: struct is a keyword

PR #462, opened 2019-08-19
Add spreads and slices to typed ast

Schaeff:
right now `[...[0, 1, 2]]` gets checked to `[...[0, 1, 2][0], ...[0, 1, 2][1], ...[0, 1, 2][2]]` which explodes. Instead, preserve the notion of slice and spread all the way until flattening.


Schaeff:
Closed in favor of #467

PR #474, opened 2019-09-17
Introduce zokrates_test crate

Schaeff:


JacobEberhardt:
We should replace the `bench`directory with a more accurate name, since there is no profiling taking place.

JacobEberhardt:
Looks good, but doc needs to be updated.

JacobEberhardt:
Edited the documentation. Would appreciate a quick review.

zokrates_book/src/reference/Testing.md
```diff
@@ -1,6 +1,17 @@
 # Testing
 
+Before running any tests, make sure make sure your `ZOKRATES_HOME` environment variable is set correctly.
+It has to point to `zokrates_stdlib/stdlib/`
+
 ## Unit tests
+In ZoKrates, unit tests comprise
+- internal tests for all zokrates crates
+- compilation tests for all examples in the `$ZOKRATES_HOME/` zokrates_cli example folder. These tests only ensure that the examples compile correctly.
```
Schaeff: You're talking about a different `ZOKRATES_HOME` here (root of the project).
```suggestion
- compilation tests for all examples in the `zokrates_cli/examples`. These tests only ensure that the examples compile.
```

JacobEberhardt: Great, good catch!

PR #475, opened 2019-09-24
Explicitly specify Solidity version for SOLIDITY_G2_ADDITION_LIB

fulldecent:
This makes the code safer if it will be used separately from the other code


JacobEberhardt:
Thanks for the suggestion! Would be great if you could have the PR target develop instead of master. The default is a bit unfortunate, here.

fulldecent:
Got it, merge base updated.

fulldecent:
Please see that build error is not related to this change

PR #480, opened 2019-09-27
Add ABI component

Schaeff:


zokrates_cli/src/bin.rs
```diff
@@ -337,52 +347,81 @@ fn cli() -> Result<(), String> {
                 println!("{}", program_ast);
             }
 
-            let expected_cli_args_count =
-                program_ast.public_arguments_count() + program_ast.private_arguments_count();
+            let signature = program_ast.signature.clone();
```
JacobEberhardt: We should change the `program_ast` variable's name into something more accurate.
I propose `program_ir` or `interpretable_program`.

PR #487, opened 2019-10-02
Adds Precedence Test

petscheit:
Adds a new test, that checks if the precedence rules of operators are met. I believe I covered all scenarios, but I can't tell for sure. Researched a bit, but wasn't able to find a method of testing this, where we can be sure that every edge case is covered. Open for any proposals regarding this


zokrates_core_test/tests/tests/precedence.zok
```diff
@@ -0,0 +1,16 @@
...
+  9 == 2 ** 2 * 2 + 1
+  7 == 2 ** 2 * 2 - 1
+  3 == 2 ** 2 / 2 + 1
+
+  field a = if 3 == 2 ** 2 / 2 + 1 && true then 1 else 0 fi        // combines arithmetic with boolean operators
+  field b = if 3 == 3 && 4 < 5 then 1 else 0 fi                    // checks precedence of boolean operators
+  field c = if 4 < 5 && 3 == 3 then 1 else 0 fi
+  field d = if 4 > 5 && 2 >= 1 || 1 == 1 then 1 else 0 fi
+  field e = if 2 >= 1 && 4 > 5 || 1 == 1 then 1 else 0 fi
+  field f = if 1 < 2 && false || 4 < 5 && 2 >= 1then 1 else 0 fi
```
Schaeff: Interesting that this compiles but for style we should change this.
```suggestion
  field f = if 1 < 2 && false || 4 < 5 && 2 >= 1 then 1 else 0 fi
```

Issue #483, opened 2019-10-01
Add equality operator for booelans


stefandeml:
this works:
`field out = if sizeUnknown  then 1 else 0 fi`

but this one fails:
`field out = if sizeUnknown  == true then 1 else 0 fi`

with the following error:

> Cannot compare sizeUnknown of type bool to true of type bool

garvita-jain:
Hi! I would like to work on this project. Can you guide me about where to start since I am fairly new to Rust.
Thanks :)

Schaeff:
Hey @garvita-jain, @petscheit is already on this issue but do contact us on Gitter so we can find something you'd be interested to help with!  🎉

Schaeff:
Addressed in #492

PR #489, opened 2019-10-02
adds equality operator for booleans

petscheit:
implements #483


zokrates_core/src/flatten/mod.rs
```diff
@@ -538,6 +538,11 @@ impl<'ast, T: Field> Flattener<'ast, T> {
 
                 FlatExpression::Identifier(sub_bits_be[bitwidth - 1])
             }
+            BooleanExpression::EqBool(box lhs, box rhs) => self.flatten_boolean_expression(
+                symbols,
+                statements_flattened,
+                BooleanExpression::And(box lhs.clone(), box rhs.clone()),
```
Schaeff: Isn't this incorrect? `false && false == false` but `(false == false) == true`
Also the cloning sounds unnecessary here.

One ok way to do this would be to use the same flattening as for `BooleanExpression::Eq` below, though with Booleans we have the guarantee that they flatten to either 0 or 1, so maybe this extra guarantee makes a cheaper check possible. The Field one is 2 constraints right now so it would need to be 1 constraint to be better I suppose?

zokrates_core/src/typed_absy/mod.rs
```diff
@@ -513,6 +513,10 @@ pub enum BooleanExpression<'ast, T: Field> {
         Box<FieldElementExpression<'ast, T>>,
         Box<FieldElementExpression<'ast, T>>,
     ),
+    EqBool(
```
Schaeff: I think for consistency, if we call this `EqBool` we should rename `Eq` to `EqField`.

I also have a preference for `BoolEq` like "boolean equality".

zokrates_core/src/flatten/mod.rs
petscheit: I agree, wasn't thought through enough from my side. 

 ```
    BooleanExpression::BoolEq(box lhs, box rhs) => self.flatten_boolean_expression(
                symbols,
                statements_flattened,
                BooleanExpression::Or(
                    box BooleanExpression::And(box lhs.clone(), box rhs.clone()), 
                    box BooleanExpression::And(
                        box BooleanExpression::Not(box lhs), 
                        box BooleanExpression::Not(box rhs)
                    ),
                )
            ),
```
Something like this should work functionally right, though I'm not sure how efficient this is. I don't think I understand the flattening part of this comment. What do you mean by the Field one?

Comment (Schaeff) [reply zokrates_core/src/flatten/mod.rs]: I meant "equality for field elements" as implemented in `flatten.rs` for `BooleanExpression::Eq`.

> I'm not sure how efficient this is

You could give it a try with a simple program:
```
def main(bool a, bool b) -> (bool):
   return a == b
```

PR #492, opened 2019-10-16
Bool eq

petscheit:
implements #489 
The bool compare is evaluated by doing: `Not((x-y)^2)` (x,y being the booleans)


zokrates_core/src/flatten/mod.rs
```diff
@@ -724,6 +724,33 @@ impl<'ast, T: Field> Flattener<'ast, T> {
 
                 FlatExpression::Identifier(sub_bits_be[bitwidth - 1])
             }
+            BooleanExpression::BoolEq(box lhs, box rhs) => {
+                // Wanted: Not((x-y)^2)
```
Schaeff: Can we put the full proof here? something like if lhs and rhs are booleans, they flatten to 1 or 0. Wanted truth table: ... . Truth table of Not((flat(x) - flat(y))**2): ... .

zokrates_core/src/flatten/mod.rs
```diff
@@ -724,6 +724,33 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+                let x = self.flatten_boolean_expression(symbols, statements_flattened, lhs);
+                let y = self.flatten_boolean_expression(symbols, statements_flattened, rhs);
+
+                let name_x_sub_y = self.use_sym();
```
Schaeff: `x - y` is linear so no need to introduce a new variable: `let x_sub_y = FlatExpression::Sub(box x, box y)`

zokrates_core/src/flatten/mod.rs
```diff
@@ -724,6 +724,33 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+                let name_x_mult_x = self.use_sym();
+
+                assert!(x.is_linear() && y.is_linear());
+                statements_flattened.push(FlatStatement::Definition(
+                    name_x_sub_y,
```
Schaeff: Then this can be removed

zokrates_core/src/flatten/mod.rs
```diff
@@ -724,6 +724,33 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+                    FlatExpression::Sub(box x, box y),
+                ));
+
+                statements_flattened.push(FlatStatement::Definition(
+                    name_x_mult_x,
+                    FlatExpression::Mult(
+                        box FlatExpression::Identifier(name_x_sub_y),
```
Schaeff: here we can use `x_sub_y`

zokrates_core_test/tests/tests/bool_compare.zok
```diff
@@ -0,0 +1,3 @@
+def main(field a, field b) -> (field):
```
Schaeff: Would it make sense to pass booleans as inputs and return booleans? It would make the code clearer

petscheit:
I've implemented the requested changes and also the renaming of BooleanExpression::Eq to BooleanExpression::FieldEq. 

Regarding the constant propagation, I'm not sure if what I did makes sense, since I don't understand what static_analysis is used for.

PR #505, opened 2019-11-05
Reordering and removing duplicates from keywords

edisinovcic:
Removing duplicate "in" keyword and ordering it alphabetically


Schaeff:
return is there twice too

edisinovcic:
Now return is also removed.

Schaeff:
LGTM, please open PRs against develop next time @edisinovcic :)

PR #506, opened 2019-11-05
fix cargo features in zokrates_abi

dark64:
**zokrates_abi** crate depends on **zokrates_core** which has "multicore" feature enabled by default. Multicore feature is not yet compatible with wasm target. This will remove the issue with targets not compatible with default "multicore" feature.


dark64:
Decided to remove default features from core to avoid doing this.

PR #511, opened 2019-11-12
Cli patch

dark64:
- Libsnark FFI
- CLI changes to support new ProofSystem API
- Removed default features in zokrates_core


zokrates_core/Cargo.toml
```diff
@@ -7,14 +7,14 @@ readme = "README.md"
...
-default = ["multicore"]
+default = []
 libsnark = ["cc", "cmake", "git2"]
 wasm_solvers = ["wasmi", "parity-wasm", "rustc-hex"]
 wasm = ["bellman_ce/wasm", "zokrates_embed/wasm"]
 multicore = ["bellman_ce/multicore"]
 
 [dependencies]
-libc = "0.2.0"
+# libc = "0.2.0"
```
Schaeff: can we remove this entirely?

dark64: Yes, should be removed.

Schaeff: Cool then happy to merge once that's changed, and
```
zokrates_core/src/flatten/mod.rs:1188:30
     |
1188 | ...                   .into_iter()
     |                        ^^^^^^^^^ help: use `.iter()` instead of `.into_iter()` to avoid ambiguity: `iter`
     |
```
is because we always use the latest nightly on CI, making that change should fix the build

Schaeff: will fix on wasm-friendly

PR #525, opened 2019-11-26
Add ace mode to parser crate

dark64:


dark64:
Updated license to LGPL

Schaeff:
Thanks, shall we have another PR adding missing keywords to the grammar?

PR #526, opened 2019-11-27
Generate ABI during compilation

Schaeff:
Instead of baking the high level signature into the ir program, generate a complete interface after semantic checking and return it alongside the ir program.

This is a breaking change, as the serialized program alone isn't sufficient to interact via ABI anymore.

Todo
- [x] Better naming for `ProgAndAbi`
- [x] Avoid tuples


zokrates_cli/src/bin.rs
```diff
@@ -51,6 +54,14 @@ fn cli() -> Result<(), String> {
             .value_name("FILE")
             .takes_value(true)
             .required(true)
+        ).arg(Arg::with_name("abi_spec")
+            .short("s")
+            .long("abi_spec")
+            .help("Path of the ABI specification")
```
JacobEberhardt: Very minor thing: The cli command descriptions should be uniform. 
Line 168 uses "Path to the ABI specification", here its "of".

JacobEberhardt: Another thing I noticed:
Until now, we explain `-o` as "Path of the output file". 
Technically, this is no longer correct. "Compiled program path" or something like this would be better.

JacobEberhardt: The export-verifier command has an "abi" input, which requires a solidity abi version, but does not explicitly refer to solidity, as this was clear before the ZoKrates ABI introduction in this PR. We need to make this description clearer, potential also rename the command to "solidity-abi" or something.

Schaeff: Very good feedback thanks

PR #530, opened 2019-12-06
This fixes macos installation issues as per: https://github.com/scipr…

Genysys:
…-lab/libsnark/issues/86

Please try this out on other targets to ensure that the omission of SuperCop doesn't break anything

### DIFF
--- a/zokrates_core/build.rs
+++ b/zokrates_core/build.rs
@@ -38,6 +38,7 @@ fn main() {
         // build libsnark
 
         let libsnark = cmake::Config::new(libsnark_source_path)
+            .define("WITH_SUPERCOP", "OFF")
             .define("WITH_PROCPS", "OFF")
             .define("CURVE", "ALT_BN128")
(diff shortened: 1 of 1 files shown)


Schaeff:
Thanks it's a good start. I think for this to be even more useful we should have complete instructions about requirements on macos, namely openssl and how to install it reliably.

Genysys:
Where do you think is best to add this? In the README.md or the documentation? I would say documentation under [HERE](https://zokrates.github.io/reference/proving_schemes.html#supporting-backends) as the README.md is nice and sweet and makes no references to the additional backends

Schaeff:
Yes I suggest this section should be updated to remove the Linux disclaimer, and instructions for boost and openssl (and anything else needed) should be added [here](https://zokrates.github.io/gettingstarted.html#from-source) by adding a Prerequisites section with instructions for Ubuntu and MacOS

PR #535, opened 2019-12-24
Fix typos in documentation

leonardoalt:


zokrates_book/src/sha256example.md
```diff
@@ -52,8 +52,8 @@ grep '~out' witness
 which should lead to the following output:
 
 ```sh
-~out_0 263561599766550617289250058199814760685
 ~out_1 65303172752238645975888084098459749904
+~out_0 263561599766550617289250058199814760685
```
leonardoalt: This is the order I got when running the command.

zokrates_book/src/concepts/imports.md
```diff
@@ -38,10 +38,10 @@ from "./path/to/my/module" import main as module
 Note that this legacy method is likely to be become deprecated, so it is recommended to use the preferred way instead.
 ### Symbols
 
-Two type of symbols can be imported
+Two type of symbols can be imported:
```
Schaeff:
```suggestion
Two types of symbol can be imported:
```

is this correct english

zokrates_book/src/sha256example.md
Schaeff: Interesting. We need to investigate the cause of this

zokrates_book/src/concepts/imports.md
leonardoalt: According to https://english.stackexchange.com/questions/5539/types-of-things-vs-types-of-thing
I think it should be `types of symbols`

zokrates_book/src/sha256example.md
leonardoalt: Do you get the original doc version if you run it? I can remove this change for now while we investigate.

zokrates_book/src/concepts/imports.md
Schaeff: fair enough
```suggestion
Two types of symbols can be imported:
```

leonardoalt: That's how it is right now

PR #544, opened 2020-01-09
added solidity v0.6 support

### LINKED ISSUE #543: Compilation error with solc 0.6.1
Hello,
I get a strange compilation error when I try to compile the generated `verifier.sol`.
When I compile it with solc, version 0.6.1 I get the following :
```
solc --bin verifier.sol
Error: Expected '(' but got ','
   --> verifier.sol:237:42:
    |
237 |             success := staticcall(sub(gas, 2000), 5, freemem, 0xC0, freemem, 0x20)
    |                                          ^
```
However i don't get the error when I downgrade the solc compiler to the version 0.5.11

leekt216:
closes #543 
from solidity v0.6, assembly `gas` has changed to `gas()`
simple fix is done by creating v0.6 templates and adding one more argument of export solidity verifier

ref:
https://solidity.readthedocs.io/en/v0.6.1/assembly.html

### DIFF
--- a/zokrates_cli/src/bin.rs
+++ b/zokrates_cli/src/bin.rs
@@ -50,6 +50,7 @@ fn cli() -> Result<(), String> {
     const JSON_PROOF_PATH: &str = "proof.json";
     let default_scheme = env::var("ZOKRATES_PROVING_SCHEME").unwrap_or(String::from("g16"));
     let default_solidity_abi = "v1";
+    let default_solidity_version = "0.5";
 
     // cli specification using clap library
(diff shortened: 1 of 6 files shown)


Schaeff:
Nice one! I'm wondering if we should keep supporting 0.5 at all, however?

leekt216:
I think it will be too heavy to support all those solidity versions.
If you want, i can change to support only 0.6.

Schaeff:
I had a go and opened #545 on this, would love to hear your thoughts there

leekt216:
seems i should close this pr

PR #545, opened 2020-01-09
Upgrade to solidity 0.6.1

Schaeff:


leekt216:
I think it's better to just use ABIEncoderV2 version and remove non-ABIEncoderV2 version. Since it's not experimental anymore from 0.6. https://github.com/ethereum/solidity/releases/tag/v0.6.0
Code will be a lot simpler in that way.

Schaeff:
Yeah I was hesitant to do that. The pragma statement is still required though, that's why I left the v1 ABI version too

PR #546, opened 2020-01-09
Enable libsnark by default

Schaeff:
- Add instructions for MacOS
- Remove `WITH_LIBSNARK` distinction

### DIFF
--- a/zokrates_book/src/gettingstarted.md
+++ b/zokrates_book/src/gettingstarted.md
@@ -31,6 +31,40 @@ cargo +nightly build --release
 cd target/release
 ```
 
+If you want to enable the libsnark backend in ZoKrates, you need to install some prerequisites:
+- boost
+- cmake
+- libssl
+
+#### Libsnark prerequisites - Linux
+
+As an example, this command installs the required dependencies on Ubuntu:
+
+```bash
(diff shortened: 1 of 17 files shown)


zokrates_book/src/gettingstarted.md
```diff
@@ -31,6 +31,40 @@ cargo +nightly build --release
 cd target/release
 ```
 
+If you want to enable the libsnark backend in ZoKrates, you need to install some prerequisites:
```
JacobEberhardt: This sounds optional, but since this PR enables libsnark compilation by default, we should probably mark these dependencies as mandatory? Alternatively, we could explicitly document how to build without libsnark, here.

JacobEberhardt: Furthermore, specifying minimum versions we tested with could be helpful. Not mandatory.

Schaeff: If we're confident enough that MacOS works, it would even make sense to remove the part about building without libsnark entirely? Or maybe it's still useful for other OS?

JacobEberhardt: good point, that sounds optimal. only need to document the optional "without-libsnark" build, then.

PR #548, opened 2020-01-14
Add CI steps for zokrates.js

dark64:
Minor code cleanup, added CI steps to circleci config and added release steps in `scripts/release.sh`

**Note**: changed bellman to point to [9e35737](https://github.com/matter-labs/bellman/commit/9e35737209d4afc2e69af1e0d7cbe6d02e32aecf) as there is a breaking change in latest bellman commit.


zokrates_js/src/lib.rs
```diff
@@ -21,82 +23,88 @@ pub struct CompilationResult {
     abi: String,
 }
 
+impl ResolverResult {
+    fn to_tuple(&self) -> (String, String) {
```
Schaeff: I suspect this could be `into_tuple` and not perform any clone?

dark64: yeah good catch

PR #550, opened 2020-01-20
Enable variable loop bounds

### LINKED ISSUE #520: Add variable 'For loop bounds' on nested For loops
Hi team! I was just exploring doing 'triangular' for loops, in order to ensure a list contains only unique values:

What I'd like to do is the following 'triangular' nested for loop. Notice how the second loop has lower bound `i`:

```
// check for duplicates
def main(field[256] values) -> ():
    field acc = 1
    for field i in 0..256 do
        for field j in i..256 do
            acc = acc * (values[j] - values[i])
            isNotZero(acc) == 1
        endfor
    endfor

    return
```

This (understandably) fails, because `i` is variable: `thread 'main' panicked at 'not yet implemented: For loop bounds should be constants, found i', zokrates_core/src/absy/from_ast.rs:288:18`

However, in the above code, it's clear that the nested loop's bounds aren't dynamic, because `i` isn't changed within the loop. Therefore, this ought to be allowable.

Are there plans to allow such syntax in future?

Schaeff:
closes #520


zokrates_core/src/static_analysis/propagate_unroll.rs
```diff
@@ -0,0 +1,209 @@
...
+//! @date 2018
+
+use static_analysis::propagation::Propagator;
+use static_analysis::unroll::Unroller;
+use typed_absy::TypedProgram;
+use zokrates_field::field::Field;
+
+pub struct PropagatedUnroller;
+
+const MAX_DEPTH: usize = 100;
```
JacobEberhardt: This seems a bit risky. Why is this sufficient?

JacobEberhardt: If we define a more or less arbitrary upper bound, it may be worth having a testcase that actually runs into this case.

PR #558, opened 2020-01-30
[wip] Optimize CI for zokrates.js

dark64:
**Some notes:**
Usually the build step includes bundling stdlib to a json file. This file represents the whole stdlib folder structure and is not really needed for building rust related code, but it needs to be bundled for everything to work properly (importing stdlib modules). This process requires both rust and node images to build everything. There are two possible solutions for this; create docker image manually that will contain all dependencies needed for whole process to work (rust nightly, wasm-pack, node and npm) and use that image in the CI step. Downside to this solution is that rust version is then static and needs to be updated manually on each new nightly version. Other solution is to use multi-stage builds inside docker but this takes a lot of time for builds (10min+) as there is no way to use circleci cache in this context as far as I know of. On each push it needs to download base images and install dependencies before starting the build process.

This PR uses the first solution proposed above. If proposed solution is accepted, we should push [environment image](https://github.com/dark64/ZoKrates/blob/ci-patch/zokrates_js/Dockerfile.env) to official zokrates dockerhub account.

### DIFF
--- a/zokrates_js/Dockerfile.env
+++ b/zokrates_js/Dockerfile.env
@@ -19,4 +19,4 @@ RUN set -eux; \
     cargo --version; \
     rustc --version;
 
-RUN curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh -s -- -y
+RUN cargo install wasm-pack --version 0.8.1
--- a/zokrates_js/Dockerfile.env
+++ b/zokrates_js/Dockerfile.env
@@ -1,0 +1,22 @@
+FROM node:10
+SHELL ["/bin/bash", "-c"]
+
+ENV RUSTUP_HOME=/usr/local/rustup \
+    CARGO_HOME=/usr/local/cargo \
+    PATH=/usr/local/cargo/bin:$PATH
+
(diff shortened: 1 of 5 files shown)


Schaeff:
I think it's fine to have a static Rust nightly version. Does that change the solution here?

zokrates_js/Dockerfile.env
```diff
@@ -0,0 +1,22 @@
...
+
+RUN set -eux; \
+    apt-get update; \
+    apt-get install -y --no-install-recommends \
+        ca-certificates \
+        gcc \
+        libc6-dev \
+        ; \
+    \
+    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path --default-toolchain nightly; \
```
Schaeff: would it make sense for the Dockerfile to have the nightly version as a parameter somewhere so we can control that? Right now it gets the latest nightly?

dark64: Yeah good point, I added build argument for specifying rust version (if argument is not specified it defaults to nightly)

PR #578, opened 2020-03-03
Memoize function calls

### LINKED ISSUE #560: Spreading a function directly into an array is very inefficient?
I've noticed that spreading a function directly into an array is inefficient (constraint-wise).

I give an example below, showing the method I initially attempted, and then the more efficient improvement:

Method 1: spread a function directly into an array:
```
import "utils/pack/nonStrictUnpack256.zok" as unpack256

def main(field[2] inputs) -> (field[512]):

    field[512] preimage512 = [...unpack256(inputs[0]), ...unpack256(inputs[1])]

    return preimage512
```

Method 2: split up the steps:
```
import "utils/pack/nonStrictUnpack256.zok" as unpack256

def main(field[2] inputs) -> (field[512]):

    field[2][256] inputsAsBits = [unpack256(inputs[0]), unpack256(inputs[1])]

    field[512] preimage512 = [...inputsAsBits[0], ...inputsAsBits[1]]

    return preimage512
```

Method 1 costs 130816 constraints, whereas Method 2 costs 766 constraints.

Is this expected behaviour?

I'm using `zokrates:0.5.1`

Schaeff:
```
def main(field a) -> (field):
    return foo(a) + foo(a) + foo(a)
```

We currently pay for 3 calls of `foo` even though the result is the same.
This PR adds memoization to calls at the inlining stage, so that the result of the first call is reused after.

This fixes #560 as `[...f()]` is rewritten as `[f()[0], ..., f()[41]]`


zokrates_core_test/tests/tests/memoize/memoize.zok
```diff
@@ -0,0 +1,14 @@
...
+def local(field a) -> (field): // this costs 3 constraints per call
+	field res = a ** 8
+	return res // currently expressions in the return statement don't get memoized
+
+def main(field a) -> ():
+	// calling a local function many times with the same arg should cost only once
+	local(a) + local(a) + local(a) + local(a) + local(a) == 5 * (a ** 8)
+
+	// calling an imported function many times with the same arg should cost only once
+	dep(a) + dep(a) + dep(a) + dep(a) + dep(a) == 4 * (a ** 4)
```
JacobEberhardt, line 12: dep(a) + dep(a) + dep(a) + dep(a) + dep(a) == 5*(a**4) != 4 * (a ** 4)

Schaeff: wow good catch

PR #580, opened 2020-03-03
Resolver patch

dark64:
Added [Resolver](https://github.com/dark64/ZoKrates/blob/resolver-patch/zokrates_core/src/compile.rs#L134) trait for code clarity


zokrates_core/src/compile.rs
```diff
@@ -129,20 +129,25 @@ impl fmt::Display for CompileErrorInner {
     }
 }
 
-// See zokrates_fs_resolver for the spec
-pub type Resolve<'a, E> = &'a dyn Fn(PathBuf, PathBuf) -> Result<(String, PathBuf), E>;
-
 type FilePath = PathBuf;
 
+pub trait Resolver<E: Into<imports::Error>> {
```
Schaeff: Can `<E: Into<imports::Error>>` go on `resolve` rather than on the trait? I think it's good practice and it might also make things clearer

zokrates_fs_resolver/Cargo.toml
```diff
@@ -3,13 +3,10 @@ name = "zokrates_fs_resolver"
 version = "0.5.0"
 authors = ["Thibaut Schaeffer <thibaut@schaeff.fr>"]
 repository = "https://github.com/JacobEberhardt/ZoKrates.git"
-
-[features]
-default = []
+edition = "2018"
 
 [dependencies]
+zokrates_core = { version = "0.4.2", path = "../zokrates_core" }
```
Schaeff: Is there a way to keep `fs_resolver` independent from `zokrates_core`? I guess having a different crate for that trait definition?

zokrates_fs_resolver/src/lib.rs
```diff
@@ -3,75 +3,63 @@ use std::io;
...
-    import_location: ImportLocation<'a>,
-) -> Result<(SourceCode, CurrentLocation), io::Error> {
-    let source = Path::new(&import_location);
-
-    if !current_location.is_file() {
-        return Err(io::Error::new(
-            io::ErrorKind::Other,
-            format!("{} was expected to be a file", current_location.display()),
-        ));
+pub struct FileSystemResolver {}
```
Schaeff:
```suggestion
pub struct FileSystemResolver;
```

zokrates_fs_resolver/Cargo.toml
dark64: We could introduce a "common" crate that contains the trait definition, and the trait should be more generic then.

Issue #554, opened 2020-01-22
Setup panics if private argument is unused


dark64:
## Steps to Reproduce
Given the following code
```
// main.zok
def main(private field a) -> (field):
    return 1
```
1. `./zokrates compile --input main.zok`
2. `RUST_BACKTRACE=1 ./zokrates setup`

Setup panics with the following stack backtrace:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: UnconstrainedVariable', src/libcore/result.rs:1192:5
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:77
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1057
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1426
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:195
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:215
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:463
  11: rust_begin_unwind
             at src/libstd/panicking.rs:371
  12: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
  13: core::result::unwrap_failed
             at src/libcore/result.rs:1192
  14: zokrates_core::proof_system::bn128::utils::bellman::Computation<zokrates_field::field::FieldPrime>::setup
  15: <zokrates_core::proof_system::bn128::g16::G16 as zokrates_core::proof_system::ProofSystem>::setup
  16: zokrates::cli
  17: zokrates::main
  18: std::rt::lang_start::{{closure}}
[...]

Issue #500, opened 2019-10-27
Error while running setup phase


sameersubudhi:
Hi,

I was trying to run the setup phase on a ZoKrates program to generate the prover/verifier keys and thus the verifier contract. The code accepts multiple input parameters, some of which are field arrays. When making the field arrays private, I encounter the below error. There are no issues when the input field array is public. ZoKrates versions used: **v0.4.10** and **v0.5.0**.

**Sample test codes:**
```
def main(private field[5] a) -> (field):
    return a[0]
```

**Commands:**
- `zokrates compile --light -i test.code` or `zokrates compile --light -i test.zok`
- `zokrates setup --light`

**Error stack (on version 0.4.10):**
```
Performing setup...
WARNING: You are using the G16 scheme which is subject to malleability. See zokrates.github.io/reference/proving_schemes.html#g16-malleability for implications.
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: UnconstrainedVariable', src/libcore/result.rs:1084:5
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:47
   3: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:36
   4: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:200
   5: std::panicking::default_hook
             at src/libstd/panicking.rs:214
   6: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:477
   7: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:384
   8: rust_begin_unwind
             at src/libstd/panicking.rs:311
   9: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
  10: core::result::unwrap_failed
             at src/libcore/result.rs:1084
  11: zokrates_core::proof_system::bn128::utils::bellman::Computation<zokrates_field::field::FieldPrime>::setup
  12: <zokrates_core::proof_system::bn128::g16::G16 as zokrates_core::proof_system::ProofSystem>::setup
[...]

PR #582, opened 2020-03-18
Add check for unused/unconstrained parameters

dark64:
Adds compile time check for issue https://github.com/Zokrates/ZoKrates/issues/554 and https://github.com/Zokrates/ZoKrates/issues/500


zokrates_core/src/static_analysis/unconstrained_vars.rs
```diff
@@ -0,0 +1,168 @@
...
+            main,
+        };
+
+        UnconstrainedVariableDetector::detect(p);
+    }
+
+    #[test]
+    fn should_pass_with_directive() {
+        // def main(private x) -> (1):
+        //   return if x == 42 then 1 else 0 fi
```
Schaeff: I think here it would make more sense to put the IR code as printed by fmt

dark64:
everything should be good now @Schaeff

PR #585, opened 2020-03-27
Mimc

petscheit:
Adds mimc hashing algo to the stdlib. The reference implementation this is based on can be found [here](https://github.com/iden3/circomlib/blob/master/circuits/mimc.circom)

**Things to note:**
- The reference implementation is slightly more efficient (regarding constraints). Because ZoKrates doesn't support lazy evaluation, the t7 array has to be one element longer and a separate index counter has to be defined in the loop to prevent an out of bounds error. It's not the prettiest solution, but it was the most efficient solution I could come up with.
- Implementation was tested with comparing hashes to the reference implementation, which were always the same, also with overflows, passing p and other edge cases.
- The number of rounds can be set up to 91. I wasn't sure how many rounds are optimal, so I created a number of them. 

**Constraints for hashing:**
- mimc10: 42
- mimc20: 82
- mimc50: 202
- mimc90: 362


petscheit:
Discussion and benchmarks for picking different numbers of rounds. https://github.com/HarryR/ethsnarks/issues/87

PR #589, opened 2020-04-02
Fix parser multidef confusion

### LINKED ISSUE #588: Parser misinterprets definition as multidef
```
field a = f() + 1
```

Since we want to catch all statements of the form `identifiers = function_call`, we try this rule first before a simple definition `assignee = expression`.

Sadly in this case it matches `field a = f()` and a parsing error is returned.

Schaeff:
fixes #588

### DIFF
--- a/zokrates_core/src/absy/from_ast.rs
+++ b/zokrates_core/src/absy/from_ast.rs
@@ -296,18 +296,6 @@ impl<'ast, T: Field> From<pest::IterationStatement<'ast>> for absy::StatementNod
     }
 }
 
-// impl<'ast, T: Field> From<pest::AssignmentStatement<'ast>> for absy::StatementNode<'ast, T> {
-//     fn from(statement: pest::AssignmentStatement<'ast>) -> absy::StatementNode<T> {
-//         use absy::NodeValue;
-
-//         absy::Statement::Definition(
-//             absy::AssigneeNode::from(statement.assignee),
-//             absy::ExpressionNode::from(statement.expression),
-//         )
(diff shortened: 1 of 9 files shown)


zokrates_core/src/absy/from_ast.rs
```diff
@@ -966,4 +966,147 @@ mod tests {
...
+    #[test]
+    fn declarations() {
+        use self::pest::Span;
+
+        let span = Span::new(&"", 0, 0).unwrap();
+
+        // For different definitions, we generate declarations
+        // Case 1: `id = expr` where `expr` is not a function call
+        // This is a simple assignment, doesn't implicitely declare a variable
+        // A `Definition` is generatedm and no `Declaration`s
```
JacobEberhardt, line 978:
```suggestion
        // A `Definition` is generated and no `Declaration`s
```

zokrates_core/src/absy/from_ast.rs
```diff
@@ -966,4 +966,147 @@ mod tests {
...
+
+        assert_eq!(statements.len(), 1);
+        match &statements[0].value {
+            absy::Statement::MultipleDefinition(..) => {}
+            s => {
+                panic!("should be a Definition, found {}", s);
+            }
+        };
+        // Case 3: `ids = expr` where `expr` is a function call
+        // This implicitely declares all variables which are type annotated
```
JacobEberhardt, line 1054: I would argue that the `type identifier`statement on the left side is an explicit declaration.

PR #593, opened 2020-04-10
MiMCSponge and MiMC7 implementation for stdlib

petscheit:
## MiMCSponge:
This implements the MiMCSponge hashing into the stdlib. The reference implementation this is based on, can be found [here](https://github.com/iden3/circomlib/blob/master/circuits/mimcsponge.circom). For testing I've created a quick repo that runs the reference implementation with circom, can be found [here](https://github.com/petscheit/MiMCReferenceTesting).

### Configuration:
I'm a bit unsure what the input and output configurations should be set to. Currently, like in the reference, two inputs and three outputs, with 220 feistel rounds are configured, but I'm not sure what the proper settings should be, so it's up for discussion.

### Performance:
Using this, with the configuration stated above, 2647 constraints are created

## MiMC7:
Adds mimc7 hashing algo to the stdlib. The reference implementation this is based on can be found [here](https://github.com/iden3/circomlib/blob/master/circuits/mimc.circom)

**Things to note:**
- The reference implementation is slightly more efficient (regarding constraints). Because ZoKrates doesn't support lazy evaluation, the t7 array has to be one element longer and a separate index counter has to be defined in the loop to prevent an out of bounds error. It's not the prettiest solution, but it was the most efficient solution I could come up with.
- Implementation was tested with comparing hashes to the reference implementation, which were always the same, also with overflows, passing p and other edge cases.
- The number of rounds can be set up to 91. I wasn't sure how many rounds are optimal, so I created a number of them. 

**Constraints for hashing:**
- mimc7R10: 42
- mimc7R20: 82
- mimc7R50: 202
- mimc7R90: 362


JacobEberhardt:
Great, thank you!
I think we could also add back the MiMC7 version you initially implemented, additionally.
Just need to make sure through comments they are not confused.

petscheit:
Currently, the MiMC7 implementation is on another branch. Should I add them to this branch or reopen the old PR?

JacobEberhardt:
We can make that one PR for the MiMC family

petscheit:
I've updated this PR accordingly

Schaeff:
The errors now are related to new explicit syntax (no implicit declaration on result of calls). Also if this is bn128 specific, it should have pragma statements.

petscheit:
Could this be a OOM error as described [here](https://github.com/rust-lang/rust/issues/51309)? Everything runs fine locally and the error message doesn't suggest an error in the zokrates code

Schaeff:
Right that's annoying. Maybe #604 helps, let's see

Schaeff:
Great addition to the stdlib!

PR #597, opened 2020-05-11
Make solidity verifier view

### LINKED ISSUE #594: Solidity Verifier Contract modifier can be made view
I noticed that with [EIP-214](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-214.md) the opcode `staticcall` was introduced which allows read-only calls to contracts.

The Solidity pairing library that is generated together with the verifier contract could replace `call` with `staticcall` for the pre-compiled elliptic curve operation contracts. This is turn allows the verification function to have a `view` modifier. The proof verification could then be called from a `view` function from another contract.

I don't see any compatibility issues with this change.

Schaeff:
Had to remove the event as it's a side effect.

Closes #594

### DIFF
--- a/zokrates_book/src/sha256example.md
+++ b/zokrates_book/src/sha256example.md
@@ -115,7 +115,7 @@ In the example we're considering, all inputs are private and there is a single r
 
 Peggy can then submit her proof by calling `verifyTx`.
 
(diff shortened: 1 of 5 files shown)


Schaeff:
Hmm actually it breaks `sha256example.md`

```
Victor monitors the verification smart contract for the `Verified` event, which is emitted upon successful verification of a transaction. As soon as he observes the event triggered by a transaction from Peggy's public address, he can be sure that Peggy has a valid pre-image for the hash he set in the smart contract.
```

PR #599, opened 2020-05-13
Add git pre-commit hook

dark64:
Added git pre-commit hook which checks rust formatting in modified files before committing.  Script can be found in `.githooks` directory and locally activated by running:

```sh
git config core.hooksPath .githooks
```

### DIFF
--- a/scripts/install_hooks.sh
+++ b/scripts/install_hooks.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+git config core.hooksPath .githooks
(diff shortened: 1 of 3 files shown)


Schaeff:
I had this in the past, I think it would be better for the hook to only check, not modify

scripts/install_hooks.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
Schaeff, line 1: Is this script absolutely necessary? If so, I think instructions should be added to the `Contributing` section of the readme

dark64: hmm we don't really need a new script for this, I've made some changes in the readme on how to enable the hooks

PR #601, opened 2020-05-14
Update book

dark64:
- [x] Add verify command to cli docs
- [x] Update docs to be inline with newest changes to the code base

### DIFF
--- a/zokrates_book/src/concepts/comments.md
+++ b/zokrates_book/src/concepts/comments.md
@@ -11,8 +11,5 @@ Inline (single-line) comments allow narrative on only one line at a time. Single
 Multi-line comments have one or more lines of narrative within a set of comment delimiters. The `/*` delimiter marks the beginning of the comment, and the `*/` marks the end. Any content between those delimiters is considered a comment.
 
 ```zokrates
-/*
-    This is a multi-line
-    comment.
-*/
+{{#include ../../../zokrates_cli/examples/book/multiline_comments.zok}}
 ```
--- a/zokrates_book/src/concepts/comments.md
+++ b/zokrates_book/src/concepts/comments.md
@@ -1,7 +1,18 @@
 ## Comments
 
-Comments can be added with double-slashes.
+## Inline comments
+Inline (single-line) comments allow narrative on only one line at a time. Single-line comments can begin in any column of a given line and end at a new line or carriage return. Inline comments can be added with double-slashes.
 
 ```zokrates
 {{#include ../../../zokrates_cli/examples/book/comments.zok}}
+```
+
+## Multi-line comments
+Multi-line comments have one or more lines of narrative within a set of comment delimiters. The `/*` delimiter marks the beginning of the comment, and the `*/` marks the end. Any content between those delimiters is considered a comment.
+
+```zokrates
+/*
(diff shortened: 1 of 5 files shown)


zokrates_book/src/concepts/comments.md
```diff
@@ -1,7 +1,18 @@
...
 
 ```zokrates
 {{#include ../../../zokrates_cli/examples/book/comments.zok}}
+```
+
+## Multi-line comments
+Multi-line comments have one or more lines of narrative within a set of comment delimiters. The `/*` delimiter marks the beginning of the comment, and the `*/` marks the end. Any content between those delimiters is considered a comment.
+
+```zokrates
+/*
```
Schaeff, line 14: This should be imported from the examples so that it gets tested in CI

zokrates_book/src/reference/cli.md
```diff
@@ -76,22 +76,45 @@ Using the proving key at `./proving.key`, generates a proof for a computation of
...
-Verifier.at(<verifier contract address>).verifyTx(A, A_p, B, B_p, C, C_p, H, K, [...publicInputs, ...outputs])
+
+## `verify`
+
+```sh
+zokrates verify
 ```
 
-Where `A, ..., K` are defined as above (adding brackets and quotes: `A = ["0x123", "0x345"]`), `publicInputs` are the public inputs supplied to witness generation and `outputs` are the results of the computation.
+Natively verifies a given proof `./proof.json` with the given verification key `./verification.key`.
```
Schaeff, line 119:
```suggestion
Natively verifies a given proof `./proof.json` with a given verification key `./verification.key`.
```

zokrates_book/src/reference/verification.md
```diff
@@ -1,10 +1,20 @@
 # Verification
 
-Passed to the verifier contract, this proof can be checked.
+Passed to the verifier contract, a given proof can be checked.
```
Schaeff, line 3:
```suggestion
Passed to the verifier contract, a proof can be checked.
```

zokrates_book/src/reference/verification.md
```diff
@@ -1,10 +1,20 @@
...
+const address = '0x456...'; // verifier contract address
 
-Where `A, ..., K` are defined as above (adding brackets and quotes: `A = ["0x123", "0x345"]`), `publicInputs` are the public inputs supplied to witness generation and `outputs` are the results of the computation.
\ No newline at end of file
+let verifier = new web3.eth.Contract(abi, address, {
+    from: accounts[0], // default from address
+    gasPrice: '20000000000000'; // default gas price in wei
+});
+
+verifier.methods.verifyTx(proof.proof, proof.inputs)
```
Schaeff, line 15: The `verifyTx` method will be made `view` with #597, I think in this case we can make a simple call rather than sending a transaction here? Doesn't have to be fixed here but we need to keep it in mind. Could add a comment on #597 to adjust this when we merge?

dark64: oh right, I fixed it here

dark64:
We should merge this after https://github.com/Zokrates/ZoKrates/pull/597 is merged

PR #602, opened 2020-05-14
Add struct names to typed_absy types

Schaeff:
In some cases, types with equal fields and different names are wrongly considered equal.

Add names to typed_absy types to prevent this.

Update the abi spec to output struct names

Todo

- [x] define and test behaviour for imported types

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,9 +42,9 @@ jobs:
       - restore_cache:
           keys:
             - v4-cargo-cache-{{ arch }}-{{ checksum "Cargo.lock" }}
-      - run:
-          name: Check format
-          command: rustup component add rustfmt; cargo fmt --all -- --check
+      # - run:
+      #     name: Check format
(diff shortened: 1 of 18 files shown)


.circleci/config.yml
```diff
@@ -42,9 +42,9 @@ jobs:
       - restore_cache:
           keys:
             - v4-cargo-cache-{{ arch }}-{{ checksum "Cargo.lock" }}
-      - run:
-          name: Check format
-          command: rustup component add rustfmt; cargo fmt --all -- --check
+      # - run:
```
JacobEberhardt, line 45: Why is the format check removed in this PR?

Schaeff: because https://rust-lang.github.io/rustup-components-history/ again 
Maybe it would make sense to ignore this check if the component isn't available, it's not blocking and would prevent these kind of changes..

Schaeff: Fixed

PR #603, opened 2020-05-15
Add operator docs to book

petscheit:
I was going through open my open PR's and found this. Updated the examples to the new ZoKrates format. Close if not needed.


Schaeff:
Thanks for this! Used in #647

PR #615, opened 2020-06-15
Fix broken links, correct declaration docs

### LINKED ISSUE #611: Function docs is out of date wrt inference

Schaeff:
closes #611

### DIFF
--- a/zokrates_cli/examples/book/declaration.zok
+++ b/zokrates_cli/examples/book/declaration.zok
@@ -1,0 +1,6 @@
+def main() -> ():
+	// declare and define a
+	field my_variable = 2
+	// redefine a
+	my_variable = 3
+	return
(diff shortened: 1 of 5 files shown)


zokrates_cli/examples/book/declaration.zok
```diff
@@ -0,0 +1,6 @@
+def main() -> ():
+	// declare and define a
```
dark64, line 2: should say `// declare and define my_variable`, or variable should be called `a`

zokrates_cli/examples/book/declaration.zok
```diff
@@ -0,0 +1,6 @@
...
+	field my_variable = 2
+	// redefine a
```
dark64, line 4: same here as above

zokrates_cli/examples/book/declaration.zok
Schaeff: good catch!

PR #623, opened 2020-06-25
Add stdlib path argument to cli

dark64:
- Added `--stdlib-path <PATH>` argument to cli
- Removed `ZOKRATES_HOME` environment variable

Todo:
- [x] Set stdlib-path argument as optional (should default to `$HOME/.zokrates/stdlib`)


zokrates_fs_resolver/src/lib.rs
```diff
@@ -83,14 +76,14 @@ mod tests {
 
     #[test]
     fn non_existing_file() {
-        let fs_resolver = FileSystemResolver::new();
+        let fs_resolver = FileSystemResolver::new("");
```
Schaeff, line 79: can we rename `new` to `with_stdlib_root` to make it clearer?

dark64: yeah good catch

zokrates_fs_resolver/src/lib.rs
```diff
@@ -31,20 +34,13 @@ impl Resolver<io::Error> for FileSystemResolver {
         }
 
         // paths starting with `./` or `../` are interpreted relative to the current file
-        // other paths `abc/def` are interpreted relative to $ZOKRATES_HOME
+        // other paths `abc/def` are interpreted relative to standard library root path
```
Schaeff, line 37:
```suggestion
        // other paths `abc/def` are interpreted relative to the standard library root path
```

PR #624, opened 2020-06-29
Put expensive optimisations behind `release` compile flag

Schaeff:
The redefinition optimiser accounts for most of the compilation time in many cases, due to the amount of copying required when substituting variables with linear combinations. Hence, introduce a `release` flag required to enable this optimisation. In the future, if other optimisations are deemed too expensive in development mode, the same could be applied to them.

As a side effect, introduce a `CompileConfig` struct to pass a configuration at compile time. This can be useful for example to enable sourcemaps @dark64


Schaeff:
CI failing because of an error installing wasm bindgen

Schaeff:
> CI failing because of an error installing wasm bindgen

fixed

PR #633, opened 2020-07-09
Add node support for zokrates.js

dark64:


zokrates_core/Cargo.toml
```diff
@@ -9,7 +9,7 @@ build = "build.rs"
 [features]
 default = []
 libsnark = ["cc", "cmake", "git2"]
-wasm = ["bellman_ce/wasm", "zokrates_embed/wasm"]
+wasm = ["bellman_ce/nolog", "bellman_ce/wasm", "zokrates_embed/wasm"]
```
Schaeff, line 12: Can we enable `nolog` in all cases rather than behind this feature?

zokrates_js/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "zokrates_js"
-version = "1.0.22"
+version = "1.0.23"
```
Schaeff, line 3: We usually bump version numbers when releasing, as if another PR changes zokrates_js we would end up bumping twice. Is there another reason to bump here?

zokrates_js/README.md
```diff
@@ -20,7 +20,7 @@ npm install zokrates-js
 ## Usage
 
 ```js
-import { initialize } from 'zokrates-js';
+import { initialize } from 'zokrates-js'; // for node: const { initialize } = require('zokrates-js/node');
```
Schaeff, line 23: Maybe it would be clearer to have two separate code blocks to describe how to import it in browser and nodejs, then this common code block to show usage?

zokrates_embed/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2018"
 
 [features]
 default = []
-wasm = ["bellman_ce/wasm", "sapling-crypto_ce/wasm"]
+wasm = ["bellman_ce/nolog", "bellman_ce/wasm", "sapling-crypto_ce/wasm"]
```
Schaeff, line 9: Same here

zokrates_js/tests/tests.js
```diff
@@ -0,0 +1,122 @@
+const assert = require('assert');
+const { initialize } = require('../node/index.js');
+
+describe('tests', function() {
+
+    before(function (done) {
```
Schaeff, line 6: Just to clarify: this initializes a single time before the suite of test, it's fine because `zokrates` is stateless, otherwise we would want `beforeEach`? If it's the case, would be nice to add a comment, or use `beforeEach`

zokrates_js/Cargo.toml
dark64: we pushed latest version manually, so 1.0.23 is the latest version on npm, we still need to bump the version for new release to 1.0.24

zokrates_js/tests/tests.js
dark64: `before` is appropriate for this case, as we only need to initialize it once before running the tests

zokrates_js/README.md
dark64: right, makes sense!

PR #640, opened 2020-07-16
Don't rely on raw encoding of Groth16 proof/verifying key for verification

ErikP0:
Before this PR, the `VerificationKey` and `ProofPoints` exposed the points (i.e. `a`,`b`,`gamma`, etc.) through serialization, but they were not used for verification. Instead, bellman's raw binary encoding was used to reconstruct bellman's verifying key and proof.

This makes it very hard to feed a proof or verifying key from an external source, e.g. json, that only contains the points. One would need to construct the binary encoding via bellman and then pass it together to Zokrates. This defies the purpose of using Zokrates in the first place.

This PR implements a conversion between Zokrates' G16 `VerificationKey` and bellman's `VerifyingKey` that doesn't rely on raw binaray encoding. And in turn enables the verification only to require the points of verifying key and proof.

Note that I needed to add `new_fq2` to the trait `Field` as I didn't find a generic way to construct the `Fqe` type. I didn't find an equivalent to `from_hex`.

### DIFF
--- a/zokrates_core/src/proof_system/bellman/groth16.rs
+++ b/zokrates_core/src/proof_system/bellman/groth16.rs
@@ -2,6 +2,7 @@ use bellman::groth16::{
     prepare_verifying_key, verify_proof, Parameters, PreparedVerifyingKey, Proof as BellmanProof,
     VerifyingKey,
 };
+use pairing::{CurveAffine, Engine};
 use regex::Regex;
 
 use zokrates_field::Field;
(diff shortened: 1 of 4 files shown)


zokrates_core/src/proof_system/bellman/groth16.rs
```diff
@@ -203,6 +221,23 @@ impl<T: Field> ProofSystem<T> for G16 {
     }
 }
 
+mod serialization {
+    use proof_system::{G1Affine, G2Affine};
+    use zokrates_field::Field;
+    use pairing::{CurveAffine, Engine, from_hex};
+
+    pub fn to_g1<E: Engine>(g1: G1Affine) -> E::G1Affine {
```
Schaeff, line 229: For consistency, how about having both `to_g1` and `to_g2` both be generic over `T: Field`? `<T::BellmanEngine as Engine>` can be used to get the associated bellman engine

PR #670, opened 2020-08-14
Add custom serializer for Type, set raw field as optional in proof

dark64:
- Added custom type serializer to fix weird serialization for uint type
- Set `raw` field as optional in `Proof`

### DIFF
--- a/zokrates_core/src/typed_absy/types.rs
+++ b/zokrates_core/src/typed_absy/types.rs
@@ -156,38 +156,83 @@ impl<'de> Deserialize<'de> for Type {
 
         let mapping = Mapping::deserialize(d)?;
         match mapping.ty.as_str() {
-            "field" => Ok(Type::FieldElement),
-            "bool" => Ok(Type::Boolean),
+            "field" => {
+                if mapping.components.is_some() {
+                    Err(D::Error::custom(format!(
+                        "unexpected `components` field for type `{}`",
+                        mapping.ty
+                    )))
+                } else {
+                    Ok(Type::FieldElement)
+                }
+            }
+            "bool" => {
+                if mapping.components.is_some() {
+                    Err(D::Error::custom(format!(
+                        "unexpected `components` field for type `{}`",
+                        mapping.ty
+                    )))
+                } else {
+                    Ok(Type::Boolean)
+                }
+            }
             "array" => {
-                let components = mapping.components.ok_or(D::Error::custom(format!(
-                    "Missing `components` field for type `{}`",
+                let components = mapping.components.ok_or(D::Error::custom(format_args!(
+                    "missing `components` field for type `{}'",
                     mapping.ty
                 )))?;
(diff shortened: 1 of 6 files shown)


zokrates_core/src/typed_absy/types.rs
```diff
@@ -96,21 +96,102 @@ impl fmt::Display for UBitwidth {
...
+        #[derive(Debug, Deserialize)]
+        struct Mapping {
+            #[serde(rename = "type")]
+            ty: String,
+            components: Option<Components>,
+        }
+
+        let mapping = Mapping::deserialize(d)?;
+        match mapping.ty.as_str() {
+            "field" => Ok(Type::FieldElement),
```
Schaeff, line 159: This should force `mapping.components` to be `None`

zokrates_core/src/typed_absy/types.rs
```diff
@@ -96,21 +96,102 @@ impl fmt::Display for UBitwidth {
...
+            "bool" => Ok(Type::Boolean),
```
Schaeff, line 160: same here

zokrates_core/src/typed_absy/types.rs
```diff
@@ -96,21 +96,102 @@ impl fmt::Display for UBitwidth {
...
+                    _ => Err(D::Error::custom(format!(
+                        "Invalid `components` field for type `{}`",
+                        mapping.ty
+                    ))),
+                }
+            }
+            "u8" => Ok(Type::Uint(UBitwidth::B8)),
+            "u16" => Ok(Type::Uint(UBitwidth::B16)),
+            "u32" => Ok(Type::Uint(UBitwidth::B32)),
+            _ => Err(D::Error::custom(format!("Unknown type `{}`", mapping.ty))),
```
Schaeff, line 190:
```suggestion
            _ => Err(D::Error::custom(format!("Invalid type `{}`", mapping.ty))),
```

zokrates_core/src/typed_absy/types.rs
```diff
@@ -156,38 +156,83 @@ impl<'de> Deserialize<'de> for Type {
 
         let mapping = Mapping::deserialize(d)?;
         match mapping.ty.as_str() {
-            "field" => Ok(Type::FieldElement),
-            "bool" => Ok(Type::Boolean),
+            "field" => {
+                if mapping.components.is_some() {
```
Schaeff, line 160: I feel this would be more idiomatic with a match or something like https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=86d9756a852c677c721add6e6eff77d1 but in this case it's readable

The match seems more readable here

dark64: replaced the `if` statement with a `match`. 

I've moved the check to a closure to reduce code duplication

Schaeff (approved):

Issue #673, opened 2020-09-10
Force field literals to be in range


Schaeff:
```
field a = 333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333
```

should not be allowed

Ideally we would catch this in semantic checks, as errors are more graceful there (and the max value depends on the curve)

PR #674, opened 2020-09-11
Field constant range check

dark64:
- `absy::Expression::FieldConstant` now takes `BigUint`
- `Field` is introduced in the `typed_absy` after range check is done

https://github.com/Zokrates/ZoKrates/issues/673


zokrates_core/src/semantics.rs
```diff
@@ -1411,7 +1411,20 @@ impl<'ast> Checker<'ast> {
                     }),
                 }
             }
-            Expression::FieldConstant(n) => Ok(FieldElementExpression::Number(n).into()),
+            Expression::FieldConstant(n) => {
+                let bits = n.bits();
+                let required_bits = T::get_required_bits();
+                match bits <= required_bits {
```
Schaeff, line 1417: I think it has to be stricter than this: `n <= T::max_value.to_biguint()`

Schaeff: To add to the unit tests with max and max + 1

zokrates_core/src/semantics.rs
```diff
@@ -1411,20 +1411,17 @@ impl<'ast> Checker<'ast> {
...
-                    false => Err(ErrorInner {
-                        pos: Some(pos),
-                        message: format!(
-                            "Field constant not in the representable range ({} bits)",
-                            required_bits
-                        ),
-                    }),
-                }
-            }
+            Expression::FieldConstant(n) => match n <= T::max_value().to_biguint() {
```
Schaeff, line 1414: Actually now that I read this again, wouldn't it make sense here to implement `impl<T: Field> TryFrom<BigUint> for T` and use this here? Only the error would need to be adjusted with `map_err`.

Right now in `zokrates_field` with have `From`, but it should be removed as it can fail, as this case shows

dark64: actually yes, makes perfect sense

zokrates_field/src/lib.rs
```diff
@@ -298,11 +299,18 @@ mod prime_field {
...
-                        value: &x - x.div_floor(&*P) * &*P,
+            impl TryFrom<BigUint> for FieldPrime {
+                type Error = ();
+
+                fn try_from(value: BigUint) -> Result<Self, ()> {
+                    match value <= Self::max_value().to_biguint() {
+                        true => {
+                            let x = ToBigInt::to_bigint(&value).unwrap();
+                            Ok(FieldPrime {
+                                value: &x - x.div_floor(&*P) * &*P,
```
Schaeff, line 310: The euclidean division can be removed

zokrates_js/tests/tests.js
```diff
@@ -26,7 +26,7 @@ describe('tests', function() {
         it('should resolve stdlib module', function() {
             const stdlib = require('../stdlib.json');
             assert.doesNotThrow(() => {
-                const code = `import "${Object.keys(stdlib)[0]}" as func\ndef main(): return`;
+                const code = `import "${Object.keys(stdlib)[0]}" as func\ndef main() -> (): return`;
```
Schaeff, line 29: can you explain this change? the old version should work

dark64: ahh this wasn't needed, i forgot to revert when testing

Schaeff (approved):
⚡

PR #681, opened 2020-10-06
Zexe backend integration

dark64:
Change log:
- Proof system refactor
- Added zexe as a backend option
- Introduced new curves `bls12_377`, `bw6_761`
- Added `verify` function to zokrates js with appropriate tests

**Note**: CI seems to fail because of the out of memory (OOM) issues


zokrates_core/src/proof_system/scheme/gm17.rs
```diff
@@ -0,0 +1,281 @@
...
+    type VerificationKey = VerificationKey<G1Affine, G2Affine>;
+    type ProofPoints = ProofPoints<G1Affine, G2Affine>;
+}
+
+impl Scheme<Bw6_761Field> for GM17 {
+    type VerificationKey = VerificationKey<G1Affine, G2AffineFq>;
+    type ProofPoints = ProofPoints<G1Affine, G2AffineFq>;
+}
+
+impl<T: Field + NotBw6_761Field> SolidityCompatibleScheme<T> for GM17 {
```
Schaeff, line 43: To work with solidity, we must be using Bn128Field. This allows more than that (for example BLS12-377)

zokrates_core/src/proof_system/scheme/mod.rs
```diff
@@ -0,0 +1,17 @@
...
+pub mod gm17;
+pub mod groth16;
+pub mod pghr13;
+
+pub trait Scheme<T: Field> {
+    type VerificationKey: Serialize + DeserializeOwned;
+    type ProofPoints: Serialize + DeserializeOwned;
+}
+
+pub trait SolidityCompatibleScheme<T: Field>: Scheme<T> {
```
Schaeff, line 15: Right now it's only Bn128, maybe we need a `SolidityCompatibleField´ trait which only `Bn128Field` implements and then have it as a bound here?

zokrates_core/src/proof_system/scheme/pghr13.rs
```diff
@@ -0,0 +1,315 @@
...
+    pub z: G2,
+    pub ic: Vec<G1>,
+}
+
+impl<T: Field> Scheme<T> for PGHR13 {
+    type VerificationKey = VerificationKey<G1Affine, G2Affine>;
+    type ProofPoints = ProofPoints<G1Affine, G2Affine>;
+}
+
+impl<T: Field> SolidityCompatibleScheme<T> for PGHR13 {
```
Schaeff, line 40: Same here

zokrates_core/src/proof_system/zexe/mod.rs
```diff
@@ -0,0 +1,306 @@
...
+use r1cs_core::{
+    ConstraintSynthesizer, ConstraintSystem, ConstraintSystemRef, LinearCombination,
+    SynthesisError, Variable,
+};
+use std::collections::BTreeMap;
+use zokrates_field::{Field, ZexeFieldExtensions};
+
+pub use self::parse::*;
+
+use rand_0_7::SeedableRng;
```
Schaeff, line 21: cf feature guard on rand_0_7

zokrates_core/src/proof_system/zexe/mod.rs
```diff
@@ -0,0 +1,306 @@
...
+
+        // extract public inputs
+        let public_inputs = self.public_inputs_values();
+
+        assert!(verify_proof(&pvk, &proof, &public_inputs).unwrap());
+
+        proof
+    }
+
+    pub fn public_inputs_values(&self) -> Vec<<T::ZexeEngine as PairingEngine>::Fr> {
```
Schaeff, line 187: This function comes up a lot and clones the witness. Would make sense to extract it to `ir` or `witness` and avoid clones.

zokrates_field/Cargo.toml
```diff
@@ -10,10 +10,15 @@ serde_derive = "1.0"
...
 serde_json = "1.0"
+bellman_ce = { version = "^0.3", default-features = false }
+algebra-core = { git = "https://github.com/scipr-lab/zexe.git", default-features = false }
 sha2 = "0.8.0"
 num-traits = { version = "0.2", default-features = false }
 num-integer = { version = "0.1", default-features = false }
-bellman_ce = { version = "^0.3", default-features = false }
+
+[dependencies.algebra]
+git = "https://github.com/scipr-lab/zexe.git"
```
Schaeff, line 20: Do we have to rely on github?

zokrates_js/src/lib.rs
```diff
@@ -32,12 +35,6 @@ pub struct ComputationResult {
     output: String,
 }
 
-impl ResolverResult {
```
Schaeff, line 35: Was this not being used?

Schaeff (changes requested):
Very nice, made a few suggestions!

zokrates_core/src/proof_system/libsnark/ffi.rs
```diff
@@ -20,7 +20,7 @@ extern "C" {
 }
 
 impl Buffer {
-    pub fn from_vec(v: &Vec<u8>) -> Buffer {
+    pub unsafe fn from_vec(v: &Vec<u8>) -> Buffer {
```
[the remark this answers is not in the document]
dark64: This function is unsafe as it itentionally leaves a memory leak which must be handled by the caller. I marked it unsafe to prevent any future usage outside of the unsafe block.

zokrates_cli/src/helpers.rs
```diff
@@ -2,91 +2,108 @@ use core::convert::TryFrom;
 
 use crate::constants::*;
 
-pub enum Curve {
+#[derive(Debug)]
+pub enum CurveDimension {
```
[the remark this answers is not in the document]
dark64: right, I renamed it as there is a trait called `Backend` and it caused a name collision, I don't really like to alias stuff in rust, but yes I agree this should be renamed to something else

zokrates_cli/tests/integration.rs
```diff
@@ -240,6 +243,24 @@ mod integration {
                 .succeeds()
                 .unwrap();
 
+                // EXPORT-VERIFIER
```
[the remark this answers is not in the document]
dark64: Good catch!

zokrates_core/src/lib.rs
```diff
@@ -8,13 +8,24 @@ extern crate serde_json;
...
 extern crate csv;
 extern crate ff_ce as ff;
+#[cfg(feature = "zexe")]
+extern crate gm17 as zexe_gm17;
 extern crate hex;
 extern crate lazy_static;
 extern crate pairing_ce as pairing;
+#[cfg(feature = "zexe")]
+extern crate r1cs_core;
+extern crate rand_0_4;
```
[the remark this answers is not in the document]
dark64: yeah good catch, it should only be included for the zexe feature

zokrates_field/Cargo.toml
dark64: Yeah it seems to not be published as a crate

zokrates_cli/src/helpers.rs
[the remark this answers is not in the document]
Schaeff: That was my assumption, and I agree with avoiding aliasing! I think `*Parameter` would be fine everywhere we use `*Dimension`

zokrates_core/src/proof_system/libsnark/ffi.rs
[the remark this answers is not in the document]
Schaeff: Interesting. I never really thought about this too much, but my assumption is that `unsafe` was exclusively to turn off some compiler safety checks, so as a result, anything that would compile without `unsafe` shouldn't have the flag. Does this function compile without `unsafe`?

zokrates_field/Cargo.toml
Schaeff: Could we target a specific commit?  `38b6f6f` was used to release some sub-crates https://github.com/scipr-lab/zexe/releases

zokrates_core/src/proof_system/libsnark/ffi.rs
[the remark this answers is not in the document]
dark64: Yes you can compile this without the `unsafe` keyword, though after further digging through the rust documentation, it looks like Rust considers memory leaks to be safe. I'll remove this for now.

zokrates_js/src/lib.rs
dark64: Yeah but I feel like this impl is unnecessary

zokrates_core/lib/util.tcc
```diff
@@ -8,67 +8,104 @@
...
+    const size_t ELEMENT_SIZE = Q * sizeof(mp_limb_t);
+
+    G1 aff = point;
     aff.to_affine_coordinates();
-    return "[\"0x" + hexStringFromLibsnarkBigint<Q>(aff.X.as_bigint()) + "\",\"0x" + hexStringFromLibsnarkBigint<Q>(aff.Y.as_bigint()) + "\"]";
+
+    auto x = aff.X.as_bigint();
+    auto y = aff.Y.as_bigint();
+
+    from_libff_bigint<Q>(x, buffer);
```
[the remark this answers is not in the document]
dark64: I'm not sure if it's good idea to have this helper functions increment the buffer as it's used elsewhere where this behavior is not wanted

zokrates_core/src/proof_system/zexe/mod.rs
dark64: I don't this is necessary here as the whole zexe module is under a feature guard? I might be wrong here

zokrates_core/src/proof_system/scheme/mod.rs
dark64: Right, I'll do some refactoring on this part as well

zokrates_core/src/proof_system/zexe/mod.rs
Schaeff: Correct

zokrates_field/Cargo.toml
dark64: Will open issue for this

PR #682, opened 2020-10-08
Test for compile errors, fix breaking tests

Schaeff:
Some tests related to compile-time errors are not being run.

This revealed and issue with the handling of struct equivalence (two structs are equal if the import chain leads to the same struct declaration) which is fixed in this PR.

The abi was changed to carry the canonical struct name rather than the renamed one.


JacobEberhardt (approved):
Struct equivalence considers import location. Identical struct definitions in different source locations are not considered equal

PR #683, opened 2020-10-20
Edits for Getting Started

qbzzt:
Changed to use the docker version. It makes the command line more cumbersome, but it removes the malleability warning, which I think is confusing.

### DIFF
--- a/zokrates_book/src/gettingstarted.md
+++ b/zokrates_book/src/gettingstarted.md
@@ -11,7 +11,7 @@ To write your first SNARK program, check out the ZoKrates plugin in the [Remix o
 ZoKrates is available on Dockerhub.
 
 ```bash
-docker run -ti zokrates/zokrates zokrates
+docker run zokrates/zokrates zokrates
 ```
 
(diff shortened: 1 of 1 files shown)


Schaeff:
Hey @qbzzt thanks for this suggestion. I am not convinced of the net gain of using docker everywhere, due to the fact that it adds a dependency, and some parts of the workflow are computationally intensive, so trying to be native when possible is recommended.

I see you're unhappy with the groth16 warning, here are some suggestions:
- we could have this warning in the documentation only https://zokrates.github.io/toolbox/proving_schemes.html#g16-malleability
- as you found out, we can use another scheme like gm17. Turns out #681 will enable gm17 without libsnark, so docker would no be required. This should be merged very soon so we could rely on it.

qbzzt:
> * as you found out, we can use another scheme like gm17. Turns out #681 will enable gm17 without libsnark, so docker would no be required. This should be merged very soon so we could rely on it.

Great. Tell me when #681 is added please?

PR #688, opened 2020-10-31
Members of groth16 Proof made public.

sladecek:
I am using ZoKrates as a Rust library and need to have direct access to members of the groth16 Proof structure.


Schaeff:
Thanks for this @sladecek, could you target the `develop` branch?

PR #695, opened 2020-11-09
Constant generics

Schaeff:


zokrates_core/src/flatten/mod.rs
```diff
@@ -853,6 +948,7 @@ impl<'ast, T: Field> Flattener<'ast, T> {
                     .map(|x| x.get_field_unchecked())
                     .collect::<Vec<_>>();
 
+                #[allow(clippy::needless_collect)]
```
JacobEberhardt: Shouldn't we remove the clippy exception here, as it's global for the file already?

zokrates_core/src/ir/expression.rs
```diff
@@ -263,6 +260,7 @@ impl<T: Field> Mul<&T> for LinComb<T> {
 impl<T: Field> Div<&T> for LinComb<T> {
     type Output = LinComb<T>;
```
JacobEberhardt:
```suggestion
// Clippy warns about multiplication in a method named div. It's okay, here, since we multiply with the inverse.
```

zokrates_core/src/static_analysis/trimmer.rs
```diff
@@ -0,0 +1,88 @@
+use crate::typed_absy::TypedModule;
```
JacobEberhardt, line 1: I found the name of this module surprising as it basically brings embeds into scope.

JacobEberhardt (dismissed):
Big one (literally^^). 
Only minor comments that can be addressed asynchronously

zokrates_core/src/static_analysis/trimmer.rs
Schaeff: I don't think it does, it's used to construct the module at line 54. Embeds are accessed with the full path `crate::embed::FlatEmbed` which I guess could be turned into an import up here.

zokrates_core/src/flatten/mod.rs
Schaeff: Done

PR #705, opened 2020-11-18
Remove unused deps, split zokrates_test

Schaeff:
In an attempt to reduce compilation times, remove unused dependencies flagged by `cargo-udeps`. Also, extract build-time functionality from the `zokrates_test` crate into `zokrates_test_derive`


dark64 (approved):
lgtm

PR #706, opened 2020-11-18
uint division

Schaeff:


JacobEberhardt (dismissed):
Clean. I'll add a short note in type doc before merge.

Issue #686, opened 2020-10-26
Option to turn off unconstrained variables check during compute witness


Westlad:
When debugging a complex proof, it's very useful to be able to comment out an assert line to check if it's the source of an 'expected 0 to equal 1' problem.  The unconstrained variables check prevents one being able to do this anymore.  Would it be possible to turn it off for debugging, maybe with a '--no-unconstrained-check' option on compute-witness?

Schaeff:
Thanks for this @Westlad !

Is this a correct description of your problem:

```
def main(field a):
   assert(a == 42) // <- commenting out this line isn't allowed
   return
```

You're mentioning witness computation, but the unused variable check happens at compilation. Is this what you meant?

If so, what we could do is make unconstrained variables a warning which can be ignored during compilation. Under the hood, that would insert a dummy constraint `a == 1 * a` per unused argument to bypass the bellman unused input check.

Westlad:
Thanks @Schaeff, that's exactly what I was thinking of.

PR #709, opened 2020-11-25
Add dummy condition statement for private field parameters

dark64:
To avoid unconstrained variables in cases like the following

```
def main(private field a) -> (field):
    return 42
```

we can introduce a dummy constraint in the format of `y == x * x`. This will unfortunately introduce an intermediate variable for each private field, which may marginally increase the size of the program and the witness. Ideally, it would be best to add a basic constraint ` x == 1 * x` but tautology step in the optimizer would remove this.

This solves https://github.com/Zokrates/ZoKrates/issues/686

Todo:
- [x] Put this under a compilation flag


Schaeff:
Nice! It would be nice to put this behind a compile time flag, in the spirit of the compilation config we removed in #668, so that this is disabled by default. What do you think?

dark64:
> Nice! It would be nice to put this behind a compile time flag, in the spirit of the compilation config we removed in #668, so that this is disabled by default. What do you think?

I agree, make sense to put this under a flag!

Schaeff (approved):

Westlad:
Nice!  Thanks @dark64 @Schaeff.

PR #720, opened 2020-12-17
Move cli commands in separate files

dark64:
For the sake of easier maintainability, it makes sense now to move commands in separate files. 
This cleans up big and messy `bin.rs` in `zokrates_cli`


zokrates_cli/src/helpers.rs
```diff
@@ -1,6 +1,5 @@
-use core::convert::TryFrom;
-
 use crate::constants::*;
+use core::convert::TryFrom;
```
Schaeff, line 2: I don't think we use the core crate anywhere else, would `std::convert::TryFrom` work here?

Schaeff (commented):
Nice one!

zokrates_cli/src/helpers.rs
dark64: Good catch, I guess this was an auto-import by the IDE, will change to `std::convert::TryFrom`

Schaeff (approved):

PR #728, opened 2021-01-05
Return of the embeds

dark64:
Brings back embed sha256 compression function from bellman_ce

- [x] Bellman embeds (sha256)
- [x] Update stdlib (should embed sha256 have the same signature as the native one?)


zokrates_core/src/embed.rs
```diff
@@ -66,12 +86,135 @@ impl FlatEmbed {
...
+///
+/// # Remarks
+///
+/// The variables inside the function are set in this order:
+/// - constraint system variables
+/// - arguments
+#[cfg(feature = "bellman")]
+pub fn sha256_round<T: Field>() -> FlatFunction<T> {
+    // Define iterators for all indices at hand
+    let (r1cs, input_indices, current_hash_indices, output_indices) =
```
Schaeff, line 143: This function is generic over T but calls `generate_sha256_round_constraints` with Bn256. Therefore I think we should `assert_eq!(T::id(), Bn128Field::id())` here

zokrates_core/src/imports.rs
```diff
@@ -149,6 +149,18 @@ impl Importer {
             // handle the case of special bellman and packing imports
             if import.source.starts_with("EMBED") {
                 match import.source.to_str().unwrap() {
+                    #[cfg(feature = "bellman")]
```
Schaeff, line 152: Am I following correctly that we can't refuse this embed here as we did not introduce T yet?

zokrates_core/src/solvers/mod.rs
```diff
@@ -114,7 +120,28 @@ impl<T: Field> Executable<T> for Solver {
                 let r = n - d * &q;
                 vec![T::try_from(q).unwrap(), T::try_from(r).unwrap()]
             }
-            _ => unimplemented!(),
+            #[cfg(feature = "bellman")]
+            Solver::Sha256Round => {
```
Schaeff, line 124: same here, it's generic over T but calls Bn256, so we should assert ids match

Schaeff (changes requested):
Nice, would be good to add this sha to stdlib under a different name, and add a test the same way we have one for the native one

zokrates_core/src/embed.rs
dark64: nice catch

dark64:
> Nice, would be good to add this sha to stdlib under a different name, and add a test the same way we have one for the native one

Added tests, I've placed this sha inside an `embed` folder eg. `hashes/sha256/embed/512bit.zok`

Schaeff:
What's the status or this PR?

Schaeff (approved):

PR #732, opened 2021-01-26
run tests in thread with bigger stack

Schaeff:
`cargo test` without `--release` currently leads to stack overflows. It appears that this is caused by tests running in spawn threads with a default stack size much smaller than that of the main thread.

This PR spawns threads where applicable, providing a stack size of 8M.

With this change, `cargo test` runs without overflowing on my machine.


dark64 (approved):
LGTM!

PR #736, opened 2021-02-01
Adds pedersen bool library

petscheit:
Simple change for having a pedersen hashes for booleans. Need this currently and it doesn't make sense to wrap and unwrap so much. The naming was chosen to ensure backward compatibility for the u32 version.  Let me know if this is needed, thinks it makes sense, also since the embedded sha256 libraries use booleans.


thecodingshrimp:
[zkRelay](https://github.com/informartin/zkRelay) uses Zokrates and would benefit from merging this PR. 
[informartin/zkRelay#13](https://github.com/informartin/zkRelay/pull/13) depends on this PR.
Would love to see your review @dark64  @Schaeff

Schaeff (approved):
Thanks for nudging me on this, and sorry for the delay!

Issue #632, opened 2020-07-09
Groth16 keys generated for the same circuit using setup are the same each time it is run


ChaitanyaKonda:
To call `create_random_proof` and `generate_random_parameters` for groth16 from bellman, zokrates uses a random number generator rng `let rng = &mut ChaChaRng::new_unseeded();`. This function `new_unseeded()` within rand crate seems to use ChaCha random number generator using the default fixed key of 8 zero words. As a result anytime setup is run on the same circuit, it produces the same set of proving and verification keys. Is there a reason for this? It might be useful to have an rng to accept entropy instead of a fixed key such that the keys generated are different each time for the same circuit. For non deterministic rng, Rand 0.7 crate provides rng created with entropy based on the target OS

Schaeff:
Good call. We changed to ChaCha for wasm compatibility, happy to change to something else as long as that's covered too.

If there's no portable rng, we could have backend functions take an rng as parameter and have the caller instantiate it beforehand.

ChaitanyaKonda:
Rand 0.7.3's `StdRng`( a CPRNG) also uses ChaCha block cipher with 20 rounds. The crate suggests that they may change this as new evidence of cipher security and performance becomes available. `StdRng` implements `SeedableRng` which provides `from_entropy` function which in itself depends on `getrandom` to provide entropy based on target OS. Taking a quick look at the `getrandom` crate, suggests that 

> For the wasm32-unknown-unknown target, one of the following features should be enabled:
> 
> wasm-bindgen
> stdweb

dark64:
This is a tricky one. We can use `rand 0.7` specifically `OsRng` with `wasm-bindgen` feature enabled to preserve wasm compatibility and this would work for zexe. Unfortunately, bellman uses and expects `Rng` from `rand 0.4` which is not compatible with `rand 0.7`. We could use `OsRng` to generate the seed for `ChaChaRng` from `rand 0.4` and use that with bellman. Seems hacky though.

PR #738, opened 2021-02-03
Use rng with random seed

dark64:
Fix for issue https://github.com/Zokrates/ZoKrates/issues/632

We can use `getrandom` crate with `js` feature enabled to preserve [wasm compatibility](https://docs.rs/getrandom/0.2.2/getrandom/#webassembly-support) and generate a random seed which can be used with `ChaChaRng::from_seed(...)`

This is done because `bellman_ce` uses `rand 0.4` which seems to be not compatible with `rand 0.7` traits.

### DIFF
--- a/zokrates_core/src/proof_system/bellman/mod.rs
+++ b/zokrates_core/src/proof_system/bellman/mod.rs
@@ -162,8 +162,19 @@ impl<T: BellmanFieldExtensions + Field> Prog<T> {
 }
 
 impl<T: BellmanFieldExtensions + Field> Computation<T> {
+    fn get_random_seed(&self) -> Result<[u32; 8], getrandom::Error> {
+        let mut seed = [0u8; 32];
+        getrandom::getrandom(&mut seed)?;
+
+        use std::mem::transmute;
+        let seed: [u32; 8] = unsafe { transmute(seed) };
+        Ok(seed)
+    }
+
     pub fn prove(self, params: &Parameters<T::BellmanEngine>) -> Proof<T::BellmanEngine> {
-        let rng = &mut ChaChaRng::new_unseeded();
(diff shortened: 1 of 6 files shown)


zokrates_core/src/proof_system/bellman/mod.rs
```diff
@@ -162,8 +162,19 @@ impl<T: BellmanFieldExtensions + Field> Prog<T> {
 }
 
 impl<T: BellmanFieldExtensions + Field> Computation<T> {
+    fn get_random_seed(&self) -> Result<[u32; 8], getrandom::Error> {
+        let mut seed = [0u8; 32];
+        getrandom::getrandom(&mut seed)?;
+
+        use std::mem::transmute;
+        let seed: [u32; 8] = unsafe { transmute(seed) };
```
Schaeff, line 170: Is there a way to achieve this without an unsafe block? If not, a comment explaining why this is safe would be good here.

Schaeff (changes requested):

zokrates_core/src/proof_system/bellman/mod.rs
dark64: I could not find any other neat native solution for this but I believe this code is safe in this case imo, and it's the fastest. It just reinterprets a slice of u8's to a slice of 8 u32 unsigned integers. Byte order or anything does not matter here, as it's used as a seed for the rng. I feel like this code is quite clear for what it does, but I can leave a comment in the code if you feel like this is needed.

Schaeff: Yes if you can put what you just explained here in there that would be nice! "This is safe because ..."

Schaeff (approved):
Nice one!

PR #744, opened 2021-02-05
Add native blake2s to stdlib

dark64:
blake2s implementation based on https://tools.ietf.org/html/rfc7693


Schaeff (approved):
LGTM, great to have this!

Issue #746, opened 2021-02-11
Improve flattening of assertions in special cases


Schaeff:
`assert(boolean_value)` is flattened to `flat(boolean_value) * 1 == 1`, which is suboptimal in cases like `assert(field_value_0 == field_value_1)` where we could directly use `flat(field_value_0) * 1 == flat(field_value_1)`. 

List these cases (cf `zir::BooleanExpression`, see which variants already looks like R1CS), implement better flattening for them where applicable.

https://github.com/Zokrates/ZoKrates/blob/a8f09103e23801967ba425bb16a295d8a4da4f30/zokrates_core/src/flatten/mod.rs#L2011

PR #747, opened 2021-02-11
Improve flattening

dark64:
https://github.com/Zokrates/ZoKrates/issues/746

### DIFF
--- a/zokrates_core/src/flatten/mod.rs
+++ b/zokrates_core/src/flatten/mod.rs
@@ -2359,7 +2359,29 @@ mod tests {
         let mut flattener = Flattener::new(&config);
 
         let flat = flattener.flatten_function(&HashMap::new(), function);
-        println!("{}", flat);
+
+        let expected = FlatFunction {
+            arguments: vec![],
+            statements: vec![
+                FlatStatement::Definition(
+                    FlatVariable::new(0),
+                    FlatExpression::Number(Bn128Field::from(1)),
+                ),
+                FlatStatement::Definition(
+                    FlatVariable::new(1),
+                    FlatExpression::Number(Bn128Field::from(1)),
+                ),
+                FlatStatement::Condition(
+                    FlatExpression::Mult(
+                        box FlatExpression::Identifier(FlatVariable::new(0)),
+                        box FlatExpression::Number(Bn128Field::from(1)),
+                    ),
+                    FlatExpression::Identifier(FlatVariable::new(1)),
+                ),
+            ],
+        };
+
+        assert_eq!(flat, expected);
     }
 
     #[test]
--- a/zokrates_core/src/flatten/mod.rs
+++ b/zokrates_core/src/flatten/mod.rs
@@ -2359,7 +2359,6 @@ mod tests {
         let mut flattener = Flattener::new(&config);
 
         let flat = flattener.flatten_function(&HashMap::new(), function);
-
         let expected = FlatFunction {
             arguments: vec![],
             statements: vec![
@@ -2415,7 +2414,6 @@ mod tests {
         let mut flattener = Flattener::new(&config);
 
         let flat = flattener.flatten_function(&HashMap::new(), function);
-
         let expected = FlatFunction {
             arguments: vec![],
             statements: vec![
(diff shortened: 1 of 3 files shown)


zokrates_core/src/flatten/mod.rs
```diff
@@ -2083,6 +2120,24 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+    ///
+    /// # Arguments
+    /// * `e` - `BooleanExpression` to be flattened
+    fn unwrap_and_chain(
+        &mut self,
+        e: BooleanExpression<'ast, T>,
+    ) -> Vec<BooleanExpression<'ast, T>> {
+        match e {
+            BooleanExpression::And(box lhs, box rhs) => {
+                let mut inner = self.unwrap_and_chain(lhs);
```
Schaeff, line 2133: Why isn't `rhs` also visited here? Because we construct trees where right children are always leaves?

zokrates_core/src/flatten/mod.rs
```diff
@@ -2147,6 +2202,36 @@ impl<'ast, T: Field> Flattener<'ast, T> {
         }
     }
 
+    fn flatten_assertion(
```
Schaeff, line 2205: Would this be clearer is named `flatten_equality` ?

zokrates_core/src/flatten/mod.rs
```diff
@@ -2246,6 +2331,174 @@ mod tests {
...
+        };
+
+        assert_eq!(flat, expected);
+    }
+
+    #[test]
+    fn assertion_uint_eq() {
+        let metadata = UMetadata {
+            max: 0xffffffff_u32.into(),
+            should_reduce: ShouldReduce::False,
```
Schaeff, line 2452: I think this would be `ShouldReduce::True` in practice

zokrates_core/src/flatten/mod.rs
```diff
@@ -2016,21 +2016,58 @@ impl<'ast, T: Field> Flattener<'ast, T> {
                 }
             }
             ZirStatement::Assertion(e) => {
-                // naive approach: flatten the boolean to a single field element and constrain it to 1
+                match e {
+                    BooleanExpression::And(..) => {
+                        let mut iter = self.unwrap_and_chain(e).into_iter();
```
Schaeff, line 2021: for boolean in self.unwrap_and_chain(e) {
    self.flatten_statement(
          symbols,
          statements_flattened,
          ZirStatement::Assertion(boolean),
      )
}

zokrates_core/src/flatten/mod.rs
```diff
@@ -2083,6 +2120,24 @@ impl<'ast, T: Field> Flattener<'ast, T> {
         }
     }
 
+    /// Unwraps a boolean & (and) expression chain to a flat vector of boolean expressions
+    ///
+    /// # Arguments
+    /// * `e` - `BooleanExpression` to be flattened
+    fn unwrap_and_chain(
```
Schaeff, line 2127: I think this should be in the `impl` of `BooleanExpression` as it only depends on the expression.

Also this looks a lot like an iterator and it's used as such, so maybe it's better to skip allocating into this vector and iterate straight out of the expression.

See an idea of what an iterator could look like [here](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018). The advantage is that its internal vector only grows when we see an `And`, as opposed to this one which allocates `vec![e]` for all elements.

zokrates_core/src/flatten/mod.rs
dark64: Yeah, according to this https://github.com/Zokrates/ZoKrates/blob/develop/zokrates_core/src/static_analysis/flatten_complex_types.rs#L602

zokrates_core/src/flatten/mod.rs
dark64: yes this is more clear!

zokrates_core/src/flatten/mod.rs
```diff
@@ -2147,6 +2202,36 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+            FlatExpression::Number(_) | FlatExpression::Identifier(_) => rhs,
+            _ => {
+                let sym = self.use_sym();
+                statements_flattened.push(FlatStatement::Definition(sym, rhs));
+                FlatExpression::Identifier(sym)
+            }
+        };
+
+        statements_flattened.push(FlatStatement::Condition(
+            FlatExpression::Mult(box lhs_flattened, box FlatExpression::Number(T::from(1))),
```
Schaeff, line 2230: I suspect we can be a bit more efficient here:

```
flatten_assertion(x * y, z)

currently:
t = x * z
t * 1 == z

optimal:
x * y = z
```

Some with the mirrored `flatten_assertion(z, x * y)` -> `x * y == z`

Something like

```
let (lhs, rhs) = match (lhs, rhs) {
   (FlatExpression::Mult(x, y), z) if x.is_linear() && y.is_linear() && z.is_linear() => (FlatExpression::Mult(x, y), z),
   (z, FlatExpression::Mult(x, y)) if x.is_linear() && y.is_linear() && z.is_linear() => (FlatExpression::Mult(x, y), z),
   (x, z) if x.is_linear() && z.is_linear() => (FlatExpression::Mult(x, 1), z)
   (x, z) if x.is_linear() => {introduce variable for z, (FlatExpression::Mult(x, 1), t)},
   (x, z) if z.is_linear() => {introduce variable for x, (FlatExpression::Mult(z, 1), t)},
   (x, z) => {introduce variables for both, FlatExpression::Mult(t, 1), u)}
}

```

zokrates_core/src/flatten/mod.rs
Schaeff: Okay, I think flattening shouldn't make this assumption though.

zokrates_core/src/zir/mod.rs
```diff
@@ -422,6 +422,39 @@ pub enum BooleanExpression<'ast, T> {
...
+                    self.current.push(right);
+                    self.next()
+                }
+                n => Some(n),
+            })
+            .flatten()
+    }
+}
+
+impl<'ast, T> IntoIterator for BooleanExpression<'ast, T> {
```
Schaeff, line 447: I suspect it would be better to have this as part of the BooleanExpression rather than into_iter, it's not immediately clear what `into_iter` means for a tree and we're doing something quite specific here. What about `into_conjunction_iterator`? The only nice thing you get with this std trait is `for in` syntax which isn't a big deal here.

zokrates_core/src/flatten/mod.rs
```diff
@@ -2183,34 +2183,57 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
     }
 
+    /// Flattens an equality expression
+    ///
+    /// # Arguments
+    ///
+    /// * `statements_flattened` - `FlatStatements<T>` Vector where new flattened statements can be added.
+    /// * `lhs` - `FlatExpression<T>` Left-hand side of the equality expression.
+    /// * `rhs` - `FlatExpression<T>` Right-hand side of the equality expression.
     fn flatten_equality(
```
Schaeff, line 2193: Would be nice to have some tests for these in this file

x == y -> x * 1 = y
x * y == z -> same
z == x * y -> mirror
x * y == z * t -> introduce var

There are similar tests already, this can be written with `flatten_function`

zokrates_core/src/zir/mod.rs
dark64: yeah good point

zokrates_core/src/flatten/mod.rs
dark64: Agreed, I wrote some tests to cover all these cases

Schaeff (approved):
LGTM!

PR #750, opened 2021-02-24
Upgrade nodejs dependencies

Schaeff:
Trying to make the vulnerability warnings go away


Schaeff:
Dismissed the vulnerability on gh, as it is low severity and we only use it in development.

Issue #748, opened 2021-02-18
ztf file is not generated when light flag is used


dark64:
I think `*.ztf` files should be always generated, but the `light` flag should be a way to disable logging bigger programs to the console (so the output is less verbose).

Current implementation skips the generation of the `*.ztf` file when `light` flag is used. Other solution would be to introduce a new flag `verbose` for the logging but keep the `light` flag as is.

Schaeff:
@JacobEberhardt any thoughts on this?

JacobEberhardt:
I agree with Darko's analysis. `Light` was intended to disable logging to the console.

However, with the growing complexity of the compiler, the `*.ztf`file becomes less and less valuable to devs.
With the planned introduction of automatic packing (to optimize on-chain gas usage), the `*.ztf`will be even more decoupled.
What about having neither console nor `'*.ztf` logging by default and enabling those through flags only?

Schaeff:
I would support that. The only downside to me is that beginners can be curious what the output looks like, but a flag could be enough there.

Shall we remove console logging entirely and only rely on the ztf file when the flag is on?

Schaeff:
Also: printing the program seems to be taking a significant amount of time in compute-witness, I guess this should go as well.

JacobEberhardt:
Yes, I support that. Printing is basically a historical artifact at this point and should not be a default.
Let's put it behind a `verbose` flag and show how that flag is used in our introductory documentation.

Schaeff:
This is done? @dark64

dark64:
> This is done? @dark64

Yes, this has been merged.

PR #751, opened 2021-03-01
Add verbose flag

dark64:
https://github.com/Zokrates/ZoKrates/issues/748

Added a global `--verbose` flag which can be used to print compiler outputs to the console (ztf, witness etc). 
I've also renamed `--light` flag to `--ztf` as `light` seems a bit ambiguous atm. Some smaller improvements to the logs and missing prints are added to improve overall consistency.

To preserve compatibility and not break things, `--light` flag is still an available option but is considered deprecated (a warning will be displayed to the user). The flag currently overrides both `--ztf` and `--verbose` flag (basically doing what it did before) but should be removed entirely after the following release.


zokrates_cli/src/ops/compile.rs
```diff
@@ -74,16 +80,23 @@ pub fn exec(sub_matches: &ArgMatches) -> Result<(), String> {
 }
 
 fn cli_compile<T: Field>(sub_matches: &ArgMatches) -> Result<(), String> {
-    println!("Compiling {}\n", sub_matches.value_of("input").unwrap());
+    // TODO: remove the warning once light flag is removed entirely
+    if sub_matches.is_present("light") {
```
Schaeff, line 84: I suspect this is always false, as it get overridden

Schaeff (changes requested):

zokrates_cli/src/ops/compile.rs
dark64: It's actually reversed, `--light` overrides `--ztf` and `--verbose`

Schaeff: I'm struggling to understand the mechanism here, maybe I need another coffee can you explain what happens in different situations? Also `overrides_with_all` that ordering matter, is that taken into account here?

dark64: When you pass a `--light` flag it overrides both `--ztf` and `--verbose` (order does not matter here), essentially as if they were never used.

so a scenario like the following
```
zokrates compile -i main.zok --ztf --verbose --light
```
will result in
```
zokrates compile -i main.zok --light
```

This is the case where a warning will be displayed. The `--light` flag doesn't do anything here because its behaviour is now by default, but if other conflicting flags are passed (`ztf`, `verbose`) it will override them to make sure the logic stays the same as it was before changes introduced by this PR. The flag is also set as hidden (but its still usable and will display a warning of deprecation) to "force" people to use newer flags.

Another scenario is when you pass eg. `--ztf` after the `--light` flag for whatever reason, in this case `--light` would not override `--ztf` because whichever argument was specified at runtime last "wins".

Some other examples can be found in the `clap` docs https://ruststudy.github.io/tao_of_rust_docs/structopt/clap/struct.Arg.html#method.overrides_with_all

Schaeff: Okay that makes sense, so in the backward compatible case ordering doesn't matter because only `light` is there? Thanks!

dark64: Exactly!

Schaeff (approved):
🆗 🆒 ❗

PR #754, opened 2021-03-05
Generic embeds

Schaeff:
- Make embeds generic (especially to be able to unpack to any constant number of bits)
- In the process, stop using `FunctionCall` with function keys after inlining (for embed calls that cannot be inlined) and instead rewrite these calls to a new `EmbedCall` which points to the Embed itself. This enables simplifying a lot of `zir` because we don't need modules and symbols anymore, as all calls link directly to embeds. This PR also applied this simplification.
- The cost of this is a new variant for `TypedExpressionList` which is internal. Other `FunctionCall`s in expressions do not need to be changed, as we rewrite all embed calls to `MultipleDefinition`. This ends up being quite compact.


dark64 (approved):
LGTM, a really nice one!

PR #758, opened 2021-03-09
Use write_all to write the whole pk in cli_setup()

aitsvet:


Schaeff (approved):
Great! The CI failure is unrelated and fixed on develop

PR #761, opened 2021-03-16
Constant range check

Schaeff:
Current range checks `x < y` are computed using a technique which is not complete (x and y must be small). When `x` or `y` is a compile-time constant, there actually are more efficient, and complete, implementations.

Implement range checks against constants using the technique in A.3.2.2 of the sapling paper.

Note that we already use this technique for range checks against `p - 1` which are required for dynamic range checks.


zokrates_core/src/flatten/mod.rs
```diff
@@ -269,12 +283,93 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+
+        res
+    }
+
+    /// Enforce a range check against a constant: the range check isn't verified iff a constraint will fail
+    ///
+    /// # Arguments
+    ///
+    /// * `statements_flattened` - Vector where new flattened statements can be added.
+    /// * `e` - the `FlatExpression` that's being checked against the range.
```
dark64: outdated docs? `a` and `b` should be explained here

dark64 (approved):

PR #762, opened 2021-03-16
Add negative and positive unary operators

Schaeff:
The goal is to be able to write `-42` instead of `0 - 42`.

The positive operator is added for completeness.

This applies to field and uint.

Todo:
- [x] Enforce precedence
- [x] More precedence tests


zokrates_parser/src/zokrates.pest
```diff
@@ -84,7 +86,8 @@ spread_or_expression = { spread | expression }
 range_or_expression = { range | expression }
 array_initializer_expression = { "[" ~ expression ~ ";" ~ constant ~ "]" }
 
-unary_expression = { op_unary ~ term }
+signeable_expression = { "(" ~ expression ~ ")" | primary_expression }
+exponent_expression = { "(" ~ expression ~ ")" | primary_expression }
```
JacobEberhardt, line 95: `signeable_expression` is redundant with 'exponent_expression' and not referred to elsewhere in the grammar. Why is it needed?

Schaeff: I thought I removed that! Good shout!

Schaeff: It is not and I actually removed it just before you posted this :p

Schaeff: I don't know why this doesn't show as outdated, maybe have a look at the latest diff here and close this if you're cool!

Schaeff:
@dark64 this is finally ready for a review

dark64 (approved):
LGTM!

Issue #767, opened 2021-03-19
Add CHANGELOG


Schaeff:
Any PR should add to the [changelog](http://keepachangelog.com/). It would be good to check that in CI also.

PR #771, opened 2021-03-22
Add changelog scripts

dark64:
https://github.com/Zokrates/ZoKrates/issues/767


changelogs/README.md
```diff
@@ -0,0 +1,14 @@
...
+The file should follow the naming convention of `pr-username` and the contents of the file 
+should be your text for the changelog.
+
+### Example
+```
+changelogs/unreleased   # folder to place changelogs
+    101-username        # your changelog file (pull request #101 by @username)
+```
+
+If a pull request does not warrant a changelog, the CI check for a changelog can be skipped by applying a `changelog-not-required` label.
```
Schaeff: I think a decision factor for this is whether the change is user-facing or not.

changelogs/README.md
```diff
@@ -0,0 +1,14 @@
+## Adding a changelog
+
+Pull request authors are expected to include a changelog file explaining the changes introduced by the pull request. 
+The changelog file should be a new file created in the `changelogs/unreleased` folder.
```
Schaeff, line 4: What's the advantage of having changelogs in different files which are then combined rather than modifying the changelog directly?

dark64: It could cause conflicts a lot, here's a good explanation https://docs.gitlab.com/ee/development/changelog.html#history-and-reasoning, also doing CI checks is easier this way

Schaeff: I see. Some things I'm wondering then:
- how do people know the format of these individual changelog files? I think your source has a cli that takes care of that, seems a bit overkill for us?
- Would we realistically run into a lot of the conflicts this aims to avoid? And even if they happen, are they hard to fix?
- Isn't it possible to just check that changelog file changed? Or are you talking about format checking?

dark64: In our case, a changelog entry is just a raw text explaining the changes (no specific format). Gitlab uses a CLI tool to generate YAML files and it needs a tool to parse them later to generate an actual changelog, which I wanted to avoid as it is an overkill in our case, I agree). I think conflicts should be easy to fix, but we should have a way to avoid them completely, as they might add unnecessary delays.

Another problem is if we modify a changelog directly, we cannot make sure it follows the same format (we can check if a file changed in CI using git diff, but then the actual format must be checked manually by us on each PR that modified a changelog)

Schaeff: Alright, let's give this a go then!

scripts/changelog.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
Schaeff, line 1: This PR should make it clear when this is invoked

Schaeff (approved):

Schaeff:
LGTM, requires pulling develop

PR #774, opened 2021-03-24
Wait for tests to finish running

Schaeff:
The tests are just creating the handler now. This is already done in `zokrates_test`. I'm a bit surprised that there is no lint catching these kinds of things...


Schaeff:
@dark64 the travis fail can be ignored, I opened the PR against master at first

dark64 (approved):

Issue #785, opened 2021-03-29
Fix masking in uint left shifting


dark64:
Replace a uint max constant `0xffffffff` with `2_u128.pow(bitwidth) - 1` on `LeftShift` and `Not` expressions because current implementation will strip higher bits on wider unsigned integers eg. u64

PR #786, opened 2021-03-29
Use mask based on bitwidth instead of u32 max constant

dark64:
Fixed https://github.com/Zokrates/ZoKrates/issues/785

CI failing because of wasm-bindgen dependencies (there is a fix already for this, hopefully lands soon)

### DIFF
--- a/zokrates_core/src/static_analysis/propagation.rs
+++ b/zokrates_core/src/static_analysis/propagation.rs
@@ -663,7 +663,9 @@ impl<'ast, T: Field> Folder<'ast, T> for Propagator<'ast, T> {
                 match (e.into_inner(), by) {
                     (UExpressionInner::Value(v), FieldElementExpression::Number(by)) => {
                         let by_as_usize = by.to_dec_string().parse::<usize>().unwrap();
-                        UExpressionInner::Value((v << by_as_usize) & 0xffffffff)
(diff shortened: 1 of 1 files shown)


zokrates_core/src/static_analysis/propagation.rs
```diff
@@ -718,7 +720,9 @@ impl<'ast, T: Field> Folder<'ast, T> for Propagator<'ast, T> {
             UExpressionInner::Not(box e) => {
                 let e = self.fold_uint_expression(e).into_inner();
                 match e {
-                    UExpressionInner::Value(v) => UExpressionInner::Value((!v) & 0xffffffff),
+                    UExpressionInner::Value(v) => {
```
Schaeff, line 723: small format fix: these curly braces are not required

dark64: hmm interesting I don't remember adding those, it seems to have been added by `cargo fmt`

Schaeff: interesting, can you confirm that?

dark64: yep, just tested it

Schaeff (approved):

PR #792, opened 2021-04-02
Introduce constant definitions

### LINKED ISSUE #778: Introduce a syntactic sugar to define constants
We should introduce some kind of syntactic sugar to provide a way to define global constants

```
const EXAMPLE_CONST 0xDEADBEEF
```

I'm not sure if this syntax fits the language best, so we should discuss this further.

dark64:
https://github.com/Zokrates/ZoKrates/issues/778


zokrates_parser/src/zokrates.pest
```diff
@@ -1,5 +1,5 @@
 
-file = { SOI ~ NEWLINE* ~ pragma? ~ NEWLINE* ~ import_directive* ~ NEWLINE* ~ ty_struct_definition* ~ NEWLINE* ~ function_definition* ~ EOI }
+file = { SOI ~ NEWLINE* ~ pragma? ~ NEWLINE* ~ import_directive* ~ NEWLINE* ~ ty_struct_definition* ~ NEWLINE* ~ const_definition* ~ NEWLINE* ~ function_definition* ~ EOI }
```
Schaeff, line 2: Let's take some time to think of the implications of this ordering (imports > structs > constants > functions) and see if we can relax some of it. Maybe in a separate issue.

dark64: Yes, I was thinking about changing this, we should discuss this further

zokrates_book/src/language/constants.md
```diff
@@ -0,0 +1,17 @@
+## Constants
+
+Constants must be globally defined outside all other scopes by using a `const` keyword. Constants can be set only to a constant expression.
+
+```zokrates
+{{#include ../../../zokrates_cli/examples/book/constant_definition.zok}}
+```
+
+The value of a constant can't be changed through reassignment, and it can't be redeclared.  Constants are essentially inlined wherever they are used, meaning that they are copied directly into the relevant context when used.
```
Schaeff:
```suggestion
The value of a constant can't be changed through reassignment, and it can't be redeclared.
```

I think the second part is an implementation detail

zokrates_book/src/language/constants.md
```diff
@@ -0,0 +1,17 @@
...
+
+Constants must be explicitly typed. One can reference other constants inside the expression, as long as the referenced constant is defined before the constant.
```
Schaeff:
```suggestion
Constants must be explicitly typed. One can reference other constants inside the expression, as long as the referenced constant is already defined.
```

zokrates_core/src/absy/mod.rs
```diff
@@ -51,20 +51,35 @@ pub struct SymbolDeclaration<'ast> {
     pub symbol: Symbol<'ast>,
 }
 
+#[derive(PartialEq, Clone)]
+pub enum SymbolDefinition<'ast> {
+    Struct(StructDefinitionNode<'ast>),
+    Constant(Box<ConstantDefinitionNode<'ast>>),
```
Schaeff: Is boxing required here?

zokrates_core/src/absy/mod.rs
```diff
@@ -146,6 +164,31 @@ impl<'ast> fmt::Display for StructDefinitionField<'ast> {
 
 type StructDefinitionFieldNode<'ast> = Node<StructDefinitionField<'ast>>;
 
+#[derive(Clone, PartialEq)]
+pub struct ConstantDefinition<'ast> {
```
Schaeff, line 173: In line with the implementation for Functions and Structs, the id should not be here, but in the `SymbolDeclaration`

zokrates_core/src/absy/mod.rs
```diff
@@ -73,8 +88,11 @@ impl<'ast> fmt::Debug for Symbol<'ast> {
 impl<'ast> fmt::Display for SymbolDeclaration<'ast> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self.symbol {
-            Symbol::HereType(ref t) => write!(f, "struct {} {}", self.id, t),
-            Symbol::HereFunction(ref fun) => write!(f, "def {}{}", self.id, fun),
+            Symbol::Here(ref kind) => match kind {
+                SymbolDefinition::Struct(t) => write!(f, "struct {} {}", self.id, t),
+                SymbolDefinition::Constant(c) => write!(f, "{}", c),
```
Schaeff: This should use `self.id` rather than getting is from `ConstantDefinition`

zokrates_core/src/semantics.rs
```diff
@@ -72,14 +73,14 @@ struct SymbolUnifier<'ast> {
 }
 
 impl<'ast> SymbolUnifier<'ast> {
-    fn insert_type<S: Into<String>>(&mut self, id: S) -> bool {
+    fn insert_symbol<S: Into<String>>(&mut self, id: S, ty: SymbolType<'ast>) -> bool {
```
Schaeff: It would be clearer to either have one function for each type of symbol, or a single function for all symbols. With this change we have one for functions, and one for Constants and Types.

zokrates_core/src/semantics.rs
```diff
@@ -72,14 +73,14 @@ struct SymbolUnifier<'ast> {
...
         let s_type = self.symbols.entry(id.into());
         match s_type {
             // if anything is already called `id`, we cannot introduce this type
```
Schaeff: This comment is then outdated because it can also be a const. I tend to prefer one function per symbol type, then it's clear what the rules are for each.

zokrates_core/src/semantics.rs
```diff
@@ -229,6 +230,7 @@ impl<'ast, T: Field> FunctionQuery<'ast, T> {
 pub struct ScopedVariable<'ast, T> {
     id: Variable<'ast, T>,
     level: usize,
+    constant: bool,
```
Schaeff: Is there a way to remove this field, and instead use `level := 0` for constants? The semantic checker has the level in its state, right now set to 0 in each function scope, but could be 0 outside functions and 1 inside with a very minor change.

```
//level = 0
const FOO = 42

def foo():
   // enter function, level = 1
   field a = 42
   for ...:
      // enter for loop scope, level = 2
   endfor
   // exit for loop, level = 1
   return

// exit function scope, level = 0
```

zokrates_core/src/semantics.rs
```diff
@@ -308,6 +310,50 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+        &mut self,
+        c: ConstantDefinitionNode<'ast>,
+        module_id: &ModuleId,
+        types: &TypeMap<'ast>,
+    ) -> Result<TypedConstant<'ast, T>, ErrorInner> {
+        let pos = c.pos();
+        let ty = self.check_type(c.value.ty.clone(), module_id, &types)?;
+        let checked_expr = self.check_expression(c.value.expression.clone(), module_id, types)?;
+
+        match ty.clone() {
```
Schaeff: This clone could happen in the match arms instead

zokrates_core/src/semantics.rs
```diff
@@ -308,6 +310,50 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+            Type::FieldElement => {
+                FieldElementExpression::try_from_typed(checked_expr).map(TypedExpression::from)
+            }
+            Type::Boolean => {
+                BooleanExpression::try_from_typed(checked_expr).map(TypedExpression::from)
+            }
+            Type::Uint(bitwidth) => {
+                UExpression::try_from_typed(checked_expr, bitwidth).map(TypedExpression::from)
+            }
+            Type::Array(array_ty) => ArrayExpression::try_from_typed(checked_expr, *array_ty.ty)
```
Schaeff: Here `array_ty` can be cloned

zokrates_core/src/semantics.rs
```diff
@@ -369,76 +416,107 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
-                            module_id.to_path_buf(),
-                            declaration.id,
-                        )
-                        .signature(funct.signature.clone()),
-                        TypedFunctionSymbol::Here(funct),
-                    );
+                                true => {}
+                            };
+                            constants.insert(declaration.id, TypedConstantSymbol::Here(c.clone()));
+                            self.insert_into_scope(Variable::with_id_and_type(c.id, c.ty), true);
```
Schaeff: Here `self.level` should be 0 by construction because we are in the first scope, cf comment before to remove the `constant` field here.

zokrates_core/src/semantics.rs
```diff
@@ -470,8 +548,21 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+                        let const_candidate = state
+                            .typed_modules
+                            .get(&import.module_id)
+                            .unwrap()
+                            .constants
+                            .as_ref()
+                            .and_then(|tc| tc.get(import.symbol_id))
+                            .and_then(|sym| match sym {
+                                TypedConstantSymbol::Here(tc) => Some(tc),
+                                _ => None,
```
Schaeff: Can you explain why `TypedConstantSymbol::There` leads to `None`? What happens if we import an imported constant? In the case of Types we resolve these chained imports, is the same happening here?

zokrates_core/src/semantics.rs
```diff
@@ -614,12 +725,14 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
                         state,
                         &mut checked_functions,
+                        &mut checked_constants,
                         &mut symbol_unifier,
                     )?
                 }
 
                 Some(TypedModule {
                     functions: checked_functions,
+                    constants: Some(checked_constants).filter(|m| !m.is_empty()),
```
Schaeff: Why `Option<HashMap>` rather than `HashMap`?

zokrates_core/src/semantics.rs
```diff
@@ -1387,10 +1498,16 @@ impl<'ast, T: Field> Checker<'ast, T> {
         // check that the assignee is declared
         match assignee.value {
             Assignee::Identifier(variable_name) => match self.get_scope(&variable_name) {
-                Some(var) => Ok(TypedAssignee::Identifier(Variable::with_id_and_type(
-                    variable_name,
-                    var.id._type.clone(),
-                ))),
+                Some(var) => match var.constant {
```
Schaeff: Could have 
```
fn is_constant(&self) -> bool {
   self.level == 0
}
```

zokrates_core/src/static_analysis/constant_inliner.rs
```diff
@@ -0,0 +1,856 @@
...
+            .unwrap()
+            .constants
+            .as_ref()
+            .and_then(|c| c.get(id.clone().try_into().unwrap()))
+            .cloned()
+            .and_then(|tc| {
+                let symbol = self.fold_constant_symbol(tc);
+                match symbol {
+                    TypedConstantSymbol::Here(tc) => Some(tc),
+                    _ => None,
```
Schaeff: Same here, is this case reachable?

zokrates_core/src/static_analysis/propagation.rs
```diff
@@ -250,6 +250,7 @@ impl<'ast, 'a, T: Field> ResultFolder<'ast, T> for Propagator<'ast, 'a, T> {
                     }
                 })
                 .collect::<Result<_, _>>()?,
+            constants: m.constants,
```
Schaeff:
```suggestion
            ..m
```

zokrates_core/src/typed_absy/result_folder.rs
```diff
@@ -813,6 +813,7 @@ pub fn fold_module<'ast, T: Field, F: ResultFolder<'ast, T>>(
             .into_iter()
             .map(|(key, fun)| f.fold_function_symbol(fun).map(|f| (key, f)))
             .collect::<Result<_, _>>()?,
+        constants: p.constants,
```
Schaeff: For consistency with the other folder, this should visit the constants as well. Or neither should, because maybe there can be weird side effects in some analysers, and constant inlining runs first anyways

Schaeff (changes requested):
Makes sense in general, made some suggestions and will have a closer look at the inliner

zokrates_core/src/semantics.rs
dark64: Hmm, I see your point, does it make sense to just wrap `insert_symbol` in `insert_type` and `insert_constant`? The implementation of both functions is exactly the same, which is the reason I refactored it to `insert_symbol`.

zokrates_core/src/semantics.rs
dark64: Good catch, this flag is unnecessary.

zokrates_core/src/semantics.rs
dark64: `fold_constant_symbol` recursively finds `Here` so in the example where we import an imported constant, it does the recursion until we are in the module that actually defined the constant (`There` -> `There` -> `Here`). I guess the default arm should be  `unreachable`.

zokrates_core/src/static_analysis/constant_inliner.rs
dark64: `fold_constant_symbol` recursively finds `Here` so in the example where we import an imported constant, it does the recursion until we are in the module that actually defined the constant (`There` -> `There` -> `Here`). So in this case, the default arm should be `unreachable`, not `None`.

zokrates_core/src/absy/mod.rs
dark64: clippy is complaining because of the large size difference between variants

dark64: clippy was complaining because of the large size difference between variants

zokrates_parser/src/zokrates.pest
dark64: @Schaeff should we do this in a separate PR or this one? might be a bigger change if we decide to relax imports as well

zokrates_core/src/semantics.rs
Schaeff: Right, I see the argument of the redundant implementation. I would go for clarity here, in general the symbols have different naming rules, so having a redundant implementation doesnt seem like a big deal to me.

Alternatively having a single function for all of them could work as well (include `insert_function` in `insert_symbol`)

zokrates_parser/src/zokrates.pest
Schaeff: Let's do it in another one, this one is quite big already

zokrates_core/src/semantics.rs
dark64: Pretty much yes, https://github.com/Zokrates/ZoKrates/blob/9f69ea37719a16ad8fc1887645514cb9ac5e31bb/zokrates_core/src/semantics.rs#L655

zokrates_core/src/absy/mod.rs
Schaeff: I see, is it still complaining after removing the id? If so, think it's fine to turn off that clippy rule here for readability.

zokrates_core/src/static_analysis/constant_inliner.rs
Schaeff: I see. In order to avoid unreachable match arms, maybe it would make sense to extract the canonicalisation process to return what's inside `Here`, rather than the `TypedConstantSymbol`. Then the runtime assumption is that any import chains ends in a `Here`, which to me sounds clearer than the runtime assumption that the `There` match arm is unreachable.

Schaeff: This was not possible in the function call inliner, because the resolving process can return either `Here` or `Flat`, and representing these two options would require defining another enum, which felt like overkill, but in the case of constants there's no `Flat` so I feel it would work?

zokrates_core/src/semantics.rs
Schaeff: Right, that's where I think it would be ideal to have having this map be canonical and store c instead of Here(c) would remove a bunch of these unreachable cases. I'm not totally sure that it's possible though

zokrates_core/src/absy/mod.rs
dark64: Yes, I'll add a clippy allow rule for this

zokrates_core/src/semantics.rs
dark64: As we discussed, we can store only the Type of the constant here, because the value is never used in the semantic phase. This removes the unreachable cases.

zokrates_core/src/typed_absy/folder.rs
```diff
@@ -711,6 +719,19 @@ pub fn fold_struct_expression<'ast, T: Field, F: Folder<'ast, T>>(
     }
 }
 
+pub fn fold_constant_symbol<'ast, T: Field, F: Folder<'ast, T>>(
+    f: &mut F,
+    p: TypedConstantSymbol<'ast, T>,
+) -> TypedConstantSymbol<'ast, T> {
+    match p {
+        TypedConstantSymbol::Here(tc) => TypedConstantSymbol::Here(TypedConstant {
```
Schaeff: We should define `fold_constant` on `TypedConstant` in the trait and use that here

Schaeff: Same in result_folder

zokrates_core/src/static_analysis/constant_inliner.rs
```diff
@@ -0,0 +1,848 @@
...
+        TypedConstantSymbol::Here(tc)
+    }
+
+    fn fold_field_expression(
+        &mut self,
+        e: FieldElementExpression<'ast, T>,
+    ) -> FieldElementExpression<'ast, T> {
+        match e {
+            FieldElementExpression::Identifier(ref id) => match self.get_constant(id) {
+                Some(c) => fold_field_expression(self, c.expression.try_into().unwrap()),
```
Schaeff: It would be clearer to do something like

```
let c = self.fold_constant(c)
c.try_into().unwrap()
```

Using the `fold_constant` trait method suggested in the other comment

Schaeff: if this `TryInto` isn't implemented, it would make sense to add it: downcast a constant to an expression of a given type. It can reuse the `TryInto` defined for `TypedExpression`, so that we don't access c.expression here. Ideally this struct field would be private and everything happens with methods.

Schaeff (changes requested):
The inliner is fine, some logic just needs to be extracted into the folder traits

zokrates_core/src/static_analysis/constant_inliner.rs
```diff
@@ -108,7 +100,10 @@ impl<'ast, T: Field> Folder<'ast, T> for ConstantInliner<'ast, T> {
     ) -> FieldElementExpression<'ast, T> {
         match e {
             FieldElementExpression::Identifier(ref id) => match self.get_constant(id) {
-                Some(c) => fold_field_expression(self, c.expression.try_into().unwrap()),
+                Some(c) => {
+                    let c = self.fold_constant(c);
+                    fold_field_expression(self, c.try_into().unwrap())
```
Schaeff: Unless I'm missing something, folding isnt required here, it was done in the previous line.

Schaeff: this could even be

```
Some(c) => self.fold_constant(c).try_into().unwrap()
```

zokrates_core/src/typed_absy/result_folder.rs
```diff
@@ -800,15 +807,22 @@ pub fn fold_struct_expression<'ast, T: Field, F: ResultFolder<'ast, T>>(
     })
 }
 
+pub fn fold_constant<'ast, T: Field, F: ResultFolder<'ast, T>>(
+    f: &mut F,
+    c: TypedConstant<'ast, T>,
+) -> Result<TypedConstant<'ast, T>, F::Error> {
+    Ok(TypedConstant {
+        expression: f.fold_expression(c.expression)?,
+        ..c
```
Schaeff: I think the type also needs to be folded: array sizes can be any expression, including a constant

Schaeff: Or the type needs to be removed from `TypedConstant`: isn't it always `expression.get_type()` ?

dark64: Nice catch, it is! Does it even make sense then to have `TypedConstant` struct instead of just using `TypedExpression` directly?

Schaeff: I think it doesn't hurt

Schaeff: Hmm I see yeah this `Clone` bound is annoying... You're right let's keep this type in there...

dark64: Yeah :/

Schaeff: Just make sure it gets visited in the folders etc

Schaeff (approved):
👍

PR #800, opened 2021-04-08
Allow underscore before type suffix

### LINKED ISSUE #784: Allow underscore before type suffix
`42_u32` should be accepted for `42u32`

dark64:
Adds an optional underscore before type suffix, `42_u32` is the same as `42u32`

Closes https://github.com/Zokrates/ZoKrates/issues/784

### DIFF
--- a/changelogs/unreleased/800-dark64
+++ b/changelogs/unreleased/800-dark64
@@ -1,0 +1,1 @@
+Allow optional underscore before type suffix (e.g. `42_u32`)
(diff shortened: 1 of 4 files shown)


Schaeff (changes requested):
Would be safe to have some tests for this (can be in `zokrates_cli/examples`), as well as update the textmate highlighter.

dark64:
> Would be safe to have some tests for this (can be in `zokrates_cli/examples`), as well as update the textmate highlighter.

Textmate seems to already have this, will write some tests!

Schaeff (approved):

PR #809, opened 2021-04-12
Allow importing of multiple symbols in a single import statement

### LINKED ISSUE #808: Import multiple symbols in a single from-import statement
## Abstract
From-import directive should support multiple import symbols. This can be achieved as a syntactic sugar.

To avoid cases like the following:
```zokrates
from "./module.zok" import func1
from "./module.zok" import func2
from "./module.zok" import func3
...
```
one should be able to import all needed functions in a single line:
```zokrates
from "./module.zok" import func1, func2, func3
```

## Motivation
This is a common feature in most languages, as it is easy to implement, we should support this as well.

## Specification
The implementation should cover all of the following cases:

```zokrates
# without aliases
from "./source.zok" import func1, func2

# with aliasing
from "./source.zok" import func1 as f1, func2
from "./source.zok" import func1 as f1, func2 as f2
```

## Backwards Compatibility
No breaking changes.

dark64:
Closes https://github.com/Zokrates/ZoKrates/issues/808

### DIFF
--- a/changelogs/unreleased/809-dark64
+++ b/changelogs/unreleased/809-dark64
@@ -1,0 +1,1 @@
+Add ability to import multiple symbols in a single import statement
--- a/changelogs/unreleased/809-dark64
+++ b/changelogs/unreleased/809-dark64
@@ -1,1 +1,1 @@
-Add ability to import multiple symbols in a single import statement
+Add the ability to import multiple symbols in a single import statement
--- a/zokrates_book/src/language/imports.md
+++ b/zokrates_book/src/language/imports.md
@@ -13,7 +13,7 @@ from "./path/to/my/module" import MySymbol
 // `MySymbol` is now in scope.
 ```
 
-To import multiple modules with a single import statement, separate the module names with commas:
+To import multiple symbols with a single import statement, separate the symbols names with commas:
 
 ```zokrates
 from "./path/to/my/module" import MySymbol, MyOtherSymbol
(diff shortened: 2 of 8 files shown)


changelogs/unreleased/809-dark64
```diff
@@ -0,0 +1 @@
+Add ability to import multiple symbols in a single import statement
```
Schaeff:
```suggestion
Add the ability to import multiple symbols in a single import statement
```

zokrates_book/src/language/imports.md
```diff
@@ -13,6 +13,12 @@ from "./path/to/my/module" import MySymbol
 // `MySymbol` is now in scope.
 ```
 
+To import multiple modules with a single import statement, separate the module names with commas:
```
Schaeff:
```suggestion
To import multiple symbols with a single import statement, separate the symbol names with commas:
```

zokrates_core/src/imports.rs
```diff
@@ -56,6 +56,25 @@ impl From<io::Error> for Error {
     }
 }
 
+#[derive(PartialEq, Clone)]
+pub enum ImportKind<'ast> {
```
Schaeff: Could this be simplified to only the Multiple variant or removed completely?

Schaeff (changes requested):

zokrates_book/src/language/imports.md
dark64: Oh, how did I not notice this :facepalm: Nice catch!

Schaeff (approved):

PR #820, opened 2021-04-14
Improve matching of overloaded functions

### LINKED ISSUE #807: Improve matching of overloaded functions
## Description

If overloaded functions have concrete values in signature, the compiler should be able to determine which overloaded function to call. Currently, the compiler fails with `"x" conflicts with another symbol`. This error should be thrown only if one of the overloaded functions is generic.

## Environment

- Compiler version: develop
- Operating system: Ubuntu 18.04 LTS

## Steps to Reproduce

```zokrates
def test() -> field[2]:
    return [1, 2]

def test() -> field[3]:
    return [1, 2, 3]

def main():
    field[2] t = test() // we should know here which function to pick
    return
```

dark64:
Solves https://github.com/Zokrates/ZoKrates/issues/807

### DIFF
--- a/zokrates_core/src/typed_absy/types.rs
+++ b/zokrates_core/src/typed_absy/types.rs
@@ -17,22 +17,27 @@ pub enum Constant<'ast> {
     Concrete(u32),
 }
 
-// At this stage we want all constants to be equal
 impl<'ast> PartialEq for Constant<'ast> {
-    fn eq(&self, _: &Self) -> bool {
-        true
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Constant::Generic(_), _) | (_, Constant::Generic(_)) => true,
+            (Constant::Concrete(a), Constant::Concrete(b)) => a == b,
+        }
     }
 }
 
 impl<'ast> PartialOrd for Constant<'ast> {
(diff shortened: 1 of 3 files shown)


dark64:
@Schaeff is changelog required for this stuff?

Schaeff:
I think so, as this changes behaviour that users would notice. Maybe more than a fix it's an improvement.

zokrates_core/src/typed_absy/types.rs
```diff
@@ -17,22 +17,27 @@ pub enum Constant<'ast> {
...
 }
 
 impl<'ast> Ord for Constant<'ast> {
-    fn cmp(&self, _: &Self) -> std::cmp::Ordering {
-        std::cmp::Ordering::Equal
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        match (self, other) {
+            (Constant::Concrete(a), Constant::Concrete(b)) => a.cmp(b),
+            _ => std::cmp::Ordering::Equal,
+        }
```
Schaeff, line 40: The Hash implementation further down should also be updated or removed for consistency.

Schaeff:
I think #826 ends up replacing this, just need to make sure the tests introduced here are covered there.

Schaeff:
This can be closed

PR #822, opened 2021-04-15
Consider input count and generics count in semantic checker

### LINKED ISSUE #821: Function selector is not strict enough
## Description

Some calls are being accepted, even though they don't have the correct number of generics or inputs.

## Environment

- Compiler version: develop
- Operating system: macos

## Steps to Reproduce

This is accepted but shouldn't be

```
def foo() -> field:
    return 1

def main() -> field:
    return foo(42)
```

This panics but should fail gracefully

```
def foo() -> field:
    return 1

def main() -> field:
    return foo::<42>()
```

Schaeff:
Closes #821 

Also add panics when inlining if these things don't match.

### DIFF
--- a/changelogs/unreleased/822-schaeff
+++ b/changelogs/unreleased/822-schaeff
@@ -1,0 +1,1 @@
+Make function selection stricter in function calls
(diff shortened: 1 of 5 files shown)


dark64 (approved):
LGTM

PR #823, opened 2021-04-15
Detect assertion failures at compile time on constant expressions

### LINKED ISSUE #812: Assertions should error out if the value is evaluated to false
## Abstract
Assertion should error out on compilation if the value is evaluated to `false`.

## Motivation
There are cases where input values must be asserted e.g.:

```zokrates
def func<N>(field[N] inputs) -> bool:
    assert(N > 0 && N <= 6) // assert that the N is in [1, 6]
    .... // some code
    return true

def main():
    bool b = func([1, 2, 3, 4, 5, 6, 7]) // this should not be allowed
    return
```

## Specification

If an assertion `assert(e)` where `e` is an expression gets evaluated to `assert(false)`, the compiler should gracefully exit with an appropriate message.

## Backwards Compatibility
Should be compatible.

dark64:
Closes https://github.com/Zokrates/ZoKrates/issues/812


zokrates_cli/examples/book/assert.zok
```diff
@@ -1,3 +1,3 @@
 def main() -> ():
-	assert(1f == 2f)
+	assert(2f == 2f)
```
dark64: @Schaeff I'm not sure about this example, I had to change it to a passing one otherwise the tests will fail. Any thoughts?

Schaeff: Yes that's okay. Maybe put `1 + 1 == 2`

Schaeff (approved):

PR #826, opened 2021-04-19
Smarter function call matching

Schaeff:
To be merged after #822 

- Use canonical generics to detect equivalence of `<N>(field[N])` and `<P>(field[P])`
- Accept all functions definitions for which there can be a non ambiguous call
- Detect ambiguous calls on each call site


dark64 (approved):
Nice one!

PR #828, opened 2021-04-21
Remove space between file name and position to make VSCode experience smoother

Schaeff:
I noticed that when the file name and line/column are joined, cmd+click on VSCode jumps to the location of the error :)


dark64 (approved):
Nice one!

PR #830, opened 2021-04-22
Fix stack overflow on fmt::Debug for zir::BooleanExpression

dark64:
Implemented a `fmt::Debug` trait for `zir::BooleanExpression`. Previous implementation `write!(f, "{:?}", self)` was causing a stack overflow.


Schaeff (approved):
🤦 🆗

PR #834, opened 2021-04-26
Fix integer inference on repeat operator

### LINKED ISSUE #804: Inline multi-dimensional array repeat not behaving properly when passed to a generic function
## Description

Compilation fails with `Failed to unroll or inline program. Check that main function arguments aren't used as array size or for-loop bounds`. Seems to be a compiler issue.

## Environment

- Compiler version: develop
- Operating system: Ubuntu 18.04 LTS

## Steps to Reproduce

```zokrates
def test<N>(field[N][N] t) -> field[N][N]:
    return t

def main() -> field[1][1]:
    return test([[0; 1]; 1]) // or just [[0]; 1], works with [[0]]
```

Note: Assigning the array to a variable and then passing it to the `test` function works, I have not found any issues with single dimension arrays.

Schaeff:
The type of the repeated element should be the inferred type, not the target type.

Closes #804

### DIFF
--- a/zokrates_core/src/typed_absy/integer.rs
+++ b/zokrates_core/src/typed_absy/integer.rs
@@ -515,7 +515,7 @@ impl<'ast, T: Field> ArrayExpression<'ast, T> {
                 match target_inner_ty.clone() {
                     Type::Int => Ok(ArrayExpressionInner::Repeat(box e, box count)
                         .annotate(Type::Int, array_ty.size)),
-                    // try to convert the repeated element to the target type
+                    // try to align the repeated element to the target type
                     t => TypedExpression::align_to_type(e, t)
                         .map(|e| {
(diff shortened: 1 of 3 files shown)


zokrates_core/src/typed_absy/integer.rs
```diff
@@ -518,8 +518,10 @@ impl<'ast, T: Field> ArrayExpression<'ast, T> {
                     // try to convert the repeated element to the target type
```
dark64: is this comment outdated now? should be the inferred type, not the target type I assume

Schaeff: Good catch, the right word here is align, not convert: aligning `int[42]` to `field[N]` yields `field[42]`

dark64 (approved):

PR #836, opened 2021-04-27
Migration to CircleCI

### LINKED ISSUE #818: Migrate cross compilation to CircleCI
## Description

Travis times out on our cross compilation jobs. It's time to move away from it.

dark64:
https://github.com/Zokrates/ZoKrates/issues/818

Note: libsnark is not included in the builds as `cross` docker images do not have required dependencies (`openssl` for `git2-rs`, `libcrypto` etc). This would require us to create docker images for each target, extending  `cross` base images with required dependencies.


rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly-2021-04-25
```
Schaeff, line 1: Do we need both this file and the toml one?

.circleci/config.yml
```diff
@@ -125,8 +132,88 @@ jobs:
...
+          at: /tmp/artifacts
+      - run:
+          name: "Publish artifacts on GitHub"
+          command: |
+            go get github.com/github-release/github-release
+            github-release release \
+                -s ${GH_TOKEN} \
+                -u ${CIRCLE_PROJECT_USERNAME} \
+                -r ${CIRCLE_PROJECT_REPONAME} \
+                -t ${CIRCLE_TAG} || true
```
Schaeff, line 166: why `|| true`?

Schaeff (approved):
Really nicely put together, very minor comments!

.circleci/config.yml
dark64: If for some reason a release exists (created manually), I don't want this to fail in that case as it takes a while to rerun the whole pipeline. This probably won't happen, but it should be safe to just upload the artifacts? A bit of an edge case, maybe not needed at all.

rust-toolchain
dark64: As the new toml format is introduced recently, the `zokrates/env` image has a bit older rustup which I believe does not support the new format, this is just a legacy thing that will be removed once we update everything else.

.circleci/config.yml
Schaeff: That's fine, I thought this could be a typo!

PR #847, opened 2021-05-05
Remove side effects on complex types

Schaeff:
When we removed SSA for complex types, the `RedefinitionOptimizer` was left there and it caused complex types to be in practice passed by mutable reference to functions. Fix that so that we always pass by value.


dark64 (approved):
nice find :+1:

PR #852, opened 2021-05-05
Fix flattening of uint subtraction for values smaller than the target

Schaeff:
The uint optimizer uses `max(right_bitwidth, range)` but at flattening we're using just `right_bitwidth`, hence shifting by something that is not a multiple of `2**range` for constants.

Align flattening to uint optimizer, add tests.


dark64 (approved):
:100:

Issue #851, opened 2021-05-05
Declaring constants before structs is a parser error


axic:
```
const u32 A = 0

struct State {
	u32 x
}
```

```
Compiling const.zok

Compilation failed:

const.zok: --> 3:1
  |
3 | struct State {␊
  | ^---
  |
  = expected EOI, function_definition, or const_definition
```

Is there a reason for this or only to discourage people trying to use constant as array sizes in structs?

I see the pest file has a very strict ordering:
```
file = { SOI ~ NEWLINE* ~ pragma? ~ NEWLINE* ~ import_directive* ~ NEWLINE* ~ ty_struct_definition* ~ NEWLINE* ~ const_definition* ~ NEWLINE* ~ function_definition* ~ EOI }
```

dark64:
Hello @axic, the ordering is very strict right now yes. We are working on easing it a bit. Using constants as array sizes will be possible soon, but is not supported yet.

axic:
Thanks!

As as I side note: I did a quick change in pest just to see how many things break, but managed to trigger a rustc ICE.

```diff
diff --git a/zokrates_parser/src/zokrates.pest b/zokrates_parser/src/zokrates.pest
index 8280154e..bfcc2a63 100644
--- a/zokrates_parser/src/zokrates.pest
+++ b/zokrates_parser/src/zokrates.pest
@@ -1,9 +1,11 @@
 
-file = { SOI ~ NEWLINE* ~ pragma? ~ NEWLINE* ~ import_directive* ~ NEWLINE* ~ ty_struct_definition* ~ NEWLINE* ~ const_definition* ~ NEWLINE* ~ function_definition* ~ EOI }
+file = { SOI ~ NEWLINE* ~ pragma? ~ NEWLINE* ~ import_directive* ~ NEWLINE* ~ const_or_struct_definition* ~ NEWLINE* ~ function_definition* ~ EOI }
 
 pragma = { "#pragma" ~ "curve" ~ curve }
 curve = @{ (ASCII_ALPHANUMERIC | "_") * }
 
+const_or_struct_definition = { ty_struct_definition | const_definition }
+
 import_directive = { main_import_directive | from_import_directive }
 from_import_directive = { "from" ~ "\"" ~ import_source ~ "\"" ~ "import" ~ import_symbol_list ~ NEWLINE* }
[...]

PR #863, opened 2021-05-12
Relax ordering of symbol declarations

### LINKED ISSUE #858: Relax the ordering of symbol declarations
## Abstract

The ordering of symbol declarations is very strict right now. We should relax everything which would allow us to eg. use constants as array sizes.

## Motivation

Required change for other improvements.

## Specification

Change pest grammar and propagate all necessary changes to ast.

## Backwards Compatibility

Should not produce any backwards compatibility issues.

dark64:
https://github.com/Zokrates/ZoKrates/issues/858 https://github.com/Zokrates/ZoKrates/issues/851


zokrates_core/src/absy/from_ast.rs
```diff
@@ -1,66 +1,70 @@
...
+                    let alias = symbol
+                        .alias
+                        .as_ref()
+                        .map(|a| a.span.as_str())
+                        .unwrap_or_else(|| symbol.id.span.as_str());
+
+                    let import = absy::CanonicalImport {
+                        source,
+                        id: absy::SymbolIdentifier::from(symbol.id.span.as_str())
+                            .alias(Some(alias)),
```
Schaeff, line 58: Seems like we're using two approaches here: line 48 we determine the alias: if none, then it's the symbol. Basically removing the option. But here we put it back into an option. 

I suspect we could either have an option down the chain, or just accept that the canonical import always has an alias which, if not user provided, is determined here. I think the latter is fine here, to avoid using unnecessary options down the line.

zokrates_core/src/absy/mod.rs
```diff
@@ -44,61 +42,123 @@ pub struct Program<'ast> {
     pub main: OwnedModuleId,
 }
 
-/// A declaration of a `FunctionSymbol`, be it from an import or a function definition
-#[derive(PartialEq, Clone, Debug)]
+#[derive(Debug, PartialEq, Clone)]
+pub struct SymbolIdentifier<'ast> {
+    pub id: Identifier<'ast>,
+    pub alias: Option<Identifier<'ast>>,
```
Schaeff, line 48: cf previous comment, this would then not be an option

zokrates_core/src/absy/mod.rs
```diff
@@ -44,61 +42,123 @@ pub struct Program<'ast> {
...
+    }
+}
+
+impl<'ast> SymbolIdentifier<'ast> {
+    pub fn alias(mut self, alias: Option<Identifier<'ast>>) -> Self {
+        self.alias = alias;
+        self
+    }
+    pub fn get_alias(&self) -> Identifier<'ast> {
+        self.alias.unwrap_or(self.id)
```
Schaeff, line 63: This is quite neat, but just reading a non-optional alias would be ok here.

zokrates_core/src/absy/mod.rs
```diff
@@ -44,61 +42,123 @@ pub struct Program<'ast> {
...
                     "const {} {} = {}",
                     c.value.ty, self.id, c.value.expression
                 ),
-                SymbolDefinition::Function(func) => write!(f, "def {}{}", self.id, func),
+                SymbolDefinition::Function(ref func) => {
+                    write!(f, "def {}{}", self.id, func)
+                }
             },
-            Symbol::There(ref import) => write!(f, "import {} as {}", import, self.id),
+            Symbol::There(ref i) => write!(f, "{}", i),
```
Schaeff: Am I right that we end up with some information duplication here? In the other match arms, we use `self.id` which is the name of the symbol in this module. This makes me wonder if `i` here doesn't have information it doesn't need? Maybe the alias needs to be kept out of `i` so we only rely on `self.id`?

zokrates_core/src/imports.rs
```diff
@@ -156,253 +68,155 @@ impl Importer {
...
-                            SymbolDeclaration {
-                                id: &alias,
-                                symbol: Symbol::Flat(FlatEmbed::U32FromBits),
-                            }
-                            .start_end(pos.0, pos.1),
-                        );
-                    }
-                    "EMBED/u16_from_bits" => {
-                        let alias = alias.unwrap_or("u16_from_bits");
+        for symbol in destination.symbols {
```
Schaeff: We're basically doing a Map here, this would be more natural written like

```
destination.symbols.into_iter().map(|s| match {...}).collect()
```

zokrates_core/src/imports.rs
```diff
@@ -156,253 +68,155 @@ impl Importer {
...
+                        Error::new(format!("Embed {} not found", s)).with_pos(Some(pos)),
+                    )
+                    .in_file(location)
+                    .into());
+                }
+            },
+            _ => match resolver {
+                Some(res) => match res.resolve(location.to_path_buf(), module_id.to_path_buf()) {
+                    Ok((source, new_location)) => {
+                        let alias = symbol.alias.unwrap_or(
```
Schaeff, line 162: Cf previous comment, isn't the `unwrap_or` unreachable here as we already determined the default alias? Then how do we determine the default alias if not here?

zokrates_core/src/imports.rs
```diff
@@ -156,253 +68,155 @@ impl Importer {
...
-                                .start_end(pos.0, pos.1),
-                            );
-                        }
-                        Err(err) => {
-                            return Err(CompileErrorInner::ImportError(
-                                err.into().with_pos(Some(pos)),
-                            )
-                            .in_file(&location)
-                            .into());
+                        match modules.get(&new_location) {
```
Schaeff, line 178: Not a big deal but the Entry API works quite well in this case:

```
modules
   .entry(new_location.clone())
   .or_insert_with(|| /* the None arm from the current code, without the final assert */)
```

iirc

Schaeff (changes requested):

zokrates_core/src/absy/from_ast.rs
```diff
@@ -1,66 +1,70 @@
...
-            ),
+    match import {
+        pest::ImportDirective::Main(import) => {
+            let span = import.span;
+            let source = Path::new(import.source.span.as_str());
+            let id = "main";
+
+            let import = absy::CanonicalImport {
+                source,
+                id: absy::SymbolIdentifier::from(id).alias(import.alias.map(|a| a.span.as_str())),
```
Schaeff: Oh so that's where the optional alias is required. I would suggest moving the alias detection code from the resolver to here, and remove the optional alias.

Schaeff (approved):
LGTM!

PR #864, opened 2021-05-13
Support the use of constants in declaration types

### LINKED ISSUE #859: Use constant expression as an array size
## Abstract
We should support the following code:

```zokrates
const field N = 2

def main(field[N] n) -> field[N]:
    return n
```

or

```zokrates
const field SIZE = 16

struct State {
    field n[SIZE]
}
```

## Specification

To support this, the constant inliner should reduce any constant expression to a literal. The following code should error out:

```zokrates
const field N = 2

def foo<N>(field[N]): // conflict
    ...
    return
```

## Backwards Compatibility
No incompatibilities.

dark64:
https://github.com/Zokrates/ZoKrates/issues/859


changelogs/unreleased/864-dark64
```diff
@@ -0,0 +1 @@
+Support the use of constants in declaration types
```
Schaeff: declaration types are an internal concept, this should maybe mention struct and function declarations?

zokrates_core/src/semantics.rs
```diff
@@ -831,7 +831,7 @@ impl<'ast, T: Field> Checker<'ast, T> {
         &mut self,
         funct_node: FunctionNode<'ast>,
         module_id: &ModuleId,
-        types: &TypeMap<'ast>,
+        state: &mut State<'ast, T>,
```
Schaeff: Does this need to be a mutable reference?

zokrates_core/src/semantics.rs
```diff
@@ -409,7 +409,7 @@ impl<'ast, T: Field> Checker<'ast, T> {
         for field in s.fields {
             let member_id = field.value.id.to_string();
             match self
-                .check_declaration_type(field.value.ty, module_id, &types, &HashMap::new())
+                .check_declaration_type(field.value.ty, module_id, state, &HashMap::new())
```
Schaeff, line 413: Generic struct types aren't in this PR then?

zokrates_core/src/semantics.rs
```diff
@@ -971,7 +971,7 @@ impl<'ast, T: Field> Checker<'ast, T> {
         &mut self,
         signature: UnresolvedSignature<'ast>,
         module_id: &ModuleId,
-        types: &TypeMap<'ast>,
+        state: &mut State<'ast, T>,
```
Schaeff: And this?

zokrates_core/src/semantics.rs
```diff
@@ -981,24 +981,39 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
-                        name: g.value,
-                        index,
-                    })));
-                }
-                false => {
-                    errors.push(ErrorInner {
-                        pos: Some(g.pos()),
-                        message: format!("Generic parameter {} is already declared", g.value),
-                    });
+            if let Some((key, _)) = state
```
Schaeff: Getting key_value for a given key and then ignoring the value, isn't that just checking `contains`?

zokrates_core/src/semantics.rs
```diff
@@ -1121,13 +1138,24 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
-                match generics_map.get(&name) {
-                    Some(index) => Ok(Constant::Generic(GenericIdentifier {name, index: *index})),
-                    None => Err(ErrorInner {
-                                    pos: Some(pos),
-                                    message: format!("Undeclared generic parameter in function definition: `{}` isn\'t declared as a generic constant", name)
-                                })
+                match (constants_map.get(name), generics_map.get(&name)) {
+                    (Some(c), None) => {
+                        match c {
+                            Type::Uint(bitwidth) => Ok(Constant::Identifier(name, bitwidth.to_usize())),
```
Schaeff: This should only accept constants of bitwidth 32 for now. There should be a test for this similar to the one with field.
```suggestion
                            Type::Uint(UBitwidth::B32) => Ok(Constant::Identifier(name, bitwidth.to_usize())),
```

zokrates_core/src/semantics.rs
```diff
@@ -1121,13 +1138,24 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+                            _ => Err(ErrorInner {
```
Schaeff, line 1145: This works but having `c` here makes it easier to see what `c` refers in this block

zokrates_core/src/semantics.rs
```diff
@@ -1144,7 +1172,7 @@ impl<'ast, T: Field> Checker<'ast, T> {
         &mut self,
         ty: UnresolvedTypeNode<'ast>,
         module_id: &ModuleId,
-        types: &TypeMap<'ast>,
+        state: &mut State<'ast, T>,
```
Schaeff: Does this need to be mutable?

zokrates_core/src/semantics.rs
```diff
@@ -1449,7 +1480,7 @@ impl<'ast, T: Field> Checker<'ast, T> {
             Statement::For(var, from, to, statements) => {
                 self.enter_scope();
 
-                let res = self.check_for_loop(var, (from, to), statements, pos, module_id, types);
+                let res = self.check_for_loop(var, (from, to), statements, pos, module_id, state);
```
Schaeff: Does this technically need the whole state and not just the types, or is it for simplicity?

zokrates_core/src/static_analysis/constant_inliner.rs
```diff
@@ -1,20 +1,26 @@
...
-impl<'ast, T: Field> ConstantInliner<'ast, T> {
-    pub fn new(modules: TypedModules<'ast, T>, location: OwnedTypedModuleId) -> Self {
-        ConstantInliner { modules, location }
-    }
-
+impl<'ast, 'a, T: Field> ConstantInliner<'ast, 'a, T> {
     pub fn inline(p: TypedProgram<'ast, T>) -> TypedProgram<'ast, T> {
-        let mut inliner = ConstantInliner::new(p.modules.clone(), p.main.clone());
+        let mut constants = HashMap::new();
+        let mut inliner = ConstantInliner {
```
Schaeff: using a constructor would be cleaner here

zokrates_core/src/typed_absy/mod.rs
```diff
@@ -290,8 +290,8 @@ impl<'ast, T: fmt::Display> fmt::Display for TypedFunction<'ast, T> {
 
 #[derive(Clone, PartialEq, Debug)]
 pub struct TypedConstant<'ast, T> {
-    ty: Type<'ast, T>,
-    expression: TypedExpression<'ast, T>,
+    pub ty: Type<'ast, T>,
```
Schaeff, line 293: Maybe I missed this before, but why do we need both a type and an expression here, where a type acan already be derived from the expression?

zokrates_core/src/typed_absy/result_folder.rs
```diff
@@ -42,13 +42,27 @@ pub trait ResultFolder<'ast, T: Field>: Sized {
...
         f: TypedFunction<'ast, T>,
     ) -> Result<TypedFunction<'ast, T>, Self::Error> {
         fold_function(self, f)
     }
 
+    fn fold_signature(
+        &mut self,
+        s: DeclarationSignature<'ast>,
+    ) -> Result<DeclarationSignature<'ast>, Self::Error> {
+        Ok(s)
```
Schaeff: Why is this not visited?

Schaeff (changes requested):
Pretty clean implementation, a few suggestions.

zokrates_core/src/semantics.rs
dark64: No, that should be a separate pull request I think

zokrates_core/src/semantics.rs
dark64: I've refactored this a bit now, this was not clean I agree

zokrates_core/src/semantics.rs
Schaeff: Yeah good, just making sure I'm following here!

zokrates_core/src/semantics.rs
dark64: It was for simplicity, but after further examination, I guess there is no need to pass a whole state where it's not actually needed, will revert some of this

zokrates_core/src/typed_absy/mod.rs
dark64: To avoid `Clone` trait bound, see discussion here https://github.com/Zokrates/ZoKrates/pull/792#discussion_r620493339

zokrates_core/src/semantics.rs
dark64: It was because of the `.or_default` on constants map entry, but I'm making some changes to remove the need for this mutability

zokrates_core/src/semantics.rs
dark64: It was because of the `.or_default` on constants map entry, but I'm making some changes to remove the need for this mutability

zokrates_core/src/semantics.rs
dark64: Same as other, will be removed

zokrates_core/src/semantics.rs
dark64: I was doing a bit of future-proofing here, but I agree this arm should check only for 32 now

zokrates_core/src/typed_absy/result_folder.rs
dark64: whoops :facepalm:

zokrates_core/src/typed_absy/mod.rs
Schaeff: Oh right, forgot about that

Schaeff (approved):

PR #865, opened 2021-05-14
Implement short-circuiting for assertions

Schaeff:
currently

```
if c then foo() else bar() fi
```

would always panic if `foo()` panics (even if `c` in false)

This is because we inline foo(), and conditionally return its result, but if a panic happened in the statements of foo, it would happen even if this branch isn't executed.

With this change, panics in each branch lead to the whole execution panicking only if the given branch is "executed", ie if `c`is true.

Both branches are still technically being executed (and paid for in terms of constraints), but the behavior is closer to what would expect.

Todo:
- [x] minimize constraints: next PR
- [x] try to reduce what we can before flattening to benefit from boolean propagation: next PR
- [x] tests nested conditionals
- [x] fix indentation on display implementation 
- [x] put behind flag?


zokrates_core/src/typed_absy/mod.rs
```diff
@@ -714,9 +715,26 @@ impl<'ast, T> MultiTyped<'ast, T> for TypedExpressionList<'ast, T> {
     }
 }
 
+#[derive(Clone, PartialEq, Debug, Hash, Eq)]
+// a block expression which returns an `E`
```
dark64: For the sake of consistency, I think this comment should go above the attribute

dark64:
Compile config should be updated in the `zokrates_js` type definition file as well: https://github.com/Zokrates/ZoKrates/blob/conditional-panics/zokrates_js/index.d.ts#L14

Schaeff:
> Compile config should be updated in the `zokrates_js` type definition file as well: https://github.com/Zokrates/ZoKrates/blob/conditional-panics/zokrates_js/index.d.ts#L14

changed

dark64 (approved):
LGTM!

PR #873, opened 2021-05-18
implement member expression and select expression

Schaeff:
Todo
- [x] remove remaining non generic impl flattening field member


dark64 (approved):
Looks clean! :100:

PR #881, opened 2021-05-20
Move rng example source code to zokrates_cli/examples (so they are compiled)

axic:


axic:
It seems the `reveal_bit` example doesn't compile anymore?

axic:
I get the following trying to compile it:
```
$ zokrates compile -i reveal_bit.zok 
Compiling reveal_bit.zok

rng_tutorial/reveal_bit.zok:14:13
	Variable in for loop cannot have type field

rng_tutorial/reveal_bit.zok:21:63
	Expected index to be of type u32, found bitNum
```

Was `field` as loop variable ever allowed?

axic:
Fixed this, however found other problems with the documentation itself. It lists `--light` as the option to be used, but for `compile` I get a deprecation warning and for `compute-witness` it fails straight. Removing it of course works, but the witness data is never displayed.

Additionally the example still uses tuples and not structs.

Schaeff:
Good catch! I just checked mdbook and we could also `include` the script parts from a file, and run the whole tutorial in CI. Opened #890 for this!

axic:
Sounds good! I will not have time this week to do anything here, but perhaps merging this is already a good step?

Schaeff (approved):

Issue #876, opened 2021-05-19
Refactor stdlib with latest features


dark64:
## Abstract

Some of the algorithms found in stdlib can be improved with the latest features we have recently introduced.

## Motivation

Any implementation that wrapped constants inside a function can now be written as:

```zokrates
def foo() -> field:
    return 4

// as

const field FOO = 4
```

## Specification

Find cases like these and refactored them using the latest language features.

## Backwards Compatibility
No  incompatibilities.

PR #893, opened 2021-05-25
Refactor stdlib

dark64:
https://github.com/Zokrates/ZoKrates/issues/876

Do not merge.

Note: `changelog-not-required` label will be removed if user-facing changes are made


zokrates_stdlib/stdlib/ecc/babyjubjubParams.zok
```diff
@@ -14,31 +14,17 @@ struct BabyJubJubParams {
...
+    JUBJUBA: 168700,  // Coefficient A
+    JUBJUBD: 168696,  // Coefficient D
+    MONTA: 168698,    // int(2*(JUBJUB_A+JUBJUB_D)/(JUBJUB_A-JUBJUB_D))
+    MONTB: 1,         // int(4/(JUBJUB_A-JUBJUB_D))
+    INFINITY: [0, 1], // Point at infinity
+    Gu: 16540640123574156134436876038791482806971768689494387082833631921987005038935, // Generator
+    Gv: 20819045374670962167435360035096875258406992893633759881276124905556507972311 // ...
 }
+
+def main() -> BabyJubJubParams:
```
Schaeff: We could omit that and just use the constant?

dark64: I had second thoughts about this, so I left it in at first so you could still use the `import` statement on this file. I have now removed this definition, a bit of a breaking change but should be fine as we will deprecate the `import` statement eventually.

PR #894, opened 2021-05-26
implement function call expression

Schaeff:
To review once #873 is in

### DIFF
--- a/zokrates_core/src/typed_absy/mod.rs
+++ b/zokrates_core/src/typed_absy/mod.rs
@@ -21,7 +21,7 @@ pub use self::parameter::{DeclarationParameter, GParameter};
 pub use self::types::{
     ConcreteFunctionKey, ConcreteSignature, ConcreteType, DeclarationFunctionKey,
     DeclarationSignature, DeclarationType, GArrayType, GStructType, GType, GenericIdentifier,
-    Signature, StructType, Type, UBitwidth, IntoTypes, Types
+    IntoTypes, Signature, StructType, Type, Types, UBitwidth,
 };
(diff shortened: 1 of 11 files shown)


zokrates_core/src/typed_absy/mod.rs
```diff
@@ -1243,6 +1263,12 @@ impl<'ast, T> From<TypedExpression<'ast, T>> for StructExpression<'ast, T> {
     }
 }
 
+impl<'ast, T> From<TypedExpression<'ast, T>> for TypedExpressionList<'ast, T> {
+    fn from(_: TypedExpression<'ast, T>) -> TypedExpressionList<'ast, T> {
+        unreachable!()
```
dark64, line 1271: This `From` seems a bit unclear at first glance, maybe you should leave a comment here on why is it needed and unreachable?

zokrates_core/src/typed_absy/mod.rs
```diff
@@ -1886,24 +1813,61 @@ impl<'ast, T: Clone> Member<'ast, T> for StructExpression<'ast, T> {
...
+
+impl<'ast, T: Field> Id<'ast, T> for StructExpression<'ast, T> {
+    fn identifier(id: Identifier<'ast>) -> Self::Inner {
+        StructExpressionInner::Identifier(id)
+    }
+}
+
+impl<'ast, T: Field> Id<'ast, T> for TypedExpressionList<'ast, T> {
+    fn identifier(_: Identifier<'ast>) -> Self::Inner {
+        unreachable!()
```
dark64, line 1858: Same here

dark64 (approved):
:100:

PR #896, opened 2021-05-26
Fix multi assignee parsing

### LINKED ISSUE #882: Parser accepts invalid code
## Description

The pest parser accepts code that contains array access in the declaration, which is invalid code.

## Environment

- Compiler version: 0.7.3
- Operating system: Ubuntu 20.04 LTS

## Steps to Reproduce

```zokrates
def main():
    field[2] a[2] = [1, 2]
    return
```

will result in

```
Compiling ./test.zok

assertion failed: `(left == right)`
  left: `1`,
  right: `0` (zokrates_core/src/absy/from_ast.rs:252:21)
```

### LINKED ISSUE #885: Tuple assignment returns wrong error message

Schaeff:
fixes #882 and fixes #885

### DIFF
--- a/zokrates_cli/examples/assign_to_element.zok
+++ b/zokrates_cli/examples/assign_to_element.zok
@@ -1,0 +1,11 @@
+// def foo() -> u32:
+//         return 0
+
+def bar() -> (u32, u32):
+        return 0, 0
+
+
+def main(u32[1] a, u32 b):
+        //a[0] = foo()
+        a[0], b = bar()
+        return
--- a/zokrates_cli/examples/assign_to_element.zok
+++ b/zokrates_cli/examples/assign_to_element.zok
@@ -1,11 +1,11 @@
-// def foo() -> u32:
-//         return 0
+def foo() -> u32:
+        return 0
 
 def bar() -> (u32, u32):
         return 0, 0
 
 
 def main(u32[1] a, u32 b):
-        //a[0] = foo()
+        a[0] = foo()
         a[0], b = bar()
         return
--- a/zokrates_cli/examples/assign_to_element.zok
+++ b/zokrates_cli/examples/assign_to_element.zok
@@ -1,11 +1,11 @@
 def foo() -> u32:
-        return 0
+    return 0
 
 def bar() -> (u32, u32):
-        return 0, 0
+    return 0, 0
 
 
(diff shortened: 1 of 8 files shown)


zokrates_core/src/compile.rs
```diff
@@ -237,6 +237,8 @@ fn check_with_arena<'ast, T: Field, E: Into<imports::Error>>(
     let typed_ast = Checker::check(compiled)
         .map_err(|errors| CompileErrors(errors.into_iter().map(CompileError::from).collect()))?;
 
+    println!("{}", typed_ast);
```
dark64: Remove this leftover debug code

zokrates_cli/examples/assign_to_element.zok
```diff
@@ -0,0 +1,11 @@
+def foo() -> u32:
+        return 0
```
dark64: Not really necessary, but we should stick to "tabs" being 4 spaces for the sake of cleaner formatting and consistency

dark64 (commented):
Some minor tweaks, but otherwise seems good to me!

zokrates_cli/examples/assign_to_element.zok
Schaeff: Can we enforce this somehow? I don't care what we use as long as it's consistent and CI enforced

Schaeff: Opened #901 for this and fixed this manually here

Schaeff:
@dark64 can you have another glance at this one?

dark64 (approved):
:100:

PR #897, opened 2021-05-26
Remove thread limit on cargo test, remove unnecessary -Zpackage-features

dark64:
- Removed `--test-threads=1` to hopefully speed up tests
- Removed opt-in `-Zpackage-features` as it's now considered stable


Schaeff (approved):
💯

PR #902, opened 2021-05-28
Add upper bound to for loops

### LINKED ISSUE #884: Huge loop sizes crash/bog down the compiler
The following piece of code will make the compiler use up a lot of memory (>20Gb) and the process gets killed:

```
def main():
    for u32 i in 0..4 do
        for u32 j in 0..(i-1) do
        endfor
    endfor
    return
```

If I replace the `(i-1)` part with a literal, it works and the above example creates 0 constraints as expected.

Schaeff:
set to 2**20

closes #884

### DIFF
--- a/changelogs/unreleased/902-schaeff
+++ b/changelogs/unreleased/902-schaeff
@@ -1,0 +1,1 @@
+Introduce a limit of 2**20 for for-loop sizes
--- a/zokrates_cli/examples/compile_errors/loop_too_large.zok
+++ b/zokrates_cli/examples/compile_errors/loop_too_large.zok
@@ -1,0 +1,4 @@
+def main():
+    for u32 i in 0..-1 do
+    endfor
+    return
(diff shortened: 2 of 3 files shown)


dark64:
I understand why this is an error now, but we should relax this to a warning at some point. Otherwise, fine with me for now!

dark64 (approved):

almatkhan:
Is anybody aware of workaround to turn this limitation off? 

I'm implementing the RSA Signature as a proof of concept with small key sizes (32-64 bit). It requires an exponent with modulo, and I'm performing this operation in the loop. However, even with 32-bit long integers, the private key exponent is far larger than the current limit.

```
// Helper function for modular exponentiation
// This function calculates (base^EXPONENT) % modulus
def mod_exp(u32 base, u32 exponent,  u32 modulus) -> (u32, bool) { 
// the bool is used to identify whether the error is due to overflow
// true means overflow
// we are not applying any sophisticated logic to prevent the overflow

    // first, we convert the input into u64 to prevent overflow
    u64 base64 = u32_to_u64(base);
    u64 modulus64 = u32_to_u64(modulus);

    u64 mut result = 1;
    for u32 i in 0..exponent {
        result = (result * base64) % modulus64;
    }

    (u32, u32) result = u64_to_u32(result);
    bool is_error = result.1 != 0; // true means overflow

    return (result.0, is_error);
}
```

Issue #892, opened 2021-05-24
Compilation hangs on zkoisc (worked on 0.7.1 + patches)


axic:
Compiling [zkoisc](https://github.com/axic/zkoisc/tree/b2b9bbc15e78ddaaf146362315e27e028c967ebf) seems to be hanging with 0.7.2 or 0.7.3. Ran it for half an hour max without any result (with or without `--isolate-branches`), and while the memory usage is slowly climbing, it does seem to plateau around 10 Gb and has fluctuations. This suggest to me there might be a non-recursive loop happening.

The code did not compile with 0.7.1, but did with d4494cb9c2fe84ec689d58ed204cc22ec0b6ac6c. The compilation with that commit takes around 15 minutes on my machine and results in around 1M constraints.

If I change the `STEP_LIMIT` to 1, 0.7.3 can compile it in 7 seconds, however using `--isolate-branches` it stalls.

axic:
#907 fixes compilation when `--isolate-branches` is not used.

if `--isolate-branches` is used, even with `STEP_LIMIT=5`, it runs for over 20 minutes and is terminated due to out of memory. I wasn't around to check what was the peak memory usage.

Schaeff:
@axic I compiled `zkoisc.zok` successfully on [this branch](https://github.com/Zokrates/ZoKrates/pull/950) with branch isolation on, `STEP_LIMIT=64`

axic:
The main change is improving the flattening to use less memory (by avoiding copies and using references)? Sounds like a plausible fix.

Schaeff:
I disproved this, I'm able to build in about the same amount of time on develop. Memory usage may be lower on this branch, but this issue is already fixed on develop. Feel free to close

PR #907, opened 2021-06-04
Avoid creating empty if-else statements, cache conditions when flattening

Schaeff:


axic:
This fixes #892 for me. It is still not possible to compile it with `--isolate-branches` though.

dark64 (approved):

PR #910, opened 2021-06-07
Visit declaration key in folder

### LINKED ISSUE #903: const-array-length parameters make a function impossible to call
## Description

Calling a function which uses a `const u32` identifier as the size of an array causes the inliner to panic.

## Environment

- Compiler version: `commit 4b2c5e7085cd1bdaa6a219c3ffe3cc148a096cc0`
- Operating system: Linux

## Steps to Reproduce

Try to compile:
```
const u32 N = 1
def foo(bool[N] arr) -> bool:
    return true

def main(bool[N] arr):
    assert(foo(arr))
    return
```
I see the following error:
```
called `Option::unwrap()` on a `None` value (.../ZoKrates/zokrates_core/src/static_analysis/reducer/inline.rs:70:10)
This is unexpected, please submit a full bug report at https://github.com/Zokrates/ZoKrates/issues
```

Schaeff:
This fixes #903 while fixing it for imported functions will be addressed in a different PR.

### DIFF
--- a/changelogs/unreleased/910-schaeff
+++ b/changelogs/unreleased/910-schaeff
@@ -1,0 +1,1 @@
+Fix access to constant in local function call
--- a/zokrates_cli/examples/constant_in_sig.zok
+++ b/zokrates_cli/examples/constant_in_sig.zok
@@ -1,0 +1,7 @@
+const u32 N = 1
+def foo(bool[N] arr) -> bool:
+    return true
+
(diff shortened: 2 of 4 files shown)


dark64 (approved):
LGTM, I guess this can be merged as the CI error is unrelated to this (https://github.com/Zokrates/ZoKrates/issues/911)

PR #913, opened 2021-06-08
Use global constant map for constant inlining

### LINKED ISSUE #915: Constant inlining is applied in the wrong context when importing functions
## Description

Imported functions still contain constants which are defined in the original file. Constant inlining then happens in the module they are imported into which leads to a crash or even an invalid result (if the importing file defines a constant with the same name)

## Environment

- Compiler version: develop

## Steps to Reproduce
```
// foo.zok
const u32 N = 42
def foo(field[N] a) -> bool:
   return true
```

```
from "./foo.zok" import foo
def main():
   return
```

Schaeff:
Fixes #915

### DIFF
--- a/zokrates_cli/examples/imports/import_with_alias.zok
+++ b/zokrates_cli/examples/imports/import_with_alias.zok
@@ -1,8 +1,8 @@
-//from "./bar" import main as bar
+from "./bar" import main as bar
 from "./baz" import BAZ as baz
 import "./foo" as f
 
 def main() -> field:
     field foo = f()
-    assert(foo == 21 + baz)
+    assert(foo == bar() + baz)
     return foo
--- a/zokrates_cli/examples/imports/import_with_alias.zok
+++ b/zokrates_cli/examples/imports/import_with_alias.zok
@@ -1,8 +1,8 @@
-from "./bar" import main as bar
+//from "./bar" import main as bar
 from "./baz" import BAZ as baz
 import "./foo" as f
 
 def main() -> field:
     field foo = f()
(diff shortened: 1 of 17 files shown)


zokrates_cli/examples/imports/import_with_alias.zok
```diff
@@ -1,8 +1,8 @@
-from "./bar" import main as bar
+//from "./bar" import main as bar
```
dark64: did this cause some issues?

zokrates_core/src/static_analysis/constant_inliner.rs
```diff
@@ -176,7 +162,7 @@ impl<'ast, T: Field> Folder<'ast, T> for ConstantInliner<'ast, T> {
     ) -> FieldElementExpression<'ast, T> {
         match e {
             FieldElementExpression::Identifier(ref id) => match self.get_constant(id) {
-                Some(c) => self.fold_expression(c).try_into().unwrap(),
+                Some(c) => c.try_into().unwrap(),
```
dark64: I suspect we can remove expression folding (`fold_expression`) on other variants like `fold_boolean_expression`, `fold_uint_expression`, etc?

dark64: like here https://github.com/Zokrates/ZoKrates/blob/759f4dd5534b779c0d73172a94a91ce63ccaac48/zokrates_core/src/static_analysis/constant_inliner.rs#L178

dark64 (approved):
LGTM! :rocket:

PR #914, opened 2021-06-08
Update book

dark64:
- Moved examples to a separate section
- Removed deprecated `--light` flag from tutorials
- Added a simple example of fs resolver to zokrates.js docs


Schaeff (approved):
🆗

PR #916, opened 2021-06-08
Introduce identifier expression

Schaeff:
The treatment of identifier expressions leads to duplicated code, for example in propagation. This increases the chances of mistakes as related code needs to be updated in 6 places (one per type).
Introduce an `IdentifierExpression` to avoid that.


dark64 (approved):
lgtm

PR #917, opened 2021-06-08
Add FIELD_SIZE_IN_BITS constant to EMBED

### LINKED ISSUE #900: Add FIELD_SIZE_IN_BITS constant to EMBED
## Abstract

Some of the stdlib is specific to altbn128, with hardcoded `254`. For example `unpack.zok`.
Now that we have constants, expose an embed constant whose value is the size of the field in bits.

## Motivation

We enable ourselves and users to write algorithms which work for different field sizes.

## Specification

Tentative:
```
from "EMBED" import FIELD_SIZE_IN_BITS
```

## Backwards Compatibility

Compatible

dark64:
Closes https://github.com/Zokrates/ZoKrates/issues/900

**Note:** Merge once constants are stable (after https://github.com/Zokrates/ZoKrates/pull/913 is in)


Schaeff (changes requested):
I think it's enough for this PR to reexport this embed inside the stdlib. I anticipate there will be an issue with the fact that no `main` function will be provided, we can put a dummy which returns this constant, and we should probably remove this requirement on having a main function at some point. We could return early just after semantic checks without an error. Will add an issue for this

Schaeff (approved):

Issue #883, opened 2021-05-20
Add constant generics to structs


dark64:
## Abstract
It would be useful to have constant generics added to structs.

## Motivation

As we have constants generics supported in function signatures, we could have the same for struct definitions:
```zokrates
struct Foo<N> {
    field[N] values
}

// something like this _should_ be possible
def foo<N>(Foo<N> f) -> field[N]:
    return f.values
```

## Specification

Any struct definition or reference to it should accept a constant generic declaration list eg. `<T, N>` after the identifier `Foo<T, N>`. The semantic checker should be updated appropriately to check for errors like missing generics, incorrect amount of supplied generics, etc.

## Backwards Compatibility
No incompatibilities.

PR #918, opened 2021-06-09
Recursive verification

### LINKED ISSUE #824: Implement recursive verifier
## Abstract

I'm opening this issue to keep track of the progress done on the recursive verification gadget.

## Motivation

This is a high priority feature.

## Specification

After some discussion with @Schaeff this could be done in two ways:

- By integrating with existing [ark](https://github.com/arkworks-rs) gadgets
- By implementing natively in the DSL (would be nice to have, but pairings could be tricky to implement)

More details will be added to the issue soon.

## Backwards Compatibility
No backwards incompatibilities.

### LINKED ISSUE #833: Add ark as an embed provider
## Abstract

Add ark as an embed provider to `zokrates_embed`. This requires implementing all necessary traits for the integration between constraint systems.

## Motivation

This will allow us to go further with https://github.com/Zokrates/ZoKrates/issues/824.

dark64:
Work in progress

Todo:
- [x] Refactor and cleanup
- [ ] ~~Wrap to structs and add to stdlib~~ (blocked by https://github.com/Zokrates/ZoKrates/issues/883) (will be done in another PR)

https://github.com/Zokrates/ZoKrates/issues/833 https://github.com/Zokrates/ZoKrates/issues/824


zokrates_core/src/embed.rs
```diff
@@ -281,6 +321,102 @@ pub fn sha256_round<T: Field>() -> FlatFunction<T> {
...
+        .clone()
+        .map(|i| FlatParameter::private(FlatVariable::new(i)));
+
+    let arguments = input_arguments
+        .chain(proof_arguments)
+        .chain(vk_arguments)
+        .collect();
+
+    let one_binding_statement = FlatStatement::Condition(
+        FlatExpression::Identifier(FlatVariable::new(0)),
```
Schaeff, line 358: Did you check that ark uses variable #0 as ~one?

Schaeff: It was the case for sha256 in bellman but worth double checking

zokrates_core/src/embed.rs
```diff
@@ -281,6 +321,102 @@ pub fn sha256_round<T: Field>() -> FlatFunction<T> {
...
+                FlatVariable::new(cs_index).into(),
+                FlatVariable::new(argument_index).into(),
+            )
+        })
+        .collect();
+
+    let constraint_statements: Vec<FlatStatement<T>> = constraints
+        .into_iter()
+        .map(|c| {
+            let c: Constraint<T> = from_ark::<T, Bls12_377>(c);
```
Schaeff: Since we're using a specific curve here, this function should check the id at the top

zokrates_core/src/embed.rs
dark64: should be `0` judging by this
https://github.com/arkworks-rs/snark/blob/master/relations/src/r1cs/mod.rs#L114

changelogs/unreleased/918-dark64
```diff
@@ -0,0 +1 @@
+Add verifier gadget to stdlib for recursive proof verification
```
Schaeff: I don't think we use the term gadget anywhere else. Maybe "Add SNARK verification..." ?

zokrates_core/src/imports.rs
```diff
@@ -120,6 +120,25 @@ impl Importer {
...
+                                Bw6_761Field::name()
+                            ))
+                            .with_pos(Some(pos)),
+                        )
+                        .in_file(location)
+                        .into());
+                    } else {
+                        SymbolDeclaration {
+                            id: symbol.get_alias(),
+                            symbol: Symbol::Flat(FlatEmbed::Verify),
```
Schaeff: Shall we rename `FlatEmbed::Verify` to something more precise? `SnarkVerifyBls12377` ?

zokrates_core/src/imports.rs
```diff
@@ -106,8 +106,8 @@ impl Importer {
                     if T::id() != Bn128Field::id() {
                         return Err(CompileErrorInner::ImportError(
                             Error::new(format!(
-                                "Embed sha256round cannot be used with curve {}",
-                                T::name()
+                                "Embed `sha256round` must be used with curve `{}`",
+                                Bn128Field::name()
```
Schaeff: Let's have both: expected x but found y

zokrates_core/src/proof_system/mod.rs
```diff
@@ -46,15 +46,15 @@ pub type Fq = String;
 pub type Fq2 = (String, String);
 
 #[derive(Serialize, Deserialize)]
-pub struct G1Affine(Fq, Fq);
+pub struct G1Affine(pub Fq, pub Fq);
```
Schaeff: Where is this needed?

dark64: Oh my bad, I forgot to revert that change, I needed it before for a reason I don't remember anymore

zokrates_core/src/imports.rs
dark64: I was thinking about this but I didn't like the naming :/ I guess we can't avoid that

zokrates_core_test/tests/tests/snark/snark_verify_bls12_377_2.json
```diff
@@ -0,0 +1,49 @@
+{
```
Schaeff, line 1: Would be nice to have a hint on how to generate these tests. Maybe in the zok file.

zokrates_embed/src/ark.rs
```diff
@@ -0,0 +1,344 @@
...
+    Vec<usize>,
+    Vec<usize>,
+    Vec<usize>,
+    Vec<Constraint<BW6Fr>>,
+    usize,
+) {
+    let cs_sys = ConstraintSystem::<BW6Fr>::new();
+    let cs = ConstraintSystemRef::new(cs_sys);
+
+    let mut rng = test_rng();
```
Schaeff: Does this make this function non-deterministic? If so, is it required? If not we should have a comment

zokrates_embed/src/ark.rs
```diff
@@ -0,0 +1,344 @@
...
+            })
+            .flatten(),
+    );
+
+    let out_index = match &res {
+        Boolean::Is(x) => x
+            .variable()
+            .get_index_unchecked(num_instance_variables)
+            .unwrap(),
+        Boolean::Not(x) => x
```
Schaeff: I find it surprising that we would return `x` for both `x` and `!x`? Are both of these match arms reachable? Or are we missing a test somewhere?

Schaeff (changes requested):
I left a few comments!

zokrates_embed/src/ark.rs
dark64: `test_rng` has a fixed seed and is required to generate a dummy proof we use to get the constraints, so it's deterministic

zokrates_embed/src/ark.rs
dark64: actually `Not` variant seems to be unreachable in this case, nice catch

zokrates_embed/src/ark.rs
Schaeff: Alright, can we add that as a comment? And/or change the variable name to `deterministic_rng`?

dark64: left a comment

zokrates_core_test/tests/tests/snark/snark_verify_bls12_377_2.json
dark64: Not sure what to write here, I've included some comments that might help (the underlying circuit, etc)

Schaeff: Is it trivial how to go from the vk generated by the setup to the inputs to this function? Same for the proof?

dark64: wrote some instructions, lmk what you think

Schaeff (approved):
LGTM!

PR #919, opened 2021-06-10
Output IR in smtlib2

leonardoalt:
Just checking whether this is wanted in the project, I'm using it locally anyway.

In case it is, this PR is still missing
- [x] Conjunction on the function level
- [x] Header including variable declaration
- [x] Tests


Schaeff:
- I wonder if this could be a separate command to which one can pass an already compiled program in the same spirit as `print-proof`
- can you run `rustup component add rustfmt` and `cargo fmt` before pushing? You can install a git hook that will make sure that happens [here](https://github.com/Zokrates/ZoKrates#git-hooks)
- As discussed offline, it would be nice if this formatting was wrapped with an `fmt::Display` implementation, see [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=f54c659a97b6d3b5532797a4193e2731) for an example. It's a bit more boilerplate, but this would prevent allocating a big string while keeping compatibility with the Rust formatting tools.
- What do you mean by "conjunction on the function level" ?

leonardoalt:
1) In that case, can the compiler re-generate the Expression tree? I mean, this could also be done by an entire separate script, the advantage of doing it inside the compiler is not having to parse.
2) Will do.
3) Will do.
4) From a logic point of view, the function is represented by the conjunction of all the printed equalities (statements). Currently this PR is only printing the equalities, but it should rather be
```
<header including variable declarations>

(assert
(and
statement1
...
statementN
)
)
```

Schaeff:
Yes this requires deserializing the program, as is already the case for `compute-witness` etc. I don't expect that to really be problematic performance-wise. I view exporting to smtlib2 as a step separate from compiling, hence the idea of a separate command.

leonardoalt:
That sounds fine to me, as long as I don't have to parse anything I'm happy :p

Schaeff:
Oh right, this logic is already in `zokrates_cli/src/ops/compute_witness` for example, you can use it as a starting point, keep the same `exec`, and adapt `cli_compute` to your case.

zokrates_cli/src/ops/generate_smtlib2.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    match ProgEnum::deserialize(&mut reader)? {
+        ProgEnum::Bn128Program(p) => cli_compute(p, sub_matches),
+        ProgEnum::Bls12_377Program(p) => cli_compute(p, sub_matches),
+        ProgEnum::Bls12_381Program(p) => cli_compute(p, sub_matches),
+        ProgEnum::Bw6_761Program(p) => cli_compute(p, sub_matches),
+    }
+}
+
+fn cli_compute<T: Field>(ir_prog: ir::Prog<T>, sub_matches: &ArgMatches) -> Result<(), String> {
```
Schaeff: For consistency with other commands, this should probably be `cli_smtlib2` or something

zokrates_cli/tests/integration.rs
```diff
@@ -314,4 +314,85 @@ mod integration {
...
+        assert_cli::Assert::command(&gen).succeeds().unwrap();
+
+        // load the expected smtlib2
+        let mut expected_smtlib2_file = File::open(&expected_smtlib2_path).unwrap();
+        let mut expected_smtlib2 = String::new();
+        expected_smtlib2_file
+            .read_to_string(&mut expected_smtlib2)
+            .unwrap();
+
+        // load the actual witness
```
Schaeff:
```suggestion
        // load the actual smtlib2
```

zokrates_cli/tests/integration.rs
```diff
@@ -314,4 +314,85 @@ mod integration {
...
+    #[test]
+    fn test_compile_and_smtlib2_dir() {
+        let dir = Path::new("./tests/code");
+        assert!(dir.is_dir());
+        for entry in fs::read_dir(dir).unwrap() {
+            let entry = entry.unwrap();
+            let path = entry.path();
+            if path.extension().unwrap() == "smt2" {
+                let program_name =
+                    Path::new(Path::new(path.file_stem().unwrap()).file_stem().unwrap());
```
Schaeff: Why is the file stem taken twice here?

Schaeff: If this is an adaptation from some existing code, it may have been to work with files of the form `file.expected.witness`?

zokrates_cli/tests/integration.rs
```diff
@@ -314,4 +314,85 @@ mod integration {
...
+                let prog = dir.join(program_name).with_extension("zok");
+                let smtlib2 = dir.join(program_name).with_extension("smt2");
+                test_compile_and_smtlib2(
+                    program_name.to_str().unwrap(),
+                    &prog,
+                    // TODO can we just pass `path` here instead of smtlib2?
```
Schaeff: I guess that works, yes... the spirit of filtering `path` by extension was to have each file stem a single time, not really to have the entire smt2 name available, hence its reconstruction here.

zokrates_core/src/ir/visitor.rs
```diff
@@ -0,0 +1,91 @@
...
+            f.visit_quadratic_combination(quad);
+            f.visit_linear_combination(lin);
+        }
+        Statement::Directive(dir) => f.visit_directive(dir),
+    }
+}
+
+pub fn visit_linear_combination<T: Field, F: Visitor<T>>(f: &mut F, e: &LinComb<T>) {
+    for expr in e.0.iter() {
+        f.visit_variable(&expr.0);
```
Schaeff, line 61: Would it make sense to also visit the field elements here? `visit_value`

Schaeff (changes requested):
Cool stuff and implementation! Just noticed a few tweaks, and we need to figure out where to insert a small entry about this in the docs.

zokrates_cli/tests/integration.rs
leonardoalt: Ah right, makes sense

leonardoalt:
Updated

zokrates_core/src/ir/smtlib2.rs
```diff
@@ -0,0 +1,141 @@
+use num_bigint::BigUint;
+use std::collections::BTreeSet;
+
+use super::*;
+use zokrates_field::Field;
+
+use super::expression::LinComb;
+use super::expression::QuadComb;
+use super::visitor::*;
+
```
leonardoalt, line 10: @Schaeff not sure if you have a convention for the order of these /\

Schaeff: No

leonardoalt:
> we need to figure out where to insert a small entry about this in the docs.

Maybe a new IR section under ZIR or so?

Schaeff (approved):
LGTM! As this is a very niche feature, we decided to limit ourselves to the CLI doc and not pollute the book with something most users would probably not use.

PR #924, opened 2021-06-22
Update for loop doc

### LINKED ISSUE #850: Assignment to for loop counter should be an error
I did the following code trying to mimic an actual loop, before I understood this really just gets "unrolled":
```
  for u32 step in 0..HARD_STEP_LIMIT do
    ...
    step = if (step == step_limit) then HARD_STEP_LIMIT else step fi
  endfor
```

I think this may be something people could try doing, so perhaps it would make sense making assignment to the loop counter a compile time error?

Schaeff:
Closes #850


dark64 (approved):

axic (approved):
As the person who opened #850, this looks like a good explainer 😉

PR #930, opened 2021-06-29
Introduce Select in zir to reduce cost of array access

### LINKED ISSUE #908: Reduce cost of variable index access
## Abstract

The variable index access is unnecessary expensive, and it grows with the size of the array. This gets worse with branch isolation.

## Motivation

```
def main(field[3] x, u32 i) -> field:
    return x[i]
```

This compiles down to 47 - 33 = 14 constraints excluding the range check for `i`:

We could instead do it in:
- 6 constraints to calculate `i == 0`, `i == 1` and `i == 2`
- `(i == 0) + (i == 1) + (i == 2) == 1` to check that the index is in range
- `(i == 0) * x_0 + (i == 1) * x_1 + (i == 2) * x_2` is the result

That's 7 constraints. More generally for an array of size N, this would cost 2N + 1 constraints

Our current approach is overly complicated, it basically rewrites to:

```
// note the usage of || here, which allows for multiple operands to be true. But we know that indices are distinct
assert(i == 0 || i == 1 || i ==2)
if i = 0 then a[0] else
    // this value is conditional to `i != 0`, but we already know that because i == 1 or i == 2 or the range check already failed
    if i = 1 then a[1] else
        // this value is conditional to `i != 1`, but we already know that because i == 2 or the range check already failed
        if i = 2 then a[2] else
            0 // if this is reached
        fi
    fi
fi
```
## Specification

The way we encode array access in the existing AST loses track of the fact that an index only has a single value at the time.
We need to somehow encode a n-ary conditional with distinct conditions.
The optimal way we described requires being able to "add" booleans, which is something we get access to in the flattening phase.

We could add `ZirExpression::Select(Vec<ZirExpression>, UExpression)` which in this simple case would look like
```
ZirExpression::Select([x_0, x_1, x_2], i)
```
And at the flattening phase apply the optimal flattening.

This would benefit from implementing #905 in zir as well to avoid code duplication.

## Backwards Compatibility

ok

Schaeff:
Closes #908


dark64 (approved):
:100:

Issue #938, opened 2021-07-12
Update code to use field size constants


Schaeff:
For example in the runtime errors, 2**252 is hardcoded where it could be curve dependent

PR #943, opened 2021-07-14
Use field constants to avoid curve-dependant code

dark64:
https://github.com/Zokrates/ZoKrates/issues/938

### DIFF
--- a/zokrates_core_test/tests/tests/split_bls.json
+++ b/zokrates_core_test/tests/tests/split_bls.json
@@ -1,5 +1,5 @@
 {
-	"entry_point": "./tests/tests/split_bls.zok",
+	"entry_point": "./tests/tests/split.zok",
 	"curves": ["Bls12_381"],
 	"tests": [
 		{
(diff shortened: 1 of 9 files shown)


zokrates_core_test/tests/tests/split_bls.json
```diff
@@ -1,5 +1,5 @@
 {
-	"entry_point": "./tests/tests/split_bls.zok",
+	"entry_point": "./tests/tests/split.zok",
 	"curves": ["Bls12_381"],
```
Schaeff, line 3: Can this now be run on more curves?

Schaeff (commented):
Looks nice!

zokrates_core_test/tests/tests/split_bls.json
dark64: There are two `json` test files (`split_bls.json` and `split_bn.json`) that use the same `zok` file (`split.zok`). If we were to specify both curves in the `curves` array we would have failing tests, because return values would not match. The way tests are structured now does not allow us to specify a curve for a specific test, only to specify many tests that would all run on each curve.

Schaeff: right!

Schaeff (approved):

Issue #890, opened 2021-05-24
Run tutorial in CI


Schaeff:
Make sure the rng tutorial always runs fine by running it in CI. Include the script from a file instead of writing it in code blocks.

dark64:
The tutorial still has outdated commands that need to be fixed as well (eg. `--light` flag), we can fix that now and do the CI stuff later.

PR #944, opened 2021-07-14
Test tutorials in CI

dark64:
https://github.com/Zokrates/ZoKrates/issues/890

Todo:
- [x] rng tutorial
- [x] sha256 tutorial


zokrates_cli/src/bin.rs
```diff
@@ -124,20 +124,14 @@ mod tests {
 
         builder
             .spawn(|| {
-                for p in glob("./examples/**/*").expect("Failed to read glob pattern") {
+                for p in glob("./examples/**/*.zok").expect("Failed to read glob pattern") {
```
Schaeff: I think there's a danger here: if one makes a typo in the extension of an example it would be skipped silently. It's imo better to crash which I think was the behaviour before this change: ignore folders, force files to have `zok` extension

zokrates_cli/src/bin.rs
```diff
@@ -196,7 +190,7 @@ mod tests {
     #[test]
     fn execute_examples_err() {
         //these examples should compile but not run
-        for p in glob("./examples/runtime_errors/*").expect("Failed to read glob pattern") {
+        for p in glob("./examples/runtime_errors/*.zok").expect("Failed to read glob pattern") {
```
Schaeff: same here

Schaeff (changes requested):
The tutorial part is fine, just some issues with the example tests. We need to make sure everything is being run.

zokrates_cli/src/bin.rs
dark64: I've introduced a `.sh` file (steps of the tutorial which are being tested in integration tests) in the tutorials which reside in the same folder as the `zok` files, so without filtering, it will fail here if there is a file with an extension other than `.zok` :/

eg. have a look here https://github.com/Zokrates/ZoKrates/tree/run-ci-tutorials/zokrates_cli/examples/book/rng_tutorial

Schaeff: Then let's explicitly ignore `.sh`, I guess glob allows that?

Schaeff (approved):

PR #951, opened 2021-07-30
Implement generic expressions in ZIR

Schaeff:
Introduce generic expressions for IfElse and Select.

To use the same implementation as for `typed_absy`, we introduce a `SelectOrExpression` enum as the output of the folder on a `Select` (same for `IfElse`). Unfortunately the `Expression` is never created as our current static analyser does not transform `IfElse` or `Select` into another expression, but it could happen, for example with a ZIR level propagation.


zokrates_core/src/zir/mod.rs
```diff
@@ -556,6 +570,65 @@ impl<'ast, T: fmt::Debug> fmt::Debug for ZirExpressionList<'ast, T> {
 
 // Common behaviour accross expressions
 
+pub trait Expr<'ast, T> {
```
Schaeff, line 574: @dark64 this may be useful to unify the treatment of expressions in zir propagation

PR #957, opened 2021-08-06
Apply propagation in ZIR

### LINKED ISSUE #952: Add propagation on ZIR
```
assert([x, 1] == [y, 1])
```

This cannot be simplified in typed_absy, but in ZIR it turns into

```
assert(x == y)
assert(1 == 1)
```

The second assert can be dropped.

Applying propagation to ZIR would take care of this simplification.

dark64:
Closes https://github.com/Zokrates/ZoKrates/issues/952

- [x] Fix failing tests
- [x] Write tests for the propagator
- [x] Refactor the propagator to use `result_folder`
- [x] Reduce embed call arguments


zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,516 @@
+use crate::zir::{
+    BooleanExpression, FieldElementExpression, Folder, UExpression, UExpressionInner, Variable,
+    ZirExpression, ZirFunction, ZirProgram, ZirStatement,
+};
+use std::collections::HashMap;
+use zokrates_field::Field;
+
```
Schaeff, line 11: Is there a reason we couldn't use the zir Folder here? Then we just need to implement folding on `ZirStatement::Definition` (adding to the constant map) and `ZirExpression` (reading from the constant map)

Schaeff: The rest would come from the default implementation of the folder

zokrates_core/src/static_analysis/uint_optimizer.rs
```diff
@@ -456,77 +456,70 @@ impl<'ast, T: Field> Folder<'ast, T> for UintOptimizer<'ast, T> {
...
+                vec![ZirStatement::MultipleDefinition(
                     lhs,
                     ZirExpressionList::EmbedCall(
                         embed,
                         generics,
                         arguments
                             .into_iter()
-                            .map(|e| self.fold_expression(e))
+                            .map(|e| match e {
+                                ZirExpression::Uint(e) => {
```
dark64: @Schaeff this needs a closer look

Schaeff: Let's always reduce

Schaeff: Then in `flatten_u_to_bits` we can assert that `from == to` in terms of bitwidths and just call `get_bits`

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,1414 @@
...
+type Constants<'ast, T> = HashMap<Identifier<'ast>, ZirExpression<'ast, T>>;
+
+#[derive(Default)]
+pub struct ZirPropagator<'ast, T> {
+    constants: Constants<'ast, T>,
+}
+
+impl<'ast, T: Field> ZirPropagator<'ast, T> {
+    pub fn new() -> Self {
```
Schaeff: We can omit that and use default where needed

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,1414 @@
...
+            FieldElementExpression::Select(e, box index) => {
+                let index = self.fold_uint_expression(index);
+                let e: Vec<FieldElementExpression<'ast, T>> = e
+                    .into_iter()
+                    .map(|e| self.fold_field_expression(e))
+                    .collect();
+
+                match index.into_inner() {
+                    UExpressionInner::Value(v) => {
+                        e.get(v as usize).cloned().expect("index out of bounds")
```
Schaeff: Is this panic reachable? I suspect it is, if the index is something like `if [a, 1] != [b, 0] then 1000000000 else 0`. That would suggest this propagation step has to be failible, which would require a `result_folder` for ZIR. Or we pretend we didn't see and return the original, which will fail at compile time, but that sounds less nice.

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,1414 @@
...
+            }
+            FieldElementExpression::Div(box e1, box e2) => {
+                match (
+                    self.fold_field_expression(e1),
+                    self.fold_field_expression(e2),
+                ) {
+                    (FieldElementExpression::Number(n1), FieldElementExpression::Number(n2)) => {
+                        FieldElementExpression::Number(n1 / n2)
+                    }
+                    (e1, e2) => FieldElementExpression::Div(box e1, box e2),
```
Schaeff: We can add the case `a / 1 = a`

Schaeff: And `a / 0 = compile error` if we use a `result_folder`

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,1414 @@
...
+
+                match condition {
+                    BooleanExpression::Value(true) => consequence,
+                    BooleanExpression::Value(false) => alternative,
+                    _ => FieldElementExpression::IfElse(
+                        box condition,
+                        box consequence,
+                        box alternative,
+                    ),
+                }
```
Schaeff, line 216: We can add `if _ then a else a fi = a`

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,1414 @@
...
+                        BooleanExpression::Value(false)
+                    }
+                    (e1, e2) => BooleanExpression::And(box e1, box e2),
+                }
+            }
+            BooleanExpression::Not(box e) => match self.fold_boolean_expression(e) {
+                BooleanExpression::Value(v) => BooleanExpression::Value(!v),
+                e => BooleanExpression::Not(box e),
+            },
+            BooleanExpression::IfElse(box condition, box consequence, box alternative) => {
```
Schaeff, line 422: Same

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,1414 @@
...
+                _ => UExpressionInner::Identifier(id),
+            },
+            UExpressionInner::Select(e, box index) => {
+                let index = self.fold_uint_expression(index);
+                let e: Vec<UExpression<'ast, T>> = e
+                    .into_iter()
+                    .map(|e| self.fold_uint_expression(e))
+                    .collect();
+
+                match index.into_inner() {
```
Schaeff, line 459: Same

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,1414 @@
...
+    ) -> BooleanExpression<'ast, T> {
+        match e {
+            BooleanExpression::Value(v) => BooleanExpression::Value(v),
+            BooleanExpression::Identifier(id) => match self.constants.get(&id) {
+                Some(ZirExpression::Boolean(BooleanExpression::Value(v))) => {
+                    BooleanExpression::Value(*v)
+                }
+                _ => BooleanExpression::Identifier(id),
+            },
+            BooleanExpression::Select(e, box index) => {
```
Schaeff, line 233: same

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,1414 @@
...
+            }
+            UExpressionInner::Div(box e1, box e2) => {
+                let e1 = self.fold_uint_expression(e1);
+                let e2 = self.fold_uint_expression(e2);
+
+                match (e1.into_inner(), e2.into_inner()) {
+                    (UExpressionInner::Value(n1), UExpressionInner::Value(n2)) => {
+                        UExpressionInner::Value((n1 / n2) % 2_u128.pow(bitwidth.to_usize() as u32))
+                    }
+                    (e1, e2) => {
```
Schaeff: `a / 1 = a`

Schaeff: We cannot add `0 / a = 0` because a could be 0, so we keep this as a runtime error.

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,1414 @@
...
+            }
+            UExpressionInner::Xor(box e1, box e2) => {
+                let e1 = self.fold_uint_expression(e1);
+                let e2 = self.fold_uint_expression(e2);
+
+                match (e1.into_inner(), e2.into_inner()) {
+                    (UExpressionInner::Value(n1), UExpressionInner::Value(n2)) => {
+                        UExpressionInner::Value(n1 ^ n2)
+                    }
+                    (e1, e2) => {
```
Schaeff: `a ^ a == true`

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,1414 @@
...
+            }
+            UExpressionInner::And(box e1, box e2) => {
+                let e1 = self.fold_uint_expression(e1);
+                let e2 = self.fold_uint_expression(e2);
+
+                match (e1.into_inner(), e2.into_inner()) {
+                    (UExpressionInner::Value(n1), UExpressionInner::Value(n2)) => {
+                        UExpressionInner::Value(n1 & n2)
+                    }
+                    (e1, e2) => {
```
Schaeff: `x & -1 == x`
`x & 0 == 0`

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,1414 @@
...
+            }
+            UExpressionInner::Or(box e1, box e2) => {
+                let e1 = self.fold_uint_expression(e1);
+                let e2 = self.fold_uint_expression(e2);
+
+                match (e1.into_inner(), e2.into_inner()) {
+                    (UExpressionInner::Value(n1), UExpressionInner::Value(n2)) => {
+                        UExpressionInner::Value(n1 | n2)
+                    }
+                    (e1, e2) => {
```
Schaeff:
```
x | -1 = -1
x | 0 = x
```

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,1414 @@
...
+                    }
+                }
+            }
+            UExpressionInner::LeftShift(box e, v) => {
+                let e = self.fold_uint_expression(e);
+                match e.into_inner() {
+                    UExpressionInner::Value(n) => {
+                        UExpressionInner::Value((n << v) & (2_u128.pow(bitwidth as u32) - 1))
+                    }
+                    e => UExpressionInner::LeftShift(box e.annotate(bitwidth), v),
```
Schaeff:
```
x << 0 = x
x << by = 0 if by >= bitwidth
```

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,1414 @@
...
+                }
+            }
+            UExpressionInner::RightShift(box e, v) => {
```
Schaeff: x >> 0 = x
x >> by = 0 if by >= bitwidth

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,1414 @@
...
+                    }
+                    e => UExpressionInner::Not(box e.annotate(bitwidth)),
+                }
+            }
+            UExpressionInner::IfElse(box condition, box consequence, box alternative) => {
+                let condition = self.fold_boolean_expression(condition);
+                let consequence = self.fold_uint_expression(consequence).into_inner();
+                let alternative = self.fold_uint_expression(alternative).into_inner();
+
+                match condition {
```
Schaeff: `(_ ? a : a) = a`

zokrates_core/src/flatten/mod.rs
```diff
@@ -1447,6 +1447,7 @@ impl<'ast, T: Field> Flattener<'ast, T> {
 
         // whether this value should be reduced, for example if it is then used in a bitwidth operation
         let should_reduce = metadata.should_reduce;
+        assert!(!should_reduce.is_unknown(), "should_reduce should be known");
```
Schaeff: can be removed

Schaeff (changes requested):
I found some other cases we can implement while we're looking at this!

zokrates_core/src/static_analysis/zir_propagation.rs
dark64: we could match for `0 / a = 0 if a != 0` no?

zokrates_core/src/static_analysis/zir_propagation.rs
dark64: Yeah, great catch. I think we should use a `result_folder` here

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -170,6 +168,10 @@ impl<'ast, T: Field> Folder<'ast, T> for ZirPropagator<'ast, T> {
                 let consequence = self.fold_field_expression(consequence);
                 let alternative = self.fold_field_expression(alternative);
 
+                if consequence.eq(&alternative) {
```
Schaeff: I think a more functional style would be better here, using a match statement on consequence and alternative also

```
match (condition, consequence, alternative) {
   (_, consequence, alternative) if consequence == alternative => consequence,
   (true, consequence, _) => consequence,
   (false, _, alternative) => alternative,
   ...
}
```

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -386,6 +388,10 @@ impl<'ast, T: Field> Folder<'ast, T> for ZirPropagator<'ast, T> {
                 let consequence = self.fold_boolean_expression(consequence);
                 let alternative = self.fold_boolean_expression(alternative);
 
+                if consequence.eq(&alternative) {
```
Schaeff: same

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -513,6 +521,12 @@ impl<'ast, T: Field> Folder<'ast, T> for ZirPropagator<'ast, T> {
                 let e2 = self.fold_uint_expression(e2);
 
                 match (e1.into_inner(), e2.into_inner()) {
+                    (e, UExpressionInner::Value(n))
```
Schaeff: the reflective case is also true

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -513,6 +521,12 @@ impl<'ast, T: Field> Folder<'ast, T> for ZirPropagator<'ast, T> {
...
+                        if n == 2_u128.pow(bitwidth.to_usize() as u32) - 1 =>
+                    {
+                        e
+                    }
+                    (_, UExpressionInner::Value(0)) => UExpressionInner::Value(0),
```
Schaeff: same, add symetric

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -526,6 +540,12 @@ impl<'ast, T: Field> Folder<'ast, T> for ZirPropagator<'ast, T> {
                 let e2 = self.fold_uint_expression(e2);
 
                 match (e1.into_inner(), e2.into_inner()) {
+                    (e, UExpressionInner::Value(0)) => e,
```
Schaeff: symetric

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -526,6 +540,12 @@ impl<'ast, T: Field> Folder<'ast, T> for ZirPropagator<'ast, T> {
...
+                    (_, UExpressionInner::Value(n))
```
Schaeff: symetric

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -564,6 +588,10 @@ impl<'ast, T: Field> Folder<'ast, T> for ZirPropagator<'ast, T> {
                 let consequence = self.fold_uint_expression(consequence).into_inner();
                 let alternative = self.fold_uint_expression(alternative).into_inner();
 
+                if consequence.eq(&alternative) {
```
Schaeff: same

zokrates_core/src/zir/result_folder.rs
```diff
@@ -0,0 +1,418 @@
+// Generic walk through a typed AST. Not mutating in place
```
Schaeff: typed -> zir

dark64: nice catch, was wrong in the `folder.rs` as well

zokrates_cli/examples/compile_errors/out_of_bounds.zok
```diff
@@ -0,0 +1,4 @@
+def main() -> field:
```
Schaeff: If you're trying to test the zir optimizer here, I suspect this gets optimized away earlier. One way to prevent that is using `[a, 1] == [b, 0]` with `a` and `b` some inputs, which is reduceable to `false` only in the zir propagator

zokrates_core/src/flatten/mod.rs
```diff
@@ -1225,19 +1194,9 @@ impl<'ast, T: Field> Flattener<'ast, T> {
 
                 // Handle complex parameters and assign values:
                 // Rename Parameters, assign them to values in call. Resolve complex expressions with definitions
-                // Clippy doesn't like the fact that we're collecting here, however not doing so leads to a borrow issue
```
dark64, line 1225: after some refactoring I did not encounter this issue explained in the comment? @Schaeff have a look just to be sure I'm not missing something here

zokrates_core/src/flatten/mod.rs
```diff
@@ -1082,38 +1080,24 @@ impl<'ast, T: Field> Flattener<'ast, T> {
         }
     }
 
-    fn flatten_u_to_bits(
+    fn u_to_bits(
```
dark64, line 1083: I left these functions in to avoid code duplication here https://github.com/Zokrates/ZoKrates/pull/957/commits/ed97e815c71f61e88cb484fb33fb2563cdf4b8a5#diff-c3ae48a36bfbdcad0b4a5d7861eea7bcbb116730f80c3832e77012b4988f248dR1131

zokrates_core/src/static_analysis/zir_propagation.rs
Schaeff: Well, unless `a` is a constant you cannot tell if `a == 0`. So yeah why not but this is quite niche.

zokrates_cli/examples/compile_errors/div_by_zero.zok
```diff
@@ -0,0 +1,3 @@
+def main(field input) -> field:
+    field divisor = if true then 0 else 1 fi
```
Schaeff: same here, this would be optimised away earlier

zokrates_core/src/flatten/mod.rs
Schaeff: I feel it was a clippy bug anyway!

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -0,0 +1,1694 @@
...
+            _ => fold_statement(self, s),
+        }
+    }
+
+    fn fold_field_expression(
+        &mut self,
+        e: FieldElementExpression<'ast, T>,
+    ) -> Result<FieldElementExpression<'ast, T>, Self::Error> {
+        match e {
+            FieldElementExpression::Number(n) => Ok(FieldElementExpression::Number(n)),
```
Schaeff, line 94: this looks like the folder's default, could be removed and covered by a single arm calling the default, see other optimization steps

Schaeff: ok it's short enough like this, closing

Schaeff (commented):
Almost there

zokrates_cli/examples/compile_errors/div_by_zero.zok
dark64: nice catch

Schaeff (approved):

PR #968, opened 2021-08-16
Use biguint in uint optimizer max calculations

### LINKED ISSUE #966: PosOverflow when creating an array and hashing it
## Description
When creating a long array and hashing it, the compiler crashes.

## Environment

- Compiler version: ZoKrates 0.7.5
- Operating system: Manjaro

## Steps to Reproduce
Paste this code into a file and compile (`zokrates compile -i filename.zok`):
```zokrates
import "hashes/sha256/sha256" as sha256dynamic

def main():
    u32[1][16] toHash = [[0; 16]]
    u32[8] gPrimary = sha256dynamic(toHash)
    return
```
It will work just fine.

However, make the array a bit longer and try again:
```zokrates
import "hashes/sha256/sha256" as sha256dynamic

def main():
    u32[50][16] toHash = [[0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16], [0; 16]]
    u32[8] gPrimary = sha256dynamic(toHash)
    return
```

This will crash
```
$ zokrates compile -i filename.zok
Compiling filename.zok

called `Result::unwrap()` on an `Err` value: ParseIntError { kind: PosOverflow } (/home/circleci/project/zokrates_core/src/static_analysis/uint_optimizer.rs:393:22)
This is unexpected, please submit a full bug report at https://github.com/Zokrates/ZoKrates/issues
```

dark64:
Closes https://github.com/Zokrates/ZoKrates/issues/966

### DIFF
--- a/zokrates_core/src/static_analysis/uint_optimizer.rs
+++ b/zokrates_core/src/static_analysis/uint_optimizer.rs
@@ -2,6 +2,7 @@ use crate::embed::FlatEmbed;
 use crate::zir::folder::*;
 use crate::zir::*;
 use std::collections::HashMap;
+use std::ops::{BitAnd, Shl, Shr};
 use zokrates_field::Field;
 
 #[derive(Default)]
@@ -366,16 +367,12 @@ impl<'ast, T: Field> Folder<'ast, T> for UintOptimizer<'ast, T> {
                 // reduce both terms
                 let e = self.fold_uint_expression(e);
 
-                let e_max: u128 = e
-                    .metadata
-                    .clone()
-                    .unwrap()
-                    .max
-                    .to_dec_string()
-                    .parse()
-                    .unwrap();
+                let e_max: num_bigint::BigUint = e.metadata.clone().unwrap().max.to_biguint();
+                let max = e_max
+                    .shl(by as usize)
+                    .bitand(&(2_u128.pow(range as u32) - 1).into());
 
-                let max = T::from((e_max << by) & (2_u128.pow(range as u32) - 1));
+                let max = T::try_from(max).unwrap();
 
                 UExpression::left_shift(force_reduce(e), by).with_max(max)
             }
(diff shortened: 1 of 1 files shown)


Schaeff:
Can we have a test for this? Maybe the one from the issue?

dark64:
> Can we have a test for this? Maybe the one from the issue?

This case is now covered in the unit optimizer module tests

zokrates_core/src/static_analysis/uint_optimizer.rs
```diff
@@ -366,35 +367,25 @@ impl<'ast, T: Field> Folder<'ast, T> for UintOptimizer<'ast, T> {
...
-                    .to_dec_string()
-                    .parse()
-                    .unwrap();
+                let e_max: num_bigint::BigUint = e.metadata.as_ref().unwrap().max.to_biguint();
+                let max = e_max
+                    .shl(by as usize)
+                    .bitand(&(2_u128.pow(range as u32) - 1).into());
 
-                let max = T::from((e_max << by) & (2_u128.pow(range as u32) - 1));
+                let max = T::try_from(max).unwrap();
```
Schaeff, line 375: Line 380 we reduce `e`, therefore I suspect `max` is `2**(bitwidth - by) - 1` no matter what `e_max` is before reducing? If this reasoning is correct, it would be clearer to reduce first in this match arm

Schaeff: We can also change to *not* always reducing `e`, in which case whether we have to reduce it or not depends on `e_max`: we have to reduce if `(e_max << by) > p - 1` which I think corresponds to line 375 panicking.

This would also mean removing the `bitand`

zokrates_core/src/static_analysis/uint_optimizer.rs
```diff
@@ -366,35 +367,25 @@ impl<'ast, T: Field> Folder<'ast, T> for UintOptimizer<'ast, T> {
...
 
-                let e_max: u128 = e
-                    .metadata
-                    .clone()
-                    .unwrap()
-                    .max
-                    .to_dec_string()
-                    .parse()
-                    .unwrap();
+                let e_max: num_bigint::BigUint = e.metadata.as_ref().unwrap().max.to_biguint();
```
Schaeff, line 383: Here we can choose to never reduce, because `(e >> by) < e` for all `by`. In other words, line 388 will never panic.

Schaeff: If we do this, we need to remove the `bitand`

Schaeff (approved):

PR #969, opened 2021-08-16
Avoid AST depth blowup on complex equality

### LINKED ISSUE #961: Large 2D array causes```fatal runtime error: stack overflow```
## Description

Unable to create moderately large 2D arrays. I need a 256x256 field array, however, I run into stack overflow for any size greater than 77x77. 

## Environment

- Compiler version: ZoKrates 0.7.5
- Operating system: macOS

## Steps to Reproduce

```zokrates
const u32 SIZE = 256
def main(private field[SIZE][SIZE] a):
    assert(a == [[0; SIZE]; SIZE])
    return
```
Here is the error message:
```bash
thread 'main' has overflowed its stack
fatal runtime error: stack overflow
[1]    25404 abort      zokrates compile -i
```

However, if I compile ZoKrates from source, I am able to work with ```field[103][103]``` arrays. 
How do I increase the stack size?

Schaeff:
closes #961

### DIFF
--- a/changelogs/unreleased/969-schaeff
+++ b/changelogs/unreleased/969-schaeff
@@ -1,0 +1,1 @@
+Fix stack overflow when testing equality on large arrays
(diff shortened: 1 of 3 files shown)


dark64 (approved):
LGTM

PR #975, opened 2021-08-20
Allow calls in constants

### LINKED ISSUE #972: Compilation failed while importing constants
## Description

Runs into ```Failed to unroll or inline program``` while importing certain constants.

## Environment

- Compiler version: 0.7.5 (b324e176)
- Operating system: macOS

## Steps to Reproduce

lib.zok

```zokrates
const u32 CONST_A = 10
const u32 CONST_B = CONST_A

def get_x_1D() -> field[CONST_B]:
    field[CONST_B] a = [20; CONST_B]
    return a
```

toy.zok

```zokrates
from "./lib.zok" import CONST_A, get_x_1D
const field[CONST_A] x_1D = get_x_1D()

def main(field[CONST_A] a_1D):
    assert(a_1D == x_1D)
    return
```

```bash
Compiling toy.zok

Compilation failed:

toy.zok:
	Failed to unroll or inline program. Check that main function arguments aren't used as array size or for-loop bounds
```
However, replacing `CONST_A` with `CONST_B` in `toy.zok` solves the issue though.

### LINKED ISSUE #985: Combatibility of functions using related generics lost when using structs
## Description
The combatibility of functions using related generics and their respective call sits is lost when replacing arrays with structs.

## Environment

- Compiler version: 0.7.6
- Operating system: Manjaro

## Steps to Reproduce
Paste the following code into a file:
```zokrates
def myFct<N, N2>(u64[N] ignored) -> u64[N2]:
	assert(2*N == N2)

	return [0; N2]


const u32 N = 3
const u32 N2 = 2*N
def main(u64[N] arg) -> bool:
    u64[N2] someVariable = myFct(arg)

    return true
```

This compiles without issues:
```
$ zokrates compile -i bug.zok 
Compiling bug.zok

Compiled code written to 'out'
Number of constraints: 196
```

Now introduce an arbitrary generic struct and replace `u64[N]` with that struct's name (`u64[N] → SomeStruct<N>`):

```zokrates
struct SomeStruct<N> {
    u64[N] f
}

def myFct<N, N2>(SomeStruct<N> ignored) -> u64[N2]:
	assert(2*N == N2)

	return [0; N2]


const u32 N = 3
const u32 N2 = 2*N
def main(SomeStruct<N> arg) -> bool:
    u64[N2] someVariable = myFct(arg)

    return true
```

This fails to compile:
```
$ zokrates compile -i bug.zok 
Compiling bug.zok

Compilation failed:

bug.zok:
        Call site `bug.zok/myFct<_, _>(SomeStruct<3>) -> u64[6]` incompatible with declaration `bug.zok/myFct<N, N2>(SomeStruct<N>) -> u64[N2]`
```

Note, however, that the following code does compile. It still uses a generic struct but does not use related generics.

```zokrates
struct SomeStruct<N> {
    u64[N] f
}

def myFct<N>(SomeStruct<N> ignored) -> u64[N]:
	return [0; N]


const u32 N = 3
def main(SomeStruct<N> arg) -> bool:
    u64[N] someVariable = myFct(arg)

    return true
```

```
$ zokrates compile -i bug.zok 
Compiling bug.zok

Compiled code written to 'out'
Number of constraints: 196
```

Schaeff:
Fixes #972 
Fixes #985 

Several refactorings happened as part of this PR:
- functions and constants are now kept in their original order in the `typed_ast`. This is because for example if we have `[c1, f1, c2]` where `f1` depends on `c1`, and `c2` depends on `f1` (by calling it), we need to process this exactly in this order, which is lost with the current way we register constants and functions in different vectors
- a new variant for `CoreIdentifier` was introduced for constants. This is because when converting a constant to an identifier, we only keep its name and forget about its module. This is problematic when importing functions, because their signature can contain constants which are local to the module the function is defined in, and lose their meaning in the destination module:

```
// dep.zok
const u32 A = 1
def foo(field[A])

// main.zok
const u32 A = 2
from "./dep.zok" import foo // with signature (field[A]) -> () here if we lost track of the module for A so we cannot find its value
```

After this change we instead have:

```
from "./dep.zok" import foo // with signature (field[dep.zok/A]) -> ()
```


zokrates_core/src/static_analysis/constant_inliner.rs
```diff
@@ -165,9 +172,8 @@ impl<'ast, T: Field> ResultFolder<'ast, T> for ConstantInliner<'ast, T> {
                         self.fold_function_symbol(fun)?,
                     ))
                 })
-                .collect::<Result<Vec<_>, _>>()
+                .collect::<Result<Vec<_>, _>>()?
```
Schaeff: This was actually pretty bad, forgot to call "?" and then called `into_iter` on an `Err` which yields an empty iterator, returning an empty array eventually..

zokrates_cli/examples/compile_errors/call_in_constant.zok
```diff
@@ -0,0 +1,9 @@
+// calling a function inside a constant definition is not possible yet
+
+def yes() -> bool:
+    return true
+
+const TRUE = yes()
```
dark64: you are missing a type here `const bool TRUE = ...`

zokrates_cli/examples/array_generic_inference.zok
```diff
@@ -0,0 +1,12 @@
+def myFct<N, N2>(u64[N] ignored) -> u64[N2]:
+	assert(2*N == N2)
+
```
dark64: The format in this example seems a bit messy, maybe removing empty lines and giving space between functions and constants would help (see example below)

zokrates_cli/examples/call_in_constant.zok
```diff
@@ -0,0 +1,9 @@
+// calling a function inside a constant definition is not possible yet
```
dark64: outdated comment?

zokrates_cli/examples/compile_errors/ambiguous_generic_call.zok
```diff
@@ -0,0 +1,14 @@
+// this should not compile, as A == B
+
+const u32 A = 2
```
dark64: judging by the comment above, shouldn't `A` and `B` have the same value here?

zokrates_cli/examples/struct_generic_inference.zok
```diff
@@ -0,0 +1,16 @@
+struct SomeStruct<N> {
+    u64[N] f
+}
+
+def myFct<N, N2, N3>(SomeStruct<N> ignored) -> u32[N2]:
+	assert(2*N == N2)
+
```
dark64: same thing as above, messy formatting

zokrates_core/src/compile.rs
```diff
@@ -249,6 +249,8 @@ fn check_with_arena<'ast, T: Field, E: Into<imports::Error>>(
     let typed_ast = Checker::check(compiled)
         .map_err(|errors| CompileErrors(errors.into_iter().map(CompileError::from).collect()))?;
 
+    log::trace!("\n{}", typed_ast);
```
dark64, line 252: probably a leftover that should be removed?

zokrates_core/src/static_analysis/constant_resolver.rs
```diff
@@ -0,0 +1,844 @@
...
+use crate::typed_absy::folder::*;
+use crate::typed_absy::*;
+use std::collections::HashMap;
+use zokrates_field::Field;
+
+// a map of the canonical constants in this program. with all imported constants reduced to their canonical value
+type ProgramConstants<'ast, T> =
+    HashMap<OwnedTypedModuleId, HashMap<ConstantIdentifier<'ast>, TypedConstant<'ast, T>>>;
+
+pub struct ConstantInliner<'ast, T> {
```
dark64: should we call this as `ConstantResolver`? seems like the file name is different now

zokrates_core/src/static_analysis/constant_resolver.rs
```diff
@@ -0,0 +1,844 @@
...
+        &mut self,
+        c: TypedConstantSymbolDeclaration<'ast, T>,
+    ) -> TypedConstantSymbolDeclaration<'ast, T> {
+        let id = self.fold_canonical_constant_identifier(c.id);
+
+        let constant = match c.symbol {
+            TypedConstantSymbol::There(imported_id) => {
+                // visit the imported symbol. This triggers visiting the corresponding module if needed
+                let imported_id = self.fold_canonical_constant_identifier(imported_id);
+                // after that, the constant must have been defined defined in the global map
```
dark64: `defined defined` -> `defined`

zokrates_core/src/static_analysis/reducer/constants_writer.rs
```diff
@@ -0,0 +1,163 @@
+// A folder to inline all constant definitions down to a single litteral and register them in the state for later use.
```
dark64: `litteral` -> `literal`

zokrates_core/src/static_analysis/reducer/constants_writer.rs
```diff
@@ -0,0 +1,163 @@
...
+        let id = self.fold_canonical_constant_identifier(d.id)?;
+
+        match d.symbol {
+            TypedConstantSymbol::Here(c) => {
+                let c = self.fold_constant(c)?;
+
+                use crate::typed_absy::{DeclarationSignature, TypedFunction, TypedStatement};
+
+                // wrap this expression in a function
+                let wrapper = TypedFunction {
```
dark64, line 109: Ah I see, this is the hacky part. I think at some point we should have some kind of expression evaluator so we don't have to wrap the expression in a function to reduce. Not ideal, but we can roll with this until then I guess.

zokrates_cli/examples/call_in_constant.zok
Schaeff: 😃

zokrates_cli/examples/compile_errors/ambiguous_generic_call.zok
Schaeff: yesss

zokrates_core/src/static_analysis/reducer/constants_writer.rs
Schaeff: will open an issue for this

Schaeff: #1021

zokrates_core/src/compile.rs
Schaeff: No, it's useful to have a trace of the AST post semantic checker, before we hit the static analyser.

dark64 (approved):

PR #977, opened 2021-08-23
Graceful handling of unconstrained variable detection

dark64:


dark64:
Should we include this in the changelog? @Schaeff

Schaeff:
Yes, this is user facing.

Schaeff (changes requested):
This would benefit from tests in this module.

Schaeff:
I missed the tests were actually there, my bad! Let's just make them stricter by not using `is_err` and `is_ok`

Schaeff (approved):

PR #982, opened 2021-08-25
Implement type aliasing

### LINKED ISSUE #860: Implement type aliasing
## Abstract

Not a priority, but it would be nice to support something like the following:

```zokrates
type byte = u8
type matrix<R, C> = field[R][C]

def foo<N>(matrix<N, N> m):
    ...
    return

def main(byte b): 
    ...
    return
```

## Specification

Implementation should include checking of semantics and proper grammar changes. Type aliases should be importable.

## Backwards Compatibility

No incompatibilities.

dark64:
Work in progress

Todo:
- [x] Cleanup
- [x] Check error messages
- [x] Add thorough tests
- [x] Update grammars (`type` keyword)


zokrates_core/src/semantics.rs
```diff
@@ -684,7 +794,7 @@ impl<'ast, T: Field> Checker<'ast, T> {
                                         }),
                                         ..t
                                     }),
-                                    _ => unreachable!()
+                                    _ => t // type alias
```
Schaeff: If I understand correctly, the `DeclarationType::Struct` case in this match can happen for a struct declaration, or for an alias declaration which points to a struct. In the latter case, I don't understand why the Struct type location should be modified?

Schaeff: Maybe this should also match `alias_generics` as if I understand correctly, this is what determines whether we have a struct declaration or an alias declaration. A similar match happens further down over both.

zokrates_core/src/semantics.rs
```diff
@@ -1263,7 +1372,58 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+                                                .and_then(|e| {
+                                                    UExpression::try_from_typed(e, &UBitwidth::B32)
+                                                        .map(|e| (g, e))
+                                                        .map_err(|e| ErrorInner {
+                                                            pos: Some(pos),
+                                                            message: format!("Expected u32 expression, but got expression of type {}", e.get_type()),
+                                                        })
+                                                })
+                                        },
+                                        None => Err(ErrorInner {
```
Schaeff: I don't think this should be allowed at all in alias definitions: all generics must always be passed. Maybe the error message is incorrect, or this is actually unreachable depending on what happens in the parser (maybe we create Some earlier and unwrap here, never creating None so this would be unreachable)

Schaeff: Oh this is for Type not DeclarationType so `_` is allowed.

dark64:
@Schaeff I cannot approve because I've opened the PR, but looks good to me! :+1:

Schaeff (approved):

PR #988, opened 2021-09-01
Introduce compilation mode

Schaeff:
```
zokrates check -i foo.zok --mode lib
```

Checks the file up to semantics, not requiring an entry point.


zokrates_core/src/compile.rs
```diff
@@ -186,37 +212,42 @@ pub fn compile<T: Field, E: Into<imports::Error>>(
...
-        .map_err(|e| CompileErrorInner::from(e).in_file(location.as_path()))?;
-
-    Ok(CompilationArtifacts {
-        prog: optimized_ir_prog,
-        abi,
-    })
+    let artifacts = check_with_arena(source, location.clone(), resolver, config, &arena)?;
+
+    match artifacts {
+        CompilationArtifacts::Lib => unreachable!(),
```
dark64, line 221: This might be reachable when called from `zokrates-js` because the config is passed by the caller:

https://github.com/Zokrates/ZoKrates/pull/988/files#diff-31d8ede13068a8e0e79ec05ff89244788c8903841d377659d712b5d2057e9ea6R109

If my understanding is correct, one could set the mode to lib and call compile which would throw here with no clear message.

zokrates_core/src/compile.rs
```diff
@@ -162,9 +168,25 @@ impl fmt::Display for CompileErrorInner {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone, Copy, PartialEq)]
+pub enum CompileMode {
+    Bin,
```
dark64, line 173: I think we should use something like this here

```rust
#[derive(Debug, Serialize, Deserialize, Clone, Copy, PartialEq)]
pub enum CompileMode {
    #[serde(rename = "bin")]
    Bin,
    #[serde(rename = "lib")]
    Lib
}
```

so it's a bit cleaner in js: `mode: "lib"` instead of `mode: "Lib"`. It's also possible to put `#[serde(rename_all = "lowercase")]` before the enum definition. 

We should also update typescript type information file:
https://github.com/Zokrates/ZoKrates/blob/add-compiler-mode-flag/zokrates_js/index.d.ts#L12

```ts
export type CompileMode = "bin" | "lib";

export interface CompileConfig {
   allow_unconstrained_variables?: boolean,
   isolate_branches?: boolean,
   mode?: CompileMode
}
```

Issue #991, opened 2021-09-06
Assertion failure for cast from field to uint for values exceeding the uint's range


m1cm1c:
## Description
When casting a field to a uint type, the compiler crashes if the field's value exceeds the uint's range at compile time.

## Environment

- Compiler version: branch `develop` (commit `b18ea015`)
- Operating system: Manjaro

## Steps to Reproduce
Compile the following code:

```zokrates
import "utils/casts/field_to_u64"

def main():
	u32 i = 30
	field x = 30
	field q = x**i
	u64 u = field_to_u64(q)
	return
```

```
zz compile -i bug4.zok
Compiling bug4.zok

assertion failed: `(left == right)`
  left: `205891132094648999999999981553255926290448385`,
 right: `0` (/home/christoph/.applications/ZoKrates/zokrates_core/src/static_analysis/propagation.rs:450:49)
This is unexpected, please submit a full bug report at https://github.com/Zokrates/ZoKrates/issues
```
(`zz` being aliased to the Zokrates binary)

PR #997, opened 2021-09-12
Make field to uint casts truncate values bigger than uint max

### LINKED ISSUE #1026: utils casting field to u64 results in error that should not occur according to compiler
## Description

Computing witness of a `.zok` file containing a type cast from field element to `u64` on output from a `sha256Packed` call results in an unexpected assertion error.

## Environment

- Compiler version: Zokrates `0.7.6`
- Operating system: macOS 11.3.1 Intel Chip

## Steps to Reproduce

```zokrates
// test.zok
import "hashes/sha256/512bitPacked" as sha256packed
import "utils/casts/field_to_u64" as fu64

def main(private field a, private field b, private field c, private field d) -> u64:
    field[2] h = sha256packed([a, b, c, d])
    return fu64(h[0])
```

```bash
$ zokrates compile -i test.zok
$ zokrates compute-witness -a 0 0 0 5

Computing witness...
Execution failed: Sum check failed
The default ZoKrates interpreter should not yield this error. Please open an issue: expected 263561599766550617289250058199814760685 to equal 18446744073709551615
```

dark64:
https://github.com/Zokrates/ZoKrates/issues/991

The following issue occurs if the value is too large to unpack. The minimal example is as follows (will panic):

```zok
from "EMBED" import unpack

def main():
    bool[2] bits = unpack(4) // we need 3 bits to unpack 4, we can detect this in compile time and error out
    return
```
If the value is an input, the following code will compile but the sum check will fail on execution:
```zok
from "EMBED" import unpack

def main(field i):
    bool[2] bits = unpack(i) // Execution failed: Sum check failed
    return
```

Also `field_to_ux` conversions can fail because one can pass a field bigger than the max uint value. This PR makes use of the safe unpacking logic we have which allows us to cast and truncate fields that are larger than the max uint value.


zokrates_stdlib/stdlib/utils/casts/field_to_u64.zok
```diff
@@ -1,5 +1,9 @@
 from "EMBED" import unpack, u64_from_bits
 
-def main(field i) -> u64:
-    bool[64] bits = unpack(i)
+const field U64_MAX = 18446744073709551615
+
+// Condition: field value is asserted to be in range [0, U64_MAX]
+def main(field input) -> u64:
+    assert(input <= U64_MAX)
```
m1cm1c: The behavior would be the same as in most other programing languages if instead of asserting `input < 2^64`, the return value were set to `input mod 2^64`.

Schaeff: I can see that argument. Can you elaborate on your main use case for the truncating behavior? Is it when packing a bunch of arguments into one field outside the circuit, then unpacking them inside knowing they are all in range? Or do you actually have cases where truncation happens?

m1cm1c: Yes, I have a use case where truncation happens. When I implemented long integer multiplication, it would've been useful to represent a long integer as an array of u64s (I switched to u32s by now because of the problems). However, I of course cannot multiply two u64s and then access the result as multiplication would then happen mod 2^64. I have to use field elements that each represent a 64 bit block of a long integer. If I multiply them, the entire result is there. However, I need to get back to u64, or rather two of them: one for the result of the current position, one for the carry. So the idea was to just stuff the field element into a u64, call that the current element, and subtract that from the multiplication result, divide by 2^64, convert to u64, and call that the carry. However, the compiler crashed as soon as the multiplication result was so big that the carry was actually needed.

zokrates_stdlib/stdlib/utils/casts/field_to_u64.zok
```diff
@@ -1,5 +1,7 @@
...
+from "field" import FIELD_SIZE_IN_BITS
+import "utils/pack/bool/unpack"
+import "utils/casts/u64_from_bits"
 
-def main(field i) -> u64:
-    bool[64] bits = unpack(i)
-    return u64_from_bits(bits)
\ No newline at end of file
+def main(field input) -> u64:
+    bool[FIELD_SIZE_IN_BITS] bits = unpack(input)
```
Schaeff, line 6: @m1cm1c this is what we ended up doing: decompose the field to N bits with N the size of the field (254 for bn128) in a safe way (intuition there is to force 0 to be decomposed as [false; 254] and not as the decomposition of the field size P as 254 bits, which requires an extra check, as in the field, 0 and P are the same). Then truncate this decomposition of N bits to 64/32/16/8 bits. Please share if you have any thoughts!

m1cm1c: This sounds like a good solution.

Schaeff (approved):

PR #998, opened 2021-09-12
Remove cast to usize causing wrong values in wasm environment

dark64:
`u64` type seems to be broken in wasm environments: `x as usize` rounds any uint number bigger than `u32` down to `u32` because in web assembly  `usize` is equal to `u32`. This means the `u64` type is unusable in `zokrates-js <= 1.0.35` :eyes:

### DIFF
--- a/changelogs/unreleased/998-dark64
+++ b/changelogs/unreleased/998-dark64
@@ -1,0 +1,1 @@
+Fix invalid cast to `usize` which caused wrong values in 32-bit environments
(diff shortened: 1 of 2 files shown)


Schaeff:
Interesting, is this the only occurence of this cast in the codebase?

dark64:
> Interesting, is this the only occurence of this cast in the codebase?

It seems like all other occurrences are fine

Schaeff (approved):

PR #1008, opened 2021-09-15
Reduce the size of the verifier for schemes G16 and PGHR13

m1cm1c:
Reduces the size of the verifier for schemes G16 and PGHR13 by 400 LOC (e.g. from 595 LOC to 195 LOC for a small proof).

PR #1009, opened 2021-09-15
Reduce the size of the verifier for scheme PGHR13.

m1cm1c:
Reduces the size of the verifier for scheme PGHR13 by 400 LOC.

Blocked by #1008.


m1cm1c:
The tests are expected to fail as #1008 is not merged into `develop` yet.

m1cm1c:
Included in  #1008.

PR #1010, opened 2021-09-15
Ternary operator

### LINKED ISSUE #999: Introduce ternary operator
## Abstract

Add the ternary `condition ? consequence : alternative`.

## Motivation

The if/else construct we have now is being used with statements. Statements themselves are not an issue (they can be emulated with a function call, and isolated with branch isolation), however side effects to existing variables is an issue. With the ternary operator `? :` it is clearer that expressions are expected.

## Specification

same as say javascript

## Backwards Compatibility

Compatible, we can have both at the same time

dark64:
Closes https://github.com/Zokrates/ZoKrates/issues/999


zokrates_core/src/semantics.rs
```diff
@@ -2228,26 +2228,35 @@ impl<'ast, T: Field> Checker<'ast, T> {
                         message: format!("{{consequence}} and {{alternative}} in conditional expression should have the same type, found {}, {}", e1.get_type(), e2.get_type()),
                     })?;
 
+                let kind = match kind {
+                    crate::absy::ConditionalKind::IfElse => {
```
Schaeff, line 2232: This can be imported to avoid the full path

Schaeff (approved):

PR #1012, opened 2021-09-20
Add optional message to assert statement

### LINKED ISSUE #990: Assertion error messages
## Abstract
Assertion error messages would make debugging lengthy code much easier.

## Motivation
When using assertions, Zokrates currently only is capable of providing error messages like the following when an assertion fails:

> Execution failed: User assertion failed: expected 1 to equal 0

Not even a line number is provided. This makes it very difficult to find the failing assertion in a proof scheme that uses lots of assertions.

## Specification
`assert()` should accept a second argument: A revert string. This string should be printed (possibly together with the line number) if the assertion fails.

## Backwards Compatibility
Backwards compatibility can be provided by making the second argument optional.

dark64:
Closes https://github.com/Zokrates/ZoKrates/issues/990


zokrates_core/src/ir/interpreter.rs
```diff
@@ -300,8 +298,28 @@ impl fmt::Display for Error {
...
+                ref error,
+            } => {
+                write!(
+                    f,
+                    "{}",
+                    error
+                        .as_ref()
+                        .map(|m| m.to_string())
+                        .unwrap_or_else(|| format!(
+                            "Unsatisfied constraint: expected {} to equal {}",
```
dark64: Still not sure if we should remove this or not, theoretically constraints that don't have a matching runtime error shouldn't fail at all (definitions, returns, etc) but still it makes sense to handle this in case we introduce something similar that could fail?

Schaeff: I can't see a case where something can fail but there is no error. But since we represent everything as `Constraint` we cannot really encode that in the Rust type system. Maybe the least bad option is to replace `unwrap_or_else` by `expect("Found an unsatisfied constraint without an attached error.")` or something

zokrates_core/src/static_analysis/variable_write_remover.rs
```diff
@@ -49,6 +49,7 @@ impl<'ast> VariableWriteRemover {
                         Access::Select(head) => {
                             statements.insert(TypedStatement::Assertion(
                                 BooleanExpression::UintLt(box head.clone(), box size.into()),
+                                None,
```
Schaeff: Why can this never fail? Seems like it's a runtime range check during an array access, like:

```
def main(u32 i, field[3] a):
    a[i] = 42
    return
```

this currently fails at runtime with `Execution failed: User assertion failed` which is incorrect

Schaeff: Is it because this metadata is for a user message, but this is an internal assertion added by the compiler, and this is the only one we insert into typed_absy? Are we incorrectly assuming that all assertions in typed_absy come from the user? If that's the case, I guess the metadata could be wrapped in an enum to capture metadata for these internal errors. Or we need to take a step back and see if there is a cleaner solution as we're starting to have a lot of options in enums in options haha

Schaeff: Note: this does not happen for variable reads, because they get carried away to zir as is.

zokrates_core/src/flatten/mod.rs
```diff
@@ -147,6 +146,15 @@ impl<T: Field> FlatUExpression<T> {
     }
 }
 
+impl From<AssertionType> for RuntimeError {
```
Schaeff: Would it make sense to rename `AssertionType` to `RuntimeError` and have that name throughout the pipeline? That would avoid this non-trivial conversion as well as using the variable `ty` which is used in the type system a lot. We could instead use `error` or something?

zokrates_core/src/flatten/mod.rs
```diff
@@ -2830,7 +2834,7 @@ mod tests {
                             .metadata(metadata.clone()),
                         box UExpressionInner::Value(42).annotate(32).metadata(metadata),
                     ),
-                    None,
+                    AssertionType::Source("".to_string()),
```
Schaeff: We could create one instance at the beginning of the tests and clone it each time, or have a mock function behind a  cfg flag

zokrates_core/src/static_analysis/variable_write_remover.rs
```diff
@@ -49,7 +49,7 @@ impl<'ast> VariableWriteRemover {
                         Access::Select(head) => {
                             statements.insert(TypedStatement::Assertion(
                                 BooleanExpression::UintLt(box head.clone(), box size.into()),
-                                None,
+                                AssertionType::SelectRangeCheck,
```
Schaeff: 💯

zokrates_core/src/flatten/mod.rs
Schaeff: Or a function on [Typed|Zir]Statement to create an assertion with this value, `TypedStatement::assertion` and `TypedStatement::assertion_with_error`

Schaeff (commented):
Looking good but I have a few suggestions

zokrates_core/src/flatten/mod.rs
dark64: went with a mock function on `RuntimeError`

Schaeff (approved):
💯

PR #1023, opened 2021-09-29
Add serde `default` attribute to compile config to accept partial objects

dark64:
This does not work currently:
```js
zokrates.compile(source, {
    config: {
      allow_unconstrained_variables: true,
    },
  });
```
The whole config object must be passed:
```js
zokrates.compile(source, {
    config: {
      allow_unconstrained_variables: true,
      isolate_branches: false,
    },
  });
```

This PR adds `#[serde(default)]` to compile config fields so partial configs are accepted.

### DIFF
--- a/changelogs/unreleased/1023-dark64
+++ b/changelogs/unreleased/1023-dark64
@@ -1,1 +1,0 @@
-Add serde `default` attribute to compile config to accept partial objects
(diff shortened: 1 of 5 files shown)


Schaeff:
Nice, note that this is also included in #988

changelogs/unreleased/1023-dark64
```diff
@@ -0,0 +1 @@
+Add serde `default` attribute to compile config to accept partial objects
```
Schaeff: I think this change is only internal so a changelog doesn't seem required?

Schaeff (approved):

PR #1030, opened 2021-10-05
Allow general postfix expressions

Schaeff:
Postfix expressions are now limited to `identifier ~ access+`.
Allow any expression to be the base of a postfix expression.

Also fixes a bug in the propagation which as far as I can tell was not reachable before this change.


dark64 (dismissed):
💯

PR #1032, opened 2021-10-11
Fail at compile time on unsatisfied equality of complex values

### LINKED ISSUE #1031: Asserting equality of constant expressions that result in a false value should fail at compile time
## Description

Asserting equality of constant expressions that result in a false value should fail at compile time. This seems to not be the case with complex types (arrays/structs).

## Environment

- Compiler version: 0.7.7
- Operating system: Ubuntu 20.04

## Steps to Reproduce

with arrays 
```zokrates
def main():
    assert([0u64] == [1u64])
    return
```

with structs

```zokrates
struct Foo {
    field a
}

def main():
    assert(Foo { a: 0 } == Foo { a: 1 })
    return
```

asserting basic types fails as expected

```zokrates
def main():
    assert(0f == 1f) // Assertion failed on expression `0f == 1f`
    return
```

Schaeff:
Closes #1031 but since this happens in zir we can't point at which assertion failed. I assume this will be easier once #1012 is in as we would still have the source location and optional message. Note that now in the `typed_absy` propagation, we show which expression ended up being false, but this is not always the original expression, it is the expression after the previous pass (which can expose some compiler internal variables unfortunately):

```
def one() -> field:
    return 1

def main():
    assert(one() + 1 == 1)
    return
```

Will fail with 

```
Assertion failed on expression `(#CALL_RETURN_AT_INDEX_0 + 1f) == 1f`
```

All in all, #1012 will be nice as it has a reference to the original source, not to the previous pass, for both typed_absy and zir propagation.


dark64 (approved):
Merged develop and added some tests, should be good now

PR #1040, opened 2021-10-22
Update lockfiles

petehunt:
`zokrates-js` does not currently work in a multithreaded (webworker) environment due to a bug in `getrandom`: https://github.com/rust-random/getrandom/issues/214.

Deleting the lockfiles and reinstalling fixes the issue.

Hopefully this is accepted, because it's super important to run `zokrates-js` off of the main thread to ensure a smooth user experience.


dark64:
I think this is an issue of defining `self` which `getrandom` uses to determine the environment (node or browser), we actually run `zokrates-js` successfully in a web worker in the remix plugin. We will gladly accept this if this resolves your issue. Thanks!

dark64 (approved):

Issue #1039, opened 2021-10-21
Powers of Tau MPC ceremony for Zokrates


dwebchapey:
## Abstract

It will be good feature addition for Zokrates to integrate some sort of MPC ceremony that generates entropy for setup phase . Currently randomness for Zokrates circuit setup is currently taken from `rand`  which is fine for a centralised solution since the entity running the setup is the only one that needs to be convinced. 

## Motivation

Integrating Powers of Tau to setup phase of Zokrates.

## Suggestion

[Perpetual Powers of Tau](https://medium.com/coinmonks/announcing-the-perpetual-powers-of-tau-ceremony-to-benefit-all-zk-snark-projects-c3da86af8377) ceremony can be used for phase 1.  For second phase , which is circuit specific an analogous approach using powers of tau can be used.

Support required for GM17 , G16 schemes.

PR #1044, opened 2021-11-02
Adding support for MPC

dark64:
https://github.com/Zokrates/ZoKrates/issues/1039


zokrates_book/src/SUMMARY.md
```diff
@@ -26,7 +26,8 @@
 	- [zokrates.js](toolbox/zokrates_js.md)
 
 - [Examples](examples/index.md)
-	- [A SNARK Powered RNG](examples/rng_tutorial.md)
-	- [Proving knowledge of a hash preimage](examples/sha256example.md)
+    - [A SNARK Powered RNG](examples/rng_tutorial.md)
+    - [Proving knowledge of a hash preimage](examples/sha256example.md)
+    - [Performing a Trusted Setup](examples/mpc_tutorial.md)
```
Schaeff: I think it makes more sense to have this in the toolbox section of the documentation

zokrates_cli/examples/book/rng_tutorial/test.sh
```diff
@@ -4,7 +4,7 @@ set -e
 bin=$1; stdlib=$2
 
 function zokrates() {
-  ZOKRATES_STDLIB=$stdlib $bin $*
+  ZOKRATES_STDLIB=$stdlib $bin "$@"
```
Schaeff, line 7: just curious: what does this do?

zokrates_cli/src/ops/mpc/beacon.rs
```diff
@@ -0,0 +1,152 @@
...
+use clap::{App, Arg, ArgMatches, SubCommand};
+use phase2::parameters::MPCParameters;
+use std::fs::File;
+use std::io::{BufReader, BufWriter};
+use std::path::Path;
+
+pub fn subcommand() -> App<'static, 'static> {
+    SubCommand::with_name("beacon")
+        .about("Applies a random beacon")
+        .display_order(3)
```
Schaeff: What does this do? If it's necessary, we should have a comment explaining what it does

zokrates_cli/src/ops/mpc/beacon.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    let reader = BufReader::new(file);
+    let mut params = MPCParameters::read(reader, true)
+        .map_err(|why| format!("Could not read `{}`: {}", path.display(), why))?;
+
+    let beacon_hash = sub_matches.value_of("hash").unwrap();
+    let num_iterations: usize = sub_matches.value_of("iterations").unwrap().parse().unwrap();
+
+    if !(10..=63).contains(&num_iterations) {
+        return Err("Number of hash iterations should be in [10, 63] range".to_string());
```
Schaeff:
```suggestion
        return Err("Number of hash iterations should be in the [10, 63] range".to_string());
```

zokrates_cli/src/ops/mpc/beacon.rs
```diff
@@ -0,0 +1,152 @@
...
+    // Create an RNG based on the outcome of the random beacon
+    let mut rng = {
+        use byteorder::{BigEndian, ReadBytesExt};
+        use crypto::digest::Digest;
+        use crypto::sha2::Sha256;
+        use rand::chacha::ChaChaRng;
+        use rand::SeedableRng;
+
+        // The hash used for the beacon
+        let hash_result = hex::decode(beacon_hash);
```
Schaeff: this would be more idiomatic with

```
let hash_result = hex::decode(beacon_hash)
   .map_err(|_| "Beacon hash should be in hexadecimal format".to_string())?;
```

zokrates_cli/src/ops/mpc/beacon.rs
```diff
@@ -0,0 +1,152 @@
...
+        if hash_result.is_err() {
+            return Err("Beacon hash should be in hexadecimal format".to_string());
+        }
+
+        let mut cur_hash = hash_result.unwrap();
```
Schaeff: Cf comment above, this can be reduced to one line I think?

zokrates_cli/src/ops/mpc/beacon.rs
```diff
@@ -0,0 +1,152 @@
...
+    let num_iterations: usize = sub_matches.value_of("iterations").unwrap().parse().unwrap();
+
+    if !(10..=63).contains(&num_iterations) {
+        return Err("Number of hash iterations should be in [10, 63] range".to_string());
+    }
+
+    println!("Creating a beacon RNG");
+
+    // Create an RNG based on the outcome of the random beacon
+    let mut rng = {
```
Schaeff, line 88: Is this step implemented in any of the underlying libraries? Any chance we can reuse that?

zokrates_cli/src/ops/mpc/beacon.rs
```diff
@@ -0,0 +1,152 @@
...
+        }
+        println!();
+    }
+
+    let output_path = Path::new(sub_matches.value_of("output").unwrap());
+    let output_file = File::create(&output_path)
+        .map_err(|why| format!("Could not create `{}`: {}", output_path.display(), why))?;
+
+    println!(
+        "\nYour contribution has been written to `{}`",
```
Schaeff, line 162: Seems like this command is doing both contributing and applying the beacon. Is there a way to decorrelate the two? Or do I understand correctly that this is basically a normal contribution but using a specific deterministic randomness instead of some private randomness, so it almost sounds like a special case of contribution? Could this be exposed as a flag on the contribution command instead of this separate command? Not necessarily suggesting to do that, but conceptually?

zokrates_cli/src/ops/mpc/contribute.rs
```diff
@@ -0,0 +1,115 @@
...
+        use rand::chacha::ChaChaRng;
+        use rand::{OsRng, Rng, SeedableRng};
+
+        let h = {
+            let mut system_rng = OsRng::new().unwrap();
+            let mut h = Blake2b::default();
+
+            // Gather 1024 bytes of entropy from the system
+            for _ in 0..1024 {
+                let r: u8 = system_rng.gen();
```
Schaeff, line 88: The tutorial mentions different types of randomness that can be used here, but we always use the Os?

zokrates_cli/src/ops/mpc/export.rs
```diff
@@ -0,0 +1,98 @@
...
+        .map_err(|why| format!("Could not read `{}`: {}", path.display(), why))?;
+
+    println!("Exporting keys from `{}`...", path.display());
+
+    let params = mpc_params.get_params();
+
+    let mut pk: Vec<u8> = Vec::new();
+    params.write(&mut pk).unwrap();
+
+    let vk = VerificationKey {
```
Schaeff: We usually have this parsing code in zokrates_core so that we can access it from js or cli

Schaeff: Now I realise it's the case for most of the additions in this PR, do you see anything preventing us from having a `zokrates_mpc` crate and only do I/O in the CLI? It could be a separate PR.

zokrates_cli/src/ops/mpc/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+pub mod contribute;
+pub mod export;
+pub mod init;
+pub mod verify;
+
+pub fn subcommand() -> App<'static, 'static> {
+    SubCommand::with_name("mpc")
+        .about("Multi-party computation (MPC) protocol")
+        .subcommands(vec![
+            init::subcommand(),
```
Schaeff: It seems clearer to add the ordering number here.

zokrates_cli/src/ops/mpc/verify.rs
```diff
@@ -0,0 +1,104 @@
...
+    // read compiled program
+    let path = Path::new(sub_matches.value_of("circuit").unwrap());
+    let file =
+        File::open(&path).map_err(|why| format!("Could not open `{}`: {}", path.display(), why))?;
+
+    let mut reader = BufReader::new(file);
+
+    match ProgEnum::deserialize(&mut reader)? {
+        ProgEnum::Bn128Program(p) => cli_mpc_verify(p, sub_matches),
+        _ => unimplemented!(),
```
Schaeff: This should be unreachable as we should check this is allowed before calling this, cf `helpers.rs`

zokrates_cli/tests/integration.rs
```diff
@@ -408,9 +408,11 @@ mod integration {
         let tmp_dir = TempDir::new(".tmp").unwrap();
         let tmp_base = tmp_dir.path();
 
-        for p in glob("./examples/book/rng_tutorial/*").expect("Failed to read glob pattern") {
-            let path = p.unwrap();
-            std::fs::hard_link(path.clone(), tmp_base.join(path.file_name().unwrap())).unwrap();
+        for p in glob("./examples/book/rng_tutorial/*")
+            .expect("Failed to read glob pattern")
+            .filter_map(Result::ok)
```
Schaeff: This seems less strict as we filter out errors, rather than failing on errors

zokrates_cli/tests/integration.rs
```diff
@@ -432,9 +434,42 @@ mod integration {
         let tmp_dir = TempDir::new(".tmp").unwrap();
         let tmp_base = tmp_dir.path();
 
-        for p in glob("./examples/book/sha256_tutorial/*").expect("Failed to read glob pattern") {
-            let path = p.unwrap();
-            std::fs::hard_link(path.clone(), tmp_base.join(path.file_name().unwrap())).unwrap();
+        for p in glob("./examples/book/sha256_tutorial/*")
+            .expect("Failed to read glob pattern")
+            .filter_map(Result::ok)
```
Schaeff: Same

Schaeff (changes requested):
Cool addition!

zokrates_cli/src/ops/mpc/beacon.rs
dark64: From what I understand applying the beacon is actually the same as contributing, the only difference is how the rng is created. I've separated it because that is the flow in other available tools, so we don't differentiate that much.

zokrates_cli/src/ops/mpc/contribute.rs
dark64: RNG is based on a mixture of system randomness and user-provided randomness

zokrates_cli/src/ops/mpc/export.rs
dark64: yeah I was thinking about doing this at first, but wanted to have this done asap

zokrates_cli/src/ops/mpc/verify.rs
dark64: yeah nice catch, I forgot to fix this :)

zokrates_cli/src/ops/mpc/beacon.rs
dark64: unfortunately I did not find any

zokrates_cli/src/ops/mpc/beacon.rs
dark64: :+1:

zokrates_cli/src/ops/mpc/beacon.rs
dark64: It makes a strict order of commands in the help section when you run `zokrates mpc --help`. I made it so it is displayed in the order it should be run

zokrates_cli/examples/book/rng_tutorial/test.sh
dark64: This was interesting, to give context this function is used as a wrapper to have cleaner commands in the book (we reference the lines in this script). I had an issue where if you would pass a string for one of the commands eg. `-e "1 2"` it would pass it down as `-e 1 2` which is wrong because now `2` is outside the string and is passed as a separate argument which would result in errors like `invalid parameter "2"`. By using `"$@"` these groups are preserved because the expansion logic is a bit different

zokrates_cli/src/ops/mpc/contribute.rs
dark64: https://github.com/Zokrates/ZoKrates/blob/mpc/zokrates_cli/src/ops/mpc/contribute.rs#L71 user-provided randomness is applied here

zokrates_cli/src/ops/mpc/beacon.rs
Schaeff: Right, see comment lower down about moving this to the mpc level

zokrates_cli/src/ops/mpc/beacon.rs
Schaeff: Did you use something as a reference? This is cryptographic code, we should be careful.

zokrates_cli/src/ops/mpc/contribute.rs
Schaeff: I see! For cryptographic code like this I feel it would be good to have a separate crate and reuse external tools as much as possible, and if not, give links to references.

zokrates_core/src/proof_system/bellman/groth16.rs
```diff
@@ -96,6 +99,63 @@ impl<T: Field + BellmanFieldExtensions> NonUniversalBackend<T, G16> for Bellman
...
+        let hash = params.contribute(rng);
+        params.write(output).map_err(|e| e.to_string())?;
+
+        Ok(hash)
+    }
+
+    fn verify<R: Read>(
+        params: &mut R,
+        program: Prog<T>,
+        phase1_radix: &mut R,
```
Schaeff, line 133: I think here you are constraining both readers to be the same type. In theory they could be different I guess? So two generic parameters would make more sense?

dark64: nice catch

Schaeff (approved):

Issue #1047, opened 2021-11-07
Declaring and assigning const array with const size


sshravan:
## Description

Const value not realized during assignment. 

## Environment

- Compiler version: 0.7.7 (fc55d881)
- Operating system: Ubuntu

## Steps to Reproduce

bug/utils/matrix.zok

```zokrates
const u32 SIZE = 4
const field[SIZE][SIZE] MATRIX = [
	[
	       1,       2,       3,       4
	],
	[
	       5,       6,       7,       8
	],
	[
	       8,       7,       6,       5
	],
	[
	       4,       3,       2,       1
	]
]
```

bug/main.zok

```zokrates
from "./utils/matrix" import MATRIX
const u32 SIZE = 4

def main(field scalar):
    field[SIZE][SIZE] result = [[0; SIZE]; SIZE]
    for u32 i in 0..SIZE do
        for u32 j in 0..SIZE do
            result[i][j] = MATRIX[i][j] * scalar
        endfor
    endfor

    assert(result == [[0; SIZE]; SIZE])
    return
```

```bash
$ zokrates compile -i bug/main.zok      
Compiling bug/main.zok

Compilation failed:

bug/main.zok:
	Expression of type `field[4][4]` cannot be assigned to constant `bug/main.zok/MATRIX` of type `field[bug/./utils/matrix.zok/SIZE][bug/./utils/matrix.zok/SIZE]`
```
However, replacing `const field[SIZE][SIZE] MATRIX` with `const field[4][4] MATRIX` in `bug/utils/matrix.zok` solves the issue though.

dark64:
Hey @sshravan, thanks for the report. This is definitely a bug.

devnote: I've reduced the issue to:

```zok
// utils/mod2.zok
// surprisingly, it works when moving this module one folder up (./mod2) 
const field A = 1
const field B = A + 1

// mod1.zok
from "./utils/mod2" import B

def main():
    return
```

PR #1050, opened 2021-11-15
Implement missing cases in constant resolution

### LINKED ISSUE #1038: Compilation failed while importing constants
## Description

Runs into ```Failed to reduce constant `SIZE_WORDS` in module `toy.zok` to a literal, try simplifying its declaration``` while importing certain constants.

## Environment

- Compiler version: 0.7.7 (fc55d881)
- Operating system: Ubuntu 20.04

## Steps to Reproduce

lib.zok

```zokrates
const u32 SIZE_BYTES = 136
const u32 SIZE_WORDS = SIZE_BYTES/8
```

toy.zok

```zokrates
from "./lib" import SIZE_WORDS

def main(field[SIZE_WORDS] a):
    assert(a == [0; SIZE_WORDS])
    return
```

```bash
Compiling toy.zok

Compilation failed:

toy.zok:
	Failed to reduce constant `SIZE_WORDS` in module `toy.zok` to a literal, try simplifying its declaration

```
However, replacing `SIZE_WORDS` with `17` in `lib.zok` solves the issue though.

Schaeff:
closes #1038 and #1047

### DIFF
--- a/changelogs/unreleased/1050-schaeff
+++ b/changelogs/unreleased/1050-schaeff
@@ -1,0 +1,1 @@
+Fix reduction of constants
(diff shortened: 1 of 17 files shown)


dark64 (commented):
Looks good, should we add a test that covers this case? Just to be sure, happy to merge without if you think it's unnecessary.

Schaeff:
Half of these tests fail on develop, which is the expected result, as file names matter.

dark64 (approved):

Issue #1056, opened 2021-11-24
Internal error: entered unreachable code with `--isolate-branches`


axic:
It happens on this branch and I don't have the time to reduce to a smaller example: https://github.com/axic/zkoisc/tree/field. It doesn't happen on the `master` branch of the repo.

The binary used is the release version of 0.7.8 for aarch64-darwin.

```sh
% zokrates compile --isolate-branches -i oisc.zok                   
Compiling oisc.zok

internal error: entered unreachable code (zokrates_core/src/flatten/mod.rs:1041:30)
This is unexpected, please submit a full bug report at https://github.com/Zokrates/ZoKrates/issues
```

(This may be related to #891)

dark64:
Hello @axic, thanks for the report! I've successfully reduced the code and identified the bug.

dark64:
Closing this, this issue is no longer present on the develop branch.

PR #1060, opened 2021-12-01
Add missing case in zir propagation

dark64:
Addressing https://github.com/Zokrates/ZoKrates/issues/1056

**Todo**: Underlying issue is connected to the hacky `condition_cache` we use to avoid cloning conditions. It might be possible to avoid this by assigning a condition to a variable before zir, which would make this cache unnecessary. Current implementation partially solves the issue, not sure if there is a case where it might fail again. This part of the issue is solved in https://github.com/Zokrates/ZoKrates/pull/1064

### DIFF
--- a/zokrates_core/src/flatten/mod.rs
+++ b/zokrates_core/src/flatten/mod.rs
@@ -459,9 +459,7 @@ impl<'ast, T: Field> Flattener<'ast, T> {
         let condition_id = self.use_sym();
         statements_flattened.push(FlatStatement::Definition(condition_id, condition_flat));
 
-        if !condition.is_constant() {
-            self.condition_cache.insert(condition, condition_id);
-        }
+        self.condition_cache.insert(condition, condition_id);
 
         let (consequence, alternative) = if self.config.isolate_branches {
             let mut consequence_statements = vec![];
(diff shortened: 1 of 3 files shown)


zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -76,6 +76,41 @@ impl<'ast, T: Field> ResultFolder<'ast, T> for ZirPropagator<'ast, T> {
                     }
                 }
             }
+            ZirStatement::IfElse(e, consequence, alternative) => {
+                match self.fold_boolean_expression(e)? {
+                    BooleanExpression::Value(true) => Ok(consequence
```
dark64, line 81: Note for review: If we determine that condition is a constant, we can propagate only the branch that is chosen to save time. This is the reason why this code does not propagate both branches before checking the condition and then picking a branch.

zokrates_core/src/flatten/mod.rs
```diff
@@ -754,9 +756,11 @@ impl<'ast, T: Field> Flattener<'ast, T> {
         statements_flattened: &mut FlatStatements<T>,
         expression: BooleanExpression<'ast, T>,
     ) -> FlatExpression<T> {
-        // check the cache
-        if let Some(c) = self.condition_cache.get(&expression) {
-            return (*c).into();
+        if !expression.is_constant() {
```
Schaeff: I feel this is not needed if we made sure we did not insert any constants into the map? The map access below would already return None

Schaeff:
Since #1064, this can be scaled down to the changes on the propagator

Schaeff (approved):

PR #1063, opened 2021-12-07
Update trusted_setup.md

Schaeff:


dark64 (approved):
Nice, reads a lot better now!

PR #1064, opened 2021-12-08
Remove condition cache, add condition redefiner

Schaeff:


zokrates_core/src/static_analysis/condition_redefiner.rs
```diff
@@ -0,0 +1,395 @@
...
+        //     if e && f then 2 else 3
+        // } else {
+        //     field b = 2
+        //     if e && f then 2 else 3
+        // }
+        //
+        // should become
+        //
+        // bool #CONDITION_0 = c && d
+        // field foo = if c && d then {
```
dark64: `field foo = if #CONDITION_0 then {`

dark64 (approved):
Nice one!

zokrates_core/src/static_analysis/condition_redefiner.rs
dark64: updated

PR #1065, opened 2021-12-08
Remove regex parsing from bellman bindings

dark64:
Just a minor refactor to remove this ugly regex parsing of bellman g1/g2 affine


Schaeff (approved):
Wow sweet

Schaeff:
Wow sweet

PR #1067, opened 2021-12-10
Update rust toolchain version

dark64:


zokrates_pest_ast/src/lib.rs
```diff
@@ -705,7 +705,7 @@ mod ast {
     #[derive(Debug, FromPest, PartialEq, Clone)]
     #[pest_ast(rule(Rule::assignee_access))]
     pub enum AssigneeAccess<'ast> {
-        Select(ArrayAccess<'ast>),
+        Select(Box<ArrayAccess<'ast>>),
```
Schaeff: This looks inconsistent, it should be implemented inside ArrayAccess, maybe also on MemberAccess, or allowed here and delegated to another issue

Schaeff (approved):

Issue #1049, opened 2021-11-08
Optimized Range Check Gadget


sshravan:
## Abstract

It would be great if ZoKrates can support an optimized range check gadget. Currently, ZoKrates uses the range check algorithm from the Sapling ([Pg 141 Section A.3.2.2](https://github.com/zcash/zips/blob/master/protocol/sapling.pdf)). To check if  ```x < constant```, it ZoKrates takes around 525+ constraints. But if I am not wrong, given the bit representation of x as a witness (bits of x doesn't have to be boolean constrained according to the sapling spec), it would be possible to range check with at most ```2*log2(constant)``` constraints.

## Motivation

Performing ```x < constant``` is costly especially for smaller constants. However, given the bit representation of `x` this can be made substantially cheaper using the range check from sapling. In certain scenarios, it may not be unreasonable to expect that bits of `x` is available. 

## Specification
Let `c` be a compile time const and `n` be the bitlength of `c`.

`range_check(const field c, field x, field[n] x_bits) -> bool`

Schaeff:
I looked more closely at this and will write down what I found:
- we use an adapted version of the Sapling algorithm, the main reason being that in our case, `x < c` is not necessarily being asserted, it can also be used as a boolean value in the program. This means that we are more limited in what we can do, because we cannot just throw (have an unsatisfied constraint) if `x < c` is false.
- there are some optimisations that apply to the general case, for example checking the bits cache for `x` instead of reducing it
- other optimisations require targeting specifically cases where `x < c` is in an assertion, and applying exactly the Sapling algorithm.

PR #1080, opened 2022-01-07
Optimize range check

dark64:
https://github.com/Zokrates/ZoKrates/issues/1049


zokrates_core/src/flatten/mod.rs
```diff
@@ -604,6 +602,49 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+    /// * `e` - the `FlatExpression` that's being checked against the range
+    /// * `c` - the constant upper bound of the range
+    fn enforce_constant_lt_check(
+        &mut self,
+        statements_flattened: &mut FlatStatements<T>,
+        e: FlatExpression<T>,
+        c: T,
+        error: RuntimeError,
+    ) {
+        if c == T::zero() {
```
Schaeff: This works but is a bit odd: `assert(e < c)` should always fail if `c == 0`. Here we encode that by adding `c == 1` to the constraints, which works as `0 != 1`, but why do we use `c == 1` and not any other unsatisfied constraint like `0 == 1` directly?

zokrates_core/src/flatten/mod.rs
```diff
@@ -1068,7 +1078,11 @@ impl<'ast, T: Field> Flattener<'ast, T> {
                 // define variables for the variable bits
                 let variables: Vec<_> = expressions
                     .into_iter()
-                    .map(|e| self.define(e.get_field_unchecked(), statements_flattened))
+                    .map(|e| {
+                        FlatExpression::Identifier(
+                            self.define(e.get_field_unchecked(), statements_flattened),
```
Schaeff: I think we can call `.into()` here to convert to identifier instead of wrapping

zokrates_core/src/flatten/mod.rs
```diff
@@ -1815,8 +1834,16 @@ impl<'ast, T: Field> Flattener<'ast, T> {
 
         let res = match should_reduce {
             true => {
-                let bits =
-                    self.get_bits(&res, actual_bitwidth, target_bitwidth, statements_flattened);
+                let bits = match &res.bits {
```
Schaeff: Doesn't `get_bits` take care of this caching logic already?

zokrates_core/src/flatten/mod.rs
```diff
@@ -2317,6 +2335,106 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+                                );
+                            }
+                            (lhs, rhs) => {
+                                let bit_width = T::get_required_bits();
+                                let safe_width = bit_width - 2; // dynamic comparison is not complete
+                                self.lt_check(statements_flattened, lhs, rhs, safe_width);
+                            }
+                        }
+                    }
+                    BooleanExpression::FieldLe(box lhs, box rhs) => {
```
Schaeff: This should be more consistent and avoid duplicating code:
- `enforce_constant_le_check` has a different signature than `enforce_constant_lt_check` (one takes the constant, the other one the bits of the constant).
- We should use the fact that `assert(a < c) <==> assert(c != 0 && a <= c - 1)` to rewrite `Lt` as `Le` or the other way around. We use `enforce_constant_le_check` in the definition of `enforce_constant_lt_check` now, so maybe we can remove `enforce_constant_lt_check` entirely and rewrite everything to `enforce_constant_le_check` here.

zokrates_core/src/flatten/mod.rs
Schaeff: The ideal would be to fail earlier at compile time but it seems like it cannot be implemented in flattening without changing a lot of things. Maybe propagation could be enhanced with rules for < > <= >= if needed.

zokrates_core/src/flatten/mod.rs
```diff
@@ -1847,18 +1874,17 @@ impl<'ast, T: Field> Flattener<'ast, T> {
 
     fn get_bits(
         &mut self,
-        e: &FlatUExpression<T>,
+        e: &FlatExpression<T>,
```
Schaeff: Can you explain the reasoning here? Is it that the `bits` in `FlatUExpression` has the same role as the bits cache?

Schaeff (changes requested):

zokrates_core/src/flatten/mod.rs
dark64: I actually reverted this change, I believe I misunderstood this part of the code and the `FlatUExpression` type

zokrates_core/src/flatten/mod.rs
dark64: Yeah, I wasn't happy with this thing, it was actually an attempt to reduce some code by extracting the logic to a function, but it resulted in having similar functions but with different signatures making in inconsistent

zokrates_core/src/flatten/mod.rs
dark64: Nice catch, yeah we can propagate this thing earlier

zokrates_core/src/flatten/mod.rs
dark64: I think I'm gonna just write everything here

zokrates_core/src/flatten/mod.rs
```diff
@@ -16,10 +16,8 @@ use crate::flat_absy::{RuntimeError, *};
 use crate::solvers::Solver;
 use crate::zir::types::{Type, UBitwidth};
 use crate::zir::*;
-use std::collections::{
-    hash_map::{Entry, HashMap},
-    VecDeque,
-};
+use std::collections::hash_map::Entry;
```
Schaeff: revert

zokrates_core/src/flatten/mod.rs
```diff
@@ -427,10 +426,55 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+    fn enforce_constant_le_check(
+        &mut self,
+        statements_flattened: &mut FlatStatements<T>,
+        e: FlatExpression<T>,
+        c: T,
+        error: RuntimeError,
+    ) {
+        let c_bit_width = c.bits() as usize;
+        let c_bits_be = c.to_bits_be();
+
```
Schaeff, line 451: add comment

zokrates_core/src/flatten/mod.rs
```diff
@@ -427,10 +426,55 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+        error: RuntimeError,
+    ) {
+        // `e < 0` will always result in false value, so we constrain `0 == 1`
+        if c == T::zero() {
+            statements_flattened.push_back(FlatStatement::Condition(
+                T::zero().into(),
+                T::one().into(),
+                error,
+            ));
+        } else {
```
Schaeff, line 492: add comment

zokrates_core/src/flatten/mod.rs
```diff
@@ -620,59 +664,27 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
-                    box FlatExpression::Identifier(*bit),
-                    box FlatExpression::Identifier(*bit),
-                ),
-                RuntimeError::ConstantLtBitness,
-            ));
-        }
-
-        // bit decomposition check
-        let mut e_sum = FlatExpression::Number(T::from(0));
+        let e: FlatExpression<T> = self.define(e, statements_flattened).into();
```
Schaeff: is this needed?

zokrates_core/src/flatten/mod.rs
```diff
@@ -620,59 +664,27 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+            bitwidth,
+            bitwidth,
+            statements_flattened,
             RuntimeError::ConstantLtSum,
-        ));
+        );
 
         // check that this decomposition does not overflow the field
-        self.enforce_constant_le_check(
+        self.enforce_constant_le_check_bits(
```
Schaeff, line 699: Can we use `enforce_constant_le_check` directly ?

zokrates_core/src/flatten/mod.rs
```diff
@@ -620,59 +664,27 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
             statements_flattened,
             &e_bits_be,
             &T::max_value().to_bits_be(),
+            RuntimeError::Le,
         );
 
-        let conditions = self.constant_le_check(statements_flattened, &e_bits_be, &c.to_bits_be());
```
Schaeff, line 675: revert

zokrates_core/src/flatten/mod.rs
dark64: I didn't think much of it because it was like that before, but I assume it's not necessarily needed here

zokrates_core/src/flatten/mod.rs
dark64: Hmm, we need bits of `e` after this call so I think it's fine as is, if we call it directly then we are doing unnecessarily calls and potentially cloning stuff.

Schaeff: I moved this to `get_bits`: if you call `get_bits` and want a decomposition of max bitwidth, you always need that check because of non determinism.

Schaeff: Ok now I understand it's not possible, as in some cases it would mean checking both `< P` (in `get_bits`) and `< C` (for a `< C` check)

dark64: yeah

Schaeff (approved):

Issue #1011, opened 2021-09-20
zokrates-js: Fatal JavaScript invalid size error


j-szulc:
## Description

It seems like functions with a lot of arguments are compiled just fine with Zokrates CLI, but they quickly create fatal errors in V8 when compiled via zokrates-js. 

node build.js:
```
#
# Fatal error in , line 0
# Fatal JavaScript invalid size error 202299660
#
#
#
#FailureMessage Object: 0x7ffc47af71b0
 1: 0xb691f1  [node]
 2: 0x1bf3094 V8_Fatal(char const*, ...) [node]
 3: 0xe82a58  [node]
 4: 0xe8fde5 v8::internal::Factory::NewJSArrayStorage(v8::internal::ElementsKind, int, v8::internal::ArrayStorageAllocationMode) [node]
 5: 0xe8feee v8::internal::Factory::NewJSArray(v8::internal::ElementsKind, int, int, v8::internal::ArrayStorageAllocationMode, v8::internal::AllocationType) [node]
 6: 0xf9eb36 v8::internal::JsonParser<unsigned short>::BuildJsonArray(v8::internal::JsonParser<unsigned short>::JsonContinuation const&, v8::base::SmallVector<v8::internal::Handle<v8::internal::Object>, 16ul> const&) [node]
 7: 0xfa6934 v8::internal::JsonParser<unsigned short>::ParseJsonValue() [node]
 8: 0xfa772f v8::internal::JsonParser<unsigned short>::ParseJson() [node]
 9: 0xd7ad8b  [node]
10: 0xd7b396 v8::internal::Builtin_JsonParse(int, unsigned long*, v8::internal::Isolate*) [node]
11: 0x15e68b9  [node]
[1]    1496129 trace trap  node build.js
29.25s user 1.58s system 111% cpu 27.756s total
```

## Environment

- Compiler version: zokrates-js@1.0.35, node v16.9.1
- Operating system: Linux

## Steps to Reproduce

test.zok
```zokrates
import "hashes/keccak/256bit" as keccak256

struct Result{
    u64[4] a
    u64[4] b
    u64[4] c
    u64[4] d
    u64[4] e
}

def main(u64[4] a, u64[4] b, u64[4] c, u64[4] d, u64[4] e) -> (Result):
    return Result{ a: keccak256(a), b: keccak256(b), c: keccak256(c), d:keccak256(d), e:keccak256(e)}
```

build.js:
```js
[...]

PR #1083, opened 2022-01-08
Optimize bindings in zokrates-js causing encoding and OOM issues

dark64:
https://github.com/Zokrates/ZoKrates/issues/1011


zokrates_js/src/lib.rs
```diff
@@ -20,18 +20,30 @@ use zokrates_core::typed_absy::abi::Abi;
...
+pub struct CompilationResult {
+    program: Vec<u8>,
+    abi: String,
+}
+
+#[wasm_bindgen]
+impl CompilationResult {
+    pub fn get_program(&self) -> js_sys::Uint8Array {
+        let arr = js_sys::Uint8Array::new_with_length(self.program.len() as u32);
+        arr.copy_from(&self.program);
```
Schaeff, line 33: Can you explain why copying is required here?

zokrates_js/index.d.ts
```diff
@@ -59,12 +64,12 @@ declare module 'zokrates-js' {
   }
 
   export interface ZoKratesProvider {
-    compile(source: string, options?: CompileOptions): CompilationArtifacts;
-    setup(program: Uint8Array): SetupKeypair;
-    computeWitness(artifacts: CompilationArtifacts, args: any[]): ComputationResult;
-    exportSolidityVerifier(verificationKey: VerificationKey): string;
-    generateProof(program: Uint8Array, witness: string, provingKey: Uint8Array): Proof;
-    verify(verificationKey: VerificationKey, proof: Proof): boolean;
+    compile(source: string, options?: CompileOptions): Promise<CompilationArtifacts>;
```
Schaeff: Can you explain why this was made asynchronous?

zokrates_js/src/lib.rs
dark64: There are some concerns around lifetimes if memory is only viewed (see here: https://docs.rs/js-sys/latest/js_sys/struct.Uint8Array.html#unsafety). There are ways to avoid copying I guess, but we would need to use unsafe code for that.

.circleci/config.yml
```diff
@@ -110,7 +110,7 @@ jobs:
           command: WITH_LIBSNARK=1 RUSTFLAGS="-D warnings" ./integration_test.sh
   deploy:
     docker:
-      - image: circleci/python:latest-node
```
dark64, line 113: This image was the cause of the broken zokrates.js builds, it is considered legacy now so I've changed it to a newer fixed version `cimg/python:3.8-node`

Schaeff (approved):

PR #1098, opened 2022-02-08
Fix encoding issue causing invalid values on u64 inputs

### LINKED ISSUE #1096: Witness computation disagrees between binary and js module
## Description

Computing a witness using the binary succeeds whilst performing the same witness computation using `zokrates-js` fails.

This seems to be inconsistent behaviour between the `zokrates` binary and the `zokrates-js` library.

## Environment

- Compiler version: `zokrates: 0.7.11`, `zokrates-js: 1.0.39`
- Operating system: macOS 11.3.1

## Steps to Reproduce

Using the `0.7.11` docker image binary and `1.0.39` of `zokrates-js`.

```
// test.zok
import "hashes/sha3/256bit" as sha3

def main(u64 currentTime, private u64[4] expected) -> u64[4]:
    u64[4] h = sha3([currentTime])
    assert(h[0] == expected[0])
    assert(h[1] == expected[1])
    assert(h[2] == expected[2])
    assert(h[3] == expected[3])
    return h
```

```bash
$ zokrates compile -i test.zok
$ zokrates setup
$ zokrates compute-witness -a 1643823900 16213210585803108009 8295259912202483971 10491929171841072598 9200200322684122894
Computing witness...
Witness file written to 'witness'
$ zokrates generate-proof
Generating proof...
WARNING: You are using the G16 scheme which is subject to malleability. See zokrates.github.io/toolbox/proving_schemes.html#g16-malleability for implications.
Proof written to '/local/proof.json'
```

Witness computation and then proof generation succeeds using the `zokrates` binary in the `0.7.11` docker container.

Then in `zokrates-js`:

```javascript
initialize().then((provider) => {
    const source = 'import "hashes/sha3/256bit" as sha3             \n \
        \n \
    def main(u64 currentTime, private u64[4] expected) -> u64[4]: \n \
            u64[4] h = sha3([currentTime]) \n \
            assert(h[0] == expected[0]) \n \
            assert(h[1] == expected[1]) \n \
            assert(h[2] == expected[2]) \n \
            assert(h[3] == expected[3]) \n \
            return h'

    const artifacts = provider.compile(source);
    const { witness, output } = provider.computeWitness(artifacts, ['1643823900', ['16213210585803108009', '8295259912202483971', '10491929171841072598', '9200200322684122894']]);
    // Execution failed: Assertion failed at main.zok:5:18
})
```

Passing in the exact same witness to the javascript library causes it to fail on the first assertion line even though the witness computation was done successfully using the same circuit on the binary.
-->

dark64:
Closes https://github.com/Zokrates/ZoKrates/issues/1096

We need to be careful with using `as usize` casts (`usize` is `u32` in wasm so it produces unexpected values with `u64` type). I had a look at the entire codebase and it seems everything else is fine, the only cast causing issues is the one fixed in this PR, residing in the `zokrates_abi` crate. To combat this from happening again, we should have more integration tests for js environments, starting at running all tests in `zokrates_core_test`. This will be done in a separate PR.

### DIFF
--- a/zokrates_core/src/proof_system/mod.rs
+++ b/zokrates_core/src/proof_system/mod.rs
@@ -12,13 +12,18 @@ pub use self::scheme::*;
 pub use self::solidity::*;
 
 use crate::ir;
-use rand_0_4::Rng;
+
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
-use std::io::{Read, Write};
-#[cfg(feature = "bellman")]
-use zokrates_field::BellmanFieldExtensions;
 use zokrates_field::{Bls12_377Field, Bls12_381Field, Bn128Field, Field};
+
+cfg_if::cfg_if! {
+    if #[cfg(feature = "bellman")] {
+        use rand_0_4::Rng;
+        use std::io::{Read, Write};
+        use zokrates_field::BellmanFieldExtensions;
+    }
+}
 
 pub trait NotBw6_761Field {}
 impl NotBw6_761Field for Bls12_377Field {}
--- a/zokrates_core/src/static_analysis/zir_propagation.rs
+++ b/zokrates_core/src/static_analysis/zir_propagation.rs
@@ -13,7 +13,7 @@ type Constants<'ast, T> = HashMap<Identifier<'ast>, ZirExpression<'ast, T>>;
 
 #[derive(Debug, PartialEq)]
 pub enum Error {
-    OutOfBounds(u128, u128),
+    OutOfBounds(usize, usize),
     DivisionByZero,
     AssertionFailed(RuntimeError),
 }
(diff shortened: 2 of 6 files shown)


zokrates_core/src/typed_absy/types.rs
```diff
@@ -544,7 +544,7 @@ pub enum UBitwidth {
 
 impl UBitwidth {
     pub fn to_usize(self) -> usize {
-        self as u32 as usize
```
dark64, line 547: unnecessary cast to `u32`

zokrates_core/src/static_analysis/zir_propagation.rs
```diff
@@ -13,7 +13,7 @@ type Constants<'ast, T> = HashMap<Identifier<'ast>, ZirExpression<'ast, T>>;
 
 #[derive(Debug, PartialEq)]
 pub enum Error {
-    OutOfBounds(u128, u128),
+    OutOfBounds(usize, usize),
```
dark64, line 16: `u128` is not actually needed here

zokrates_core/src/proof_system/mod.rs
```diff
@@ -12,14 +12,19 @@ pub use self::scheme::*;
...
+
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
-use std::io::{Read, Write};
-#[cfg(feature = "bellman")]
-use zokrates_field::BellmanFieldExtensions;
 use zokrates_field::{Bls12_377Field, Bls12_381Field, Bn128Field, Field};
 
+cfg_if::cfg_if! {
+    if #[cfg(feature = "bellman")] {
```
dark64, line 21: these imports were unused if `bellman` feature is not enabled

Schaeff (approved):

PR #1103, opened 2022-02-23
Marlin solidity verifier

nirvantyagi:
**Work in progress**
This PR implements the verifier for Marlin proofs produced in ZoKrates. It does the following:

- Updates Marlin data structures from raw bytes to structured group and field elements
- Updates `arkworks` dependencies to version `0.3`
- Adds `zokrates_solidity_test` crate to provide utils for executing and testing solidity contracts on a local EVM
- Implements `SolidityCompatibleScheme` for Marlin by providing verifier contract

This PR supersedes #1035 and #1036.

PR #1104, opened 2022-03-07
Contract testing with local EVM

nirvantyagi:
This PR adds support for local unit tests of solidity contracts by compiling and running contracts locally on an EVM. It is modeled after the work done by the Fe team [1, 2]. Perhaps in the future, a shared testing library can be maintained between the two projects.

A simple example of compilation, deploying, and calling a contract is provided in ``lib.rs``. A more complex example for testing the verification of the Marlin SNARK is provided in a follow-up PR (https://github.com/Zokrates/ZoKrates/pull/1103).

[1] - https://github.com/ethereum/fe
[2] - https://github.com/WilfredTA/fe/tree/tannr/evm-upgrade


dark64:
Hey @nirvantyagi, thanks for this, looks good to me. It seems like a newer `cmake` version is required ( `>=3.13.0`) for solc rust bindings to compile, I will try to update our environment image to hopefully resolve this issue.

PR #1107, opened 2022-03-10
Up ubuntu version to 20.04 in the environment image, replace git2 with local git wrapping

dark64:


.circleci/config.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           command: cargo generate-lockfile
       - restore_cache:
           keys:
-            - cargo-cache-{{ .Environment.CACHE_VERSION }}-{{ arch }}-{{ checksum "Cargo.lock" }}
+            - cargo-cache-v{{ .Environment.CACHE_VERSION }}-{{ arch }}-{{ checksum "Cargo.lock" }}
```
Schaeff: does this have any impact?

Schaeff (changes requested):
LGTM apart from these cache changes, and the PR should be renamed to reflect the changes.

.circleci/config.yml
dark64: No, but I forgot to revert this 🤦

Schaeff (approved):

PR #1121, opened 2022-04-07
Use curly braces and semi colons

dark64:
### **Breaking change**

Changes made in this PR:

- Use curly braces `{}` to encapsulate blocks (actual block expressions are not implemented in this PR)
- Use a semicolon to separate statements instead of a new line
- Change `if else` expression to accept blocks, but disallow usage of statements in the block
- Align the whole codebase with the new syntax (examples, core tests, stdlib, and stdlib tests)
- Fix `fmt::Display` implementations for AST's


zokrates_core/src/semantics.rs
```diff
@@ -2461,10 +2461,24 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+            Expression::Conditional(box conditional) => {
+                let condition_checked =
+                    self.check_expression(*conditional.condition, module_id, types)?;
+
+                if !conditional.consequence_statements.is_empty()
+                    || !conditional.alternative_statements.is_empty()
+                {
+                    return Err(ErrorInner {
+                        pos: Some(pos),
+                        message: "statements in the if-else block are not yet supported"
```
dark64: @Schaeff not sure how to cleanly write this, any better ideas?

changelogs/unreleased/1121-dark64
```diff
@@ -0,0 +1 @@
+Use curly braces to enclose blocks
```
Schaeff: This is not accurate

zokrates_core/src/semantics.rs
Schaeff:
```suggestion
                        message: "Statements are not supported in conditional branches"
```

maybe like this?

zokrates_core/src/zir/mod.rs
```diff
@@ -125,13 +125,13 @@ impl<'ast, T: fmt::Display> fmt::Display for ZirStatement<'ast, T> {
...
-                write!(f, "")
+                write!(f, ";")
             }
-            ZirStatement::Definition(ref lhs, ref rhs) => write!(f, "{} = {}", lhs, rhs),
+            ZirStatement::Definition(ref lhs, ref rhs) => write!(f, "{} = {};", lhs, rhs),
             ZirStatement::IfElse(ref condition, ref consequence, ref alternative) => {
                 write!(
                     f,
-                    "if {} then {{{}}} else {{{}}} fi",
+                    "if {} {{{}}} else {{{}}}",
```
Schaeff:
```suggestion
                    "if {} {{{}}} else {{{}}};",
```

For consistency

Schaeff (changes requested):
Looks good, just a few tweaks required

zokrates_core/src/typed_absy/mod.rs
```diff
@@ -1051,11 +1051,27 @@ impl<'ast, T, E> ConditionalExpression<'ast, T, E> {
...
         match self.kind {
-            ConditionalKind::IfElse => write!(
-                f,
-                "if {} {{ {} }} else {{ {} }}",
-                self.condition, self.consequence, self.alternative
-            ),
+            ConditionalKind::IfElse => {
+                let consequence = self.consequence.to_string();
+                let alternative = self.alternative.to_string();
+                let is_block = consequence.starts_with("{") && alternative.starts_with("{");
```
Schaeff: This looks hacky, maybe there is a better way? Before `to_string` we have enums which seem cleaner for this sort of thing

Schaeff: I looked into it and the issue doesn't seem to be with the formatting code, but rather with the fact that the branch isolator currently wraps blocks in blocks. It should keep blocks as is, and wrap everything else in a block. Then formatting can be done the natural way.

Schaeff (approved):
👍

PR #1129, opened 2022-04-13
Fix bits cache panic due to different bitwidths

Schaeff:
If we cached a representation of `n` bits, it can be turned into a representation of `m >= n` bits by left padding with zeroes. Currently we force `n == m`. Relax that constraint.


dark64 (approved):
That makes sense, nice one!

PR #1136, opened 2022-04-22
fix padding inside of keccak base function

thecodingshrimp:
# Summary
The base [keccak function](https://github.com/thecodingshrimp/ZoKrates/blob/7fae54d598ad2a9240e4634acd2f58d7518d2190/zokrates_stdlib/stdlib/hashes/keccak/keccak.zok#L79) pads the wrong u64 entry in `u64[N] input` if the byte length of the input is != N * 8.

# Problem Description
https://github.com/thecodingshrimp/ZoKrates/blob/7fae54d598ad2a9240e4634acd2f58d7518d2190/zokrates_stdlib/stdlib/hashes/keccak/keccak.zok#L89-L99
In the for loop in [line 90](https://github.com/thecodingshrimp/ZoKrates/blob/7fae54d598ad2a9240e4634acd2f58d7518d2190/zokrates_stdlib/stdlib/hashes/keccak/keccak.zok#L90), `pt` is incremented after XORing it with the last 8 bytes from `u64 input[N]`. This is only correct behavior when we hash 8 * N bytes since in [line 97](https://github.com/thecodingshrimp/ZoKrates/blob/7fae54d598ad2a9240e4634acd2f58d7518d2190/zokrates_stdlib/stdlib/hashes/keccak/keccak.zok#L97) we add padding in the first byte after 8 * N.
Example: Let's assume we want to hash 4 bytes `0x9fef1ec4`. We call `keccak::<_, 256>([0x9fef1ec400000000], 0x0000000100000000)` which, in the current implementation, pads the 12th byte since we went through the loop 1 time and incremented `pt` to 1.

# Solution
Check if `pad != 0x0000000000000001` and if so, decrement `pt` to pad the correct byte.

### DIFF
--- a/zokrates_stdlib/stdlib/hashes/keccak/keccak.zok
+++ b/zokrates_stdlib/stdlib/hashes/keccak/keccak.zok
@@ -93,6 +93,9 @@ def main<N, W>(u64[N] input, u64 pad) -> u64[25]:
         q = if pt == 0 then keccakf(q) else q fi
     endfor
 
+    // if pad then start after last byte
+    pt = if pad != 0x0000000000000001 then pt - 1 else pt fi
+
     // finalize
     q[pt] = q[pt] ^ pad
     q[rate - 1] = q[rate - 1] ^ 0x8000000000000000
--- a/zokrates_stdlib/tests/tests/hashes/keccak/keccak.json
+++ b/zokrates_stdlib/tests/tests/hashes/keccak/keccak.json
@@ -1,0 +1,15 @@
+{
+    "entry_point": "./tests/tests/hashes/keccak/keccak.zok",
+    "tests": [
+      {
+        "input": {
+          "values": []
+        },
+        "output": {
+          "Ok": {
+            "values": []
+          }
+        }
+      }
+    ]
+  }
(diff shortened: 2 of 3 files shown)


dark64:
Hi @thecodingshrimp, nice catch! I don't think your solution considers all cases eg. we use the same base function for sha3 with a different `pad` value, see [here]( https://github.com/Zokrates/ZoKrates/blob/develop/zokrates_stdlib/stdlib/hashes/sha3/256bit.zok#L4). I think we didn't consider to support hashing of lengths that do not follow the `N * 8` rule mostly because this can be solved by encoding the inputs, which is usually done in solidity, etc. Not sure yet what the best solution is, feel free to share any suggestions.

```
pt = if (pad & ((1 << 8) - 1)) != 0 then pt else pt - 1 fi
```

Something like this would work but is a bit dirty.

thecodingshrimp:
Can you elaborate on what you mean with encoding in solidity? Are you referring to RLP?

Your PR looks much more sophisticated ;D

dark64:
Closing this as this was resolved in https://github.com/Zokrates/ZoKrates/pull/1139

Thanks @thecodingshrimp

PR #1143, opened 2022-05-02
Add circom/snarkjs compatible output for r1cs and witness

Schaeff:
Add optional snarkjs output for r1cs and witness so that snarkjs can be used alongside ZoKrates:
- compile with zokrates (generating a snarkjs r1cs)
- setup with snarkjs
- compute with zokrates (generating a snarkjs witness)
- proof generation with snarkjs

~~Not blocking but annoying: https://github.com/iden3/snarkjs/issues/152~~

Todo:
- [ ] Fix CLI flags


zokrates_cli/src/ops/compute_witness.rs
```diff
@@ -38,6 +39,13 @@ pub fn subcommand() -> App<'static, 'static> {
         .takes_value(true)
         .required(false)
         .default_value(cli_constants::WITNESS_DEFAULT_PATH)
+    ).arg(Arg::with_name("witness")
```
dark64: I think this should be called `circom-witness` or something similar, as we also output the witness.

xhliu:
Any update on this? Would be great to be compatible with circom/snarkjs.

Schaeff:
> Any update on this? Would be great to be compatible with circom/snarkjs.

This will be out soon

dark64 (approved):
LGTM

PR #1163, opened 2022-06-07
Avoid double wrapping of blocks in branch isolation

Schaeff:
Introduce `BlockOrExpression` so we can wrap all expressions except blocks (which are already wrapped)

### DIFF
--- a/zokrates_core/src/static_analysis/branch_isolator.rs
+++ b/zokrates_core/src/static_analysis/branch_isolator.rs
@@ -31,4 +31,16 @@ impl<'ast, T: Field> Folder<'ast, T> for Isolator {
             e.kind,
         ))
     }
+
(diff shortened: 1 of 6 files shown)


dark64:
```
# test.zok
def main(bool a) -> field {
    return if a { 1 } else { 2 };
}
```

`RUST_LOG=trace cargo run compile -i ./test.zok --verbose --isolate-branches`

```
[2022-06-13T10:02:59Z DEBUG zokrates_core::compile] Check semantics
[2022-06-13T10:02:59Z TRACE zokrates_core::compile] 
    | ./test.zok: |<---- main
    ----------------------------------------------------------------------------------------------------
    def main(bool a) -> field {
        return if a 1f else 2f; // still missing { }
    }
```

PR #1169, opened 2022-06-19
Fix tuple constant inlining

dark64:


zokrates_core/src/typed_absy/folder.rs
```diff
@@ -208,6 +208,7 @@ pub trait Folder<'ast, T: Field>: Sized {
         match t {
             Array(array_type) => Array(self.fold_declaration_array_type(array_type)),
             Struct(struct_type) => Struct(self.fold_declaration_struct_type(struct_type)),
+            Tuple(tuple_type) => Tuple(self.fold_declaration_tuple_type(tuple_type)),
```
Schaeff, line 211: The same change is required in `result_folder.rs`

Schaeff (changes requested):
Good catch, let's just make sure tuples are visited in all places

Schaeff (approved):

PR #1170, opened 2022-06-21
Remove multi returns

dark64:


changelogs/unreleased/1170-dark64
```diff
@@ -0,0 +1 @@
+Deprecate multiple returns
```
Schaeff: Is it technically deprecation and not removal?

zokrates_core/src/semantics.rs
```diff
@@ -1176,29 +1153,6 @@ impl<'ast, T: Field> Checker<'ast, T> {
 
                     match self.check_statement(stat, module_id, &state.types) {
                         Ok(statement) => {
-                            if let TypedStatement::Return(e) = &statement {
```
Schaeff, line 1179: Is this being checked somewhere else? The type of the return value must match the return type of the signature

zokrates_core/src/semantics.rs
```diff
@@ -3757,15 +3705,9 @@ mod tests {
         .mock()
     }
 
-    /// Helper function to create ((private field a): return)
+    /// Helper function to create def function1(private field a) { return; }
```
Schaeff: I didnt name these because the `Function` struct does not hold the name

zokrates_core/src/static_analysis/reducer/constants_reader.rs
```diff
@@ -33,6 +33,26 @@ impl<'a, 'ast, T: Field> ConstantsReader<'a, 'ast, T> {
 }
 
 impl<'a, 'ast, T: Field> Folder<'ast, T> for ConstantsReader<'a, 'ast, T> {
+    fn fold_declaration_constant(
```
Schaeff: seems like this was just moved up the file? maybe the diff can be made smaller by reverting the move?

Schaeff (commented):
Good job on this, just a few tweaks

changelogs/unreleased/1170-dark64
dark64: oh you are right, would be better to use the phrase "removal"

zokrates_core/src/static_analysis/reducer/constants_reader.rs
dark64: So I have this thing in my IDE where it detects weird order of functions and tries to match the same order as it is defined in a trait, this happened here. I think it's better because then it's easier to navigate.

zokrates_core/src/semantics.rs
dark64: oh I see

zokrates_core/src/semantics.rs
dark64: yes, when the return statement is checked: https://github.com/Zokrates/ZoKrates/blob/deprecate-multi-returns/zokrates_core/src/semantics.rs#L1726

zokrates_core/src/static_analysis/reducer/constants_reader.rs
dark64: But I will revert for the sake of this PR

changelogs/unreleased/1170-dark64
dark64: unfortunately, this PR will close if I rename the branch

Schaeff: it's fine not to rename

Schaeff (approved):

Issue #791, opened 2021-04-01
Signature outputs aren't being used for constant parameter inference


Schaeff:
## Description

Signature outputs aren't being used for constant parameter inference

## Environment

- Compiler version: develop
- Operating system: macos

## Steps to Reproduce

```zokrates
def foo<N>() -> field[N]:
   return [42; N]

def main() -> field[2]:
   return foo() // this should infer `N` to `2`
```

PR #1172, opened 2022-06-25
Use signature output for inference

dark64:
https://github.com/Zokrates/ZoKrates/issues/791


zokrates_core/src/semantics.rs
```diff
@@ -2155,6 +2053,153 @@ impl<'ast, T: Field> Checker<'ast, T> {
         }
     }
 
+    fn check_function_call_expression(
+        &mut self,
+        function_id: ExpressionNode<'ast>,
+        generics: Option<Vec<Option<ExpressionNode<'ast>>>>,
+        arguments: Vec<ExpressionNode<'ast>>,
+        assignee_type: Option<Type<'ast, T>>,
```
Schaeff: How about `expected_return_type`? We use this function outside of the assignee context also.

Schaeff (changes requested):
Great to finally have a fix for this!

Schaeff (approved):

PR #1192, opened 2022-07-20
Avoid validity checks on the deserialization of the proving key to improve proving time

dark64:
To deserialize the proving key, we are using [this](https://docs.rs/ark-serialize/0.3.0/ark_serialize/trait.CanonicalDeserialize.html#method.deserialize_uncompressed) function which performs validity checks and takes a long time. We can avoid performing validity checks and improve proving time by using the [unchecked](https://docs.rs/ark-serialize/0.3.0/ark_serialize/trait.CanonicalDeserialize.html#method.deserialize_unchecked) deserialization.

For the `merkleTree/sha256PathProof3.zok` example the overall proving time difference is:

before (0.8.0):  37,289s
after:     2,237s


Schaeff (approved):
Great!

PR #1193, opened 2022-07-26
Allow shadowing of variables

Schaeff:
Enable shadowing.
The change happens during the semantic phase, where `Identifiers` are annotated with a `shadow` number corresponding to the depth at which they are declared.
When accessing an identifier, the highest id so far (if any) is used: it is the only one that is accessible.
When exiting a scope, all identifiers whose shadow id is at that depth are removed.

```
field mut a = 2;
for uint i in 0..0 {
    a = 3
    field a = 4
}
a = 5

// becomes

field mut a_0 = 2; // function scope is level 0
for uint i_1 in 0..0 { // loop counter is in the next scope, so level 1
    a_0 = 3 // closest accessible `a` is `a_0`
    field a_1 = 4 // shadowing `a` leads to `a_1` because we are at scope depth 1
} // a_1 is cleared 
a_0 = 5
```


dark64 (approved):
LGTM

PR #1199, opened 2022-08-09
Refactor zokrates-js, add backend option

dark64:
- Added `bellman` backend as an option (again)
- Added a log writer to print messages from a log statement
- Removed gulp tasks, rewrote all exports and packing in rust


zokrates_js/build.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    let out_dir = env::var_os("OUT_DIR").unwrap();
+    let dest_path = Path::new(&out_dir).join("stdlib.json");
+
+    fs::write(dest_path, stdlib.dump()).unwrap();
+}
+
+fn export_metadata() {
+    let path = "../zokrates_cli/Cargo.toml";
+    let config: toml::Value = dbg!(toml::from_str(&fs::read_to_string(path).unwrap()).unwrap());
```
Schaeff: does this `dbg!` need to stay?

zokrates_js/build.rs
```diff
@@ -0,0 +1,45 @@
...
+    export_metadata();
+}
+
+fn export_stdlib_js() {
+    let root = "../zokrates_stdlib/stdlib";
+    let mut stdlib = json::JsonValue::new_object();
+
+    for entry in WalkDir::new(root)
+        .into_iter()
+        .filter_map(Result::ok)
```
Schaeff: This ignores errors, instead it should probably throw if one happens: `.map(Result::unwrap)`

zokrates_book/src/toolbox/zokrates_js.md
```diff
@@ -123,12 +128,13 @@ const fileSystemResolver = (from, to) => {
...
 ```
 
-##### computeWitness(artifacts, args)
+##### computeWitness(artifacts, args[, options])
 Computes a valid assignment of the variables, which include the results of the computation.
 
 Parameters:
 * `artifacts` - Compilation artifacts
 * `args` - Array of arguments (eg. `["1", "2", true]`)
+* `options` - Computation options
```
Schaeff, line 137: The log stream is not mentioned here, should it be?

Schaeff (changes requested):

zokrates_js/build.rs
dark64: whops

zokrates_book/src/toolbox/zokrates_js.md
dark64: I'm not sure about this one, I think most people won't touch this part, I added this just to have logs in the playground. It is possible to check the options if typescript is used though. The same lack of documentation is present for the compilation options too. I think the whole document should have a better structure, which can be improved in a separate PR.

Schaeff (approved):

PR #1201, opened 2022-08-11
Cast and string utils

nanaknihal:
I made a few new utils for type casts and parsing strings (represented as u8 arrays). I've tested them locally but was having difficulty adding and running tests in `zokrates_stdlib/tests/tests/utils`. Compiler was saying were saying the new utils were not found, even though I rebuilt after adding the utils. I rebuilt with [these](https://zokrates.github.io/gettingstarted.html#from-source) instructions.

I also wrote a base64 conversion string util which could be helpful, but requires these as dependencies, so perhaps let's make sure I contribute these correctly before submitting those.

### DIFF
--- a/zokrates_stdlib/stdlib/utils/casts/bool_array_to_u8_array.zok
+++ b/zokrates_stdlib/stdlib/utils/casts/bool_array_to_u8_array.zok
@@ -1,0 +1,13 @@
+from "EMBED" import u8_from_bits;
+
+def main<N, P>(bool[N] bits) -> u8[P] {
+    assert(N == 8 * P);
+
+    u8[P] mut res = [0; P];
+
+    for u32 i in 0..P {
+        res[i] = u8_from_bits(bits[8 * i..8 * (i + 1)]);
+    }
+
+    return res;
+}
--- a/zokrates_stdlib/stdlib/utils/casts/u8_array_to_bool_array.zok
+++ b/zokrates_stdlib/stdlib/utils/casts/u8_array_to_bool_array.zok
@@ -1,0 +1,16 @@
+from "EMBED" import u8_to_bits;
+
+def main<N, P>(u8[N] input) -> bool[P] {
+    assert(P == 8 * N);
+
+    bool[P] mut res = [false; P];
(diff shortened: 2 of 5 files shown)


Schaeff:
Hi @nanaknihal! Thanks for the PR and apologies fr the delay. Some thoughts here:
- can you explain a bit more your use case when it comes to mapping bytes to a field element using ASCII encoding? It seems to me that this method is broken: `string[0] - 48` may underflow, and `12 == 12 * 10**0 == 2*10**0 + 1 * 10**1` so the two values `0x0c` and `0x0102` map to the same field element `12`. Are you assuming that all characters represented as u8 are going to be letters? Even then, it seems that by using buckets of 10 you cannot fit the whole alphabet without having collisions like the one I shared above.
- when it comes to the utils to go between bool and u8, we've been working on a more complete and ergonomic set of tools, see #1215
Cheers

nanaknihal:
Thanks for the review @Schaeff 
It has the assumption all inputs are ASCII representations of numbers, i.e. between 48 (representation of 0) and 57 (representation of 9). This assumption was made since there's no need to worry about malformed inputs for our use-case (verifying JWTs signed by an authority). You are right; for an open-source library where other users can't make those security assumptions, it should assert input validity to avoid those errors.

With an assertion that `(string[i] >= 48) && (string[i] <=57)`, I think underflow and collisions are not an issue anymore. All values will be between 0 and 9, so neither underflow nor `12*10**0` can happen.

I added that line in a new commit to this PR; however, it adds ~8000 constraints to cast a 10-digit integer. So, I broke it up into two fucntions, `unsafe_efficient` and `safe_inefficient`, instead of one `main` function. If more convenient, i can add a main function which is just a wrapper for `safe_inefficient`

Also -- new casting overloads seems like a really elegant interface

Schaeff:
Right, if you have stronger assumptions, then this may well be safe. And yes, the range check ends up being expensive. In general my opinion is that this is not standard/unconditionally safe/efficient enough to be added to the standard library and should rather live outside the zokrates repo.

PR #1206, opened 2022-08-16
Introduce dead code elimination

### LINKED ISSUE #303: Skip computation of unused variables
field[2] infinity = [context[2], context[3]] 

    field[2] doubledP = pt
    field[2] accumulatedP = infinity
    
    for field i in 0..256 do
        field j = 255 - i
        candidateP = add(accumulatedP, doubledP, context)
        accumulatedP = if exponent[j] == 1 then candidateP else accumulatedP fi
        doubledP = add(doubledP, doubledP, context)
    endfor

in the last execution of the iteration the variable `doubledP` is computed while never being used. 
The optimiser should catch this and not compute this value.

Schaeff:
Dead code is code which does not influence the result of execution. We can thus avoid executing dead code.
The tricky part here is that the "result of execution" is not only the return value. It is also any unverified constraint.

```
def main(field x) {
   field y = 1 / x;
   return;
}
```

`1 / x` here does not influence the return value (as we return `()`), but is `x == 0` then the computation throws. The idea of this change is that we extract the part of any expression which can cause the computation to throw, and never consider that dead code. Once that's removed, the associated checks can be removed on the expression itself, so that it can be removed entirely if it does not influence the result.

So we turn:

```
def main(field x) {
   field y = 1 / x;
   return;
}
```

(with panic extraction) into

```
def main(field x) {
   assert(x != 0);
   field y = 1 / x; // here we can assume that x != 0 as we just asserted that
   return;
}
```

and then (with dead code elimination) into

```
def main(field x) {
   assert(x != 0);
   return;
}
```

Dead code elimination starts from the end of the program, adds all used variables to a set when they are seen in an expression. If it reaches the definition of a variable which is not in the set of variables used after it, that definition can be removed entirely.

Other changes in this PR:
- Refactor `typed_ast` removing `EmbedCallDefinition` and including it in `Definition`
- Fix formatting in zir (some parentheses are missing)
- Simplify zir (remove `>` and `>=` and express them in terms of `<` and `<=`)

closes #303


zokrates_core/src/static_analysis/panic_extractor.rs
```diff
@@ -0,0 +1,185 @@
...
+                },
+                _,
+            )
+            | e @ BooleanExpression::UintLt(
+                _,
+                box UExpression {
+                    inner: UExpressionInner::Value(_),
+                    ..
+                },
+            ) => fold_boolean_expression(self, e),
```
dark64: Is this arm necessary? There is a default one that calls the same on line `182`

zokrates_core_test/tests/tests/div.json
```diff
@@ -0,0 +1,65 @@
+{
```
dark64, line 1: It seems like this file uses both tabs and spaces which looks weird

zokrates_core/src/static_analysis/panic_extractor.rs
Schaeff: Yes it is because in the case where in `a < b`, `a` or `b` is a constant, the comparison check is complete (works for all values) so it does not generate a panic. The next match case treats other cases where both `a` and `b` are unknown at compile time, which leads to a possible panic. If we remove this arm, then we would add an unnecessary check as these patterns would match the next case. Maybe this should be documented better in the code?

Schaeff: Actually `UintLt` could be removed, but not `FieldLt`.

zokrates_core_test/tests/tests/div.json
Schaeff: Fixed in #1221

zokrates_core/src/static_analysis/panic_extractor.rs
Schaeff: Simplified and added comment

dark64 (approved):

PR #1215, opened 2022-08-23
Add a `cast` module to stdlib

dark64:


Schaeff:
How about performance on this? I understood it was the main blocker?

dark64:
> How about performance on this? I understood it was the main blocker?

Not a blocker

zokrates_stdlib/tests/tests/utils/casts/field.zok
```diff
@@ -0,0 +1,25 @@
+from "utils/casts.zok" import cast;
```
Schaeff: I'm not sure we should have `field -> u*` in casts because this is not zero-cost. How about overloading `unpack` instead?

zokrates_stdlib/stdlib/utils/casts.zok
```diff
@@ -0,0 +1,556 @@
...
+def cast(field input) -> u32 {
+    bool[32] bits = unpack(input);
+    return u32_from_bits(bits);
+}
+
+// Cast a field element to a 64-bit unsigned integer (u64)
+def cast(field input) -> u64 {
+    bool[64] bits = unpack(input);
+    return u64_from_bits(bits);
+}
```
Schaeff: I suggest moving this to `unpack`

Schaeff (dismissed):

Schaeff (approved):

PR #1222, opened 2022-08-29
Make return statement optional

dark64:
Makes it possible to write the following 

```zokrates
def main() { 
    return; 
}
``` 

as `def main() {}`


zokrates_cli/examples/compile_errors/no_return.zok
```diff
@@ -1,5 +0,0 @@
-def foo() {}
-
-def main() {
```
Schaeff, line 3: We should still have this test, just need to change the signature to one which requires a return statement.

zokrates_core/src/semantics.rs
```diff
@@ -1186,10 +1186,22 @@ impl<'ast, T: Field> Checker<'ast, T> {
                 }
 
                 if !found_return {
-                    errors.push(ErrorInner {
-                        pos: Some(pos),
-                        message: "Expected a return statement".to_string(),
-                    });
+                    match &s.output {
+                        box DeclarationType::Tuple(tuple_type)
+                            if tuple_type.elements.is_empty() =>
```
Schaeff: How about adding a method on `GTupleType` to make this less verbose? Or even on `GType`? `is_empty_tuple`? This could be used elsewhere also.
Also I suspect there's a way to remove `box` from the match arm without taking ownership? `match *&s.output` maybe?

zokrates_core/src/semantics.rs
```diff
@@ -1186,10 +1186,22 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+                        {
+                            statements_checked.push(TypedStatement::Return(TypedExpression::Tuple(
```
Schaeff: Same here, `TypedExpression::empty_tuple()`

Schaeff (changes requested):

zokrates_core/src/semantics.rs
dark64: yep with `&*`

zokrates_cli/examples/compile_errors/no_return.zok
dark64: true :+1:

zokrates_book/src/language/functions.md
```diff
@@ -18,4 +20,8 @@ The generic parameters can be provided explicitly, especially when they cannot b
 
 ```zokrates
 {{#include ../../../zokrates_cli/examples/book/explicit_generic_parameters.zok}}
-```
\ No newline at end of file
+```
+
+Functions that "don't" return a value, actually return an empty tuple `()`. 
+When a function returns an empty tuple `()`, the return type can be omitted from 
+the signature. In this case, a return statement can be omitted.
```
Schaeff:
```suggestion
If the return type of a function is the empty tuple `()`, the return type as well as the return statement can be omitted.
```

Then link to code like

```
def main() {
}
```

Schaeff (approved):

PR #1224, opened 2022-09-06
Disallow the use of the private keyword on non-entrypoint functions

### LINKED ISSUE #1204: Parameter visibility on non-main functions
## Description

It is currently possible to use the visibility keyword on a parameter defined on a non-main function, which is semantically not valid.

## Environment

- Compiler version: 0.8.0
- Operating system: Ubuntu

## Steps to Reproduce

```zokrates
def mul(private field a) -> field { // `private` should not be allowed here
    return a * a;
}

def main(private field a, field b) {
    assert(mul(a) == b);
    return;
}
```

dark64:
Closes https://github.com/Zokrates/ZoKrates/issues/1204

### DIFF
--- a/changelogs/unreleased/1224-dark64
+++ b/changelogs/unreleased/1224-dark64
@@ -1,0 +1,1 @@
+Disallow the use of the `private` keyword on non-entrypoint functions
--- a/changelogs/unreleased/1224-dark64
+++ b/changelogs/unreleased/1224-dark64
@@ -1,1 +1,1 @@
-Disallow the use of the `private` keyword on non-entrypoint functions
+Disallow the use of the `private` and `public` keywords on non-entrypoint functions
--- a/zokrates_cli/examples/compile_errors/non_main_private_args.zok
+++ b/zokrates_cli/examples/compile_errors/non_main_private_args.zok
@@ -1,0 +1,8 @@
+def mul(private field a) -> field { // `private` should not be allowed here
+    return a * a;
+}
+
+def main(private field a, field b) {
+    assert(mul(a) == b);
+    return;
+}
--- a/zokrates_core/src/semantics.rs
+++ b/zokrates_core/src/semantics.rs
@@ -89,6 +89,7 @@ type ConstantMap<'ast, T> =
 /// The global state of the program during semantic checks
 #[derive(Debug)]
 struct State<'ast, T> {
+    main_id: OwnedModuleId,
     /// The modules yet to be checked, which we consume as we explore the dependency tree
     modules: Modules<'ast>,
     /// The already checked modules, which we're returning at the end
@@ -166,8 +167,9 @@ impl<'ast, T: std::cmp::Ord> SymbolUnifier<'ast, T> {
 }
 
 impl<'ast, T: Field> State<'ast, T> {
-    fn new(modules: Modules<'ast>) -> Self {
+    fn new(modules: Modules<'ast>, main_id: OwnedModuleId) -> Self {
         State {
+            main_id,
             modules,
             typed_modules: BTreeMap::new(),
             types: BTreeMap::new(),
@@ -340,12 +342,13 @@ impl<'ast, T: Field> Checker<'ast, T> {
         &mut self,
         program: Program<'ast>,
     ) -> Result<TypedProgram<'ast, T>, Vec<Error>> {
-        let mut state = State::new(program.modules);
+        let main_id = program.main.clone();
 
+        let mut state = State::new(program.modules, main_id.clone());
         let mut errors = vec![];
 
         // recursively type-check modules starting with `main`
-        match self.check_module(&program.main, &mut state) {
+        match self.check_module(&main_id, &mut state) {
             Ok(()) => {}
             Err(e) => errors.extend(e),
         };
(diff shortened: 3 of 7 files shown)


zokrates_core/src/semantics.rs
```diff
@@ -762,24 +763,40 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
-                        );
+                        // parameters defined on a non-entrypoint function should not be set as private
+                        match (state.main_id != module_id || declaration.id != "main")
+                            && funct.arguments.iter().any(|a| a.private)
+                        {
+                            true => {
+                                errors.push(
+                                    ErrorInner {
+                                        pos: Some(pos),
+                                        message: "Private argument(s) are only allowed on the entrypoint function".into(),
```
Schaeff: This is not strict enough: in `from_ast.rs` we have the following:

```rust
let is_private = param
            .visibility
            .map(|v| match v {
                pest::Visibility::Private(_) => true,
                pest::Visibility::Public(_) => false,
            })
            .unwrap_or(false);
```

Based on this, the semantic checked cannot differentiate between `a` and `public a`, but the latter should not be allowed.

`absy` should be changed to include `Option<bool>` for the visibility, and it should be checked to be `None` for non main, and default to `true` for main.

Schaeff: Also the book seems to be lacking documentation on the visibility keywords, and generally on the entry point.

zokrates_core/src/semantics.rs
```diff
@@ -3755,10 +3772,9 @@ mod tests {
     fn function1() -> FunctionNode<'static> {
         let statements = vec![Statement::Return(None).mock()];
 
-        let arguments = vec![untyped::Parameter {
-            id: untyped::Variable::immutable("a", UnresolvedType::FieldElement.mock()).mock(),
-            is_private: true,
-        }
+        let arguments = vec![untyped::Parameter::public(
```
Schaeff: Why is this change required?

zokrates_core/src/semantics.rs
```diff
@@ -3952,18 +3970,18 @@ mod tests {
...
+            // def foo(field[3] a) {
             //   return;
             // }
             //
             // should succeed as P could be different from 3
 
             let mut f0 = function0();
 
-            f0.value.arguments = vec![untyped::Parameter::private(
+            f0.value.arguments = vec![untyped::Parameter::public(
```
Schaeff: Ok this seems to be related to what I mentioned above, the visibility should be an option on each parameter in the untyped ast, and in the typed ast it should probably only appear in TypedProgram, not on each function if possible.

Schaeff (changes requested):
The changes to the semantic checker are nice but I think we should encode this more in our types in untyped and typed ast.

zokrates_core/src/semantics.rs
dark64: I found this inconsistent with other tests, we use the wrapper functions everywhere except here. The parameter must be public otherwise the test fails because of the check introduced by this PR.

zokrates_stdlib/stdlib/signatures/verifyEddsa.zok
```diff
@@ -29,7 +29,7 @@ import "utils/casts/u32_8_to_bool_256";
 ///
 /// Returns:
 ///     Return true for S being a valid EdDSA Signature, false otherwise.
-def main(private field[2] R, private field S, field[2] A, u32[8] M0, u32[8] M1, BabyJubJubParams context) -> bool {
+def main(field[2] R, field S, field[2] A, u32[8] M0, u32[8] M1, BabyJubJubParams context) -> bool {
```
dark64, line 32: @Schaeff also this is the one example I was talking about, this module had `private` parameters but is used as a submodule elsewhere

Schaeff: Right, this change makes sense!

Schaeff:
Nice simplification of the implementation but I think we can go further by doing the following:

- Remove the `typed::GParameter` type entirely, replace all its occurrences with `typed::GVariable`
- Store a single array of `enum Visibility {Private, Public}` of visibilities at the `TypedProgram` level, as at that point I know they only exist for the entry point
- Change `fmt::Display` for `TypedProgram` so that when displaying the entry point, the array of visibilities is used

Thoughts?

Schaeff:
Had a look at this suggestion and decided it should be a separate PR because a lot of tests have to be rewritten

changelogs/unreleased/1224-dark64
```diff
@@ -0,0 +1 @@
+Disallow the use of the `private` keyword on non-entrypoint functions
```
Schaeff:
```suggestion
Disallow the use of the `private` and `public` keywords on non-entrypoint functions
```

zokrates_cli/examples/compile_errors/non_main_private_args.zok
```diff
@@ -0,0 +1,8 @@
+def mul(private field a) -> field { // `private` should not be allowed here
```
Schaeff, line 1: we should also test for `public` here

Schaeff (approved):
LGTM!

PR #1228, opened 2022-09-20
Implement a cheaper check for boolean array equality

### LINKED ISSUE #810: Check equality of boolean arrays using field chunks
## Abstract

When checking if a boolean array `conditions` only contains `true`, we can simply use the fact that `all(conditions) == (sum(conditions) == len(conditions))` if `len(conditions) < P`, which is the case in practice 

## Motivation

This too expensive now

```
def main(bool[123] a) -> bool:
   return a == [true; 123]
```

This is 2 constraints per bit (excluding the initial bitness check which must happen in any case) but with this change it would be one constraint *total* (not per bit)

## Specification

This is very much an edge case, but it seems like we would catch this in the flattening phase by checking if we're comparing to a constant array of only true values. The same holds for only false, where the sum should be 0.

## Backwards Compatibility

Compatible

Schaeff:
closes #810


zokrates_core/src/static_analysis/boolean_array_comparator.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    #[test]
+    fn simplify_short_array_eq() {
+        // x == y // type bool[2]
+        // should become
+        // [x[0] ? 2**1 : 0 + x[1] ? 2**0 : 0] == [y[0] ? 2**1 : 0 + y[1] ? 2**0 : 0]
+        // a single field is sufficient, as the prime we're working with is 3 bits long, so we can pack up to 2 bits
+
+        let x = a_id("x").annotate(Type::Boolean, 2u32);
+        let y = a_id("x").annotate(Type::Boolean, 2u32);
```
dark64: `a_id("y")`

zokrates_core/src/static_analysis/boolean_array_comparator.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    #[test]
+    fn simplify_long_array_eq() {
+        // x == y // type bool[3]
+        // should become
+        // [x[0] ? 2**2 : 0 + x[1] ? 2**1 : 0, x[2] ? 2**0 : 0] == [y[0] ? 2**2 : 0 + y[1] ? 2**1 : 0 y[2] ? 2**0 : 0]
+
+        let x = a_id("x").annotate(Type::Boolean, 3u32);
+        let y = a_id("x").annotate(Type::Boolean, 3u32);
```
dark64: same as above

dark64 (changes requested):

dark64 (approved):

PR #1234, opened 2022-09-28
Add support for the pallas curve field

Schaeff:
Support the pallas curve for compilation. 
- No proof system is implemented for it as of now
- Refactored the PrimeField macro as it made the assumption that the field belonged to a `PairingEngine` implementor, which is not the case for Pallas (which is not pairing-friendly)


Schaeff:
Close in favor of #1235

PR #1246, opened 2022-11-17
Assembly blocks

dark64:
Introducing assembly blocks


zokrates_analysis/src/assembly_transformer.rs
```diff
@@ -0,0 +1,169 @@
...
+        write!(f, "{}", self.0)
+    }
+}
+
+pub struct AssemblyTransformer;
+
+impl AssemblyTransformer {
+    pub fn transform<T: Field>(p: ZirProgram<T>) -> Result<ZirProgram<T>, Error> {
+        let mut f = AssemblyTransformer;
+        f.fold_program(p)
```
Schaeff: Can be written in a single line if I'm not mistaken

zokrates_analysis/src/assembly_transformer.rs
```diff
@@ -0,0 +1,169 @@
+use std::fmt;
```
Schaeff, line 5: Can you add a specification of what happens in this step? An example can be found at `propagation.rs`

zokrates_analysis/src/assembly_transformer.rs
```diff
@@ -0,0 +1,169 @@
...
+                            c if c == T::from(0) => linear,
+                            c => FieldElementExpression::Add(
+                                box FieldElementExpression::Number(c),
+                                box linear,
+                            ),
+                        };
+
+                        let rhs: FieldElementExpression<'ast, T> = if lqc.quadratic.len() > 1 {
+                            let is_common_factor = |id: &Identifier<'ast>,
+                                                    q: &Vec<(
```
Schaeff: This would be clearer if we introduced some type aliases for quadratic terms

```
type QuadraticTerm<'ast, T> = (T, Identifier<'ast>, Identifier<'ast>);
```
Or even a struct

zokrates_analysis/src/assembly_transformer.rs
```diff
@@ -0,0 +1,169 @@
...
+                                T,
+                                Identifier<'ast>,
+                                Identifier<'ast>,
+                            )>| {
+                                q.iter().all(|(_, i0, i1)| i0.eq(id) || i1.eq(id))
+                            };
+
+                            let common_factor: Option<Identifier<'ast>> =
```
Schaeff: I think this is unnecessarily expensive (O(n**2)) because for all identifiers (linear), it checks if that identifier is a common factor in the entire list (linear). But a common factor in the quadratic terms must be a factor in the first term. So I think something like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=50bfd624c5fd5301c1b4ee620cd95b37) would be better. Or like [this](https://gist.github.com/5e5d525f19fdcbf3c7b374b9c36f048b) to add short circuiting.

zokrates_analysis/src/constant_argument_checker.rs
```diff
@@ -72,6 +72,28 @@ impl<'ast, T: Field> ResultFolder<'ast, T> for ConstantArgumentChecker {
         }
     }
 
+    fn fold_field_expression(
```
Schaeff: Was this missing before?

zokrates_analysis/src/flatten_complex_types.rs
```diff
@@ -449,12 +458,107 @@ impl<'ast, T: Field> Flattener<T> {
     }
 }
 
+#[derive(Default)]
+pub struct ArgumentFinder<'ast, T> {
+    pub identifiers: HashMap<zir::Identifier<'ast>, zir::Type>,
+    _phantom: PhantomData<T>,
```
Schaeff, line 485: Is this necessary?

zokrates_analysis/src/flatten_complex_types.rs
```diff
@@ -449,12 +458,107 @@ impl<'ast, T: Field> Flattener<T> {
     }
 }
 
+#[derive(Default)]
```
Schaeff, line 482: Could you add a specification of what this does with some examples?

zokrates_analysis/src/flatten_complex_types.rs
```diff
@@ -449,12 +458,107 @@ impl<'ast, T: Field> Flattener<T> {
...
+            let function = zir::ZirFunction {
+                signature: zir::types::Signature::default()
+                    .inputs(finder.identifiers.values().cloned().collect())
+                    .outputs(a.iter().map(|a| a.get_type()).collect()),
+                arguments: finder
+                    .identifiers
+                    .into_iter()
+                    .map(|(id, ty)| zir::Parameter {
+                        id: zir::Variable::with_id_and_type(id, ty),
+                        private: false,
```
Schaeff: Visibility doesnt make much sense here, but I think we should use private to avoid confusion.

zokrates_analysis/src/propagation.rs
```diff
@@ -220,6 +217,38 @@ impl<'ast, 'a, T: Field> ResultFolder<'ast, T> for Propagator<'ast, 'a, T> {
...
         match s {
+            TypedStatement::Assembly(statements) => {
+                let mut assembly_statement_buffer = vec![];
+                let mut statement_buffer = vec![];
+
+                for s in statements {
+                    match self.fold_assembly_statement(s)? {
+                        TypedAssemblyStatement::Assignment(assignee, expr) => {
+                            // invalidate the cache
+                            let v = self
```
Schaeff: This seems to be the following case:

```
def main(u32 i, field x) -> field[3]:
   field[3] a = [1, 2, 3];
   asm {
      a[i] <-- x;
      a[i] === x
   }
   return a;
```

Is this allowed? If so, it would be good to have some tests on this to make sure it's equivalent to

```
def main(u32 i, field x) -> field[3]:
   field[3] a = [1, 2, 3];
   a[i] = x;
   return a;
```

And if the assembly constraint is removed, that an unconstrained variable is detected.

zokrates_analysis/src/propagation.rs
```diff
@@ -372,7 +401,9 @@ impl<'ast, 'a, T: Field> ResultFolder<'ast, T> for Propagator<'ast, 'a, T> {
                 match embed_call.arguments.iter().all(|a| a.is_constant()) {
                     true => {
                         let r: Option<TypedExpression<'ast, T>> = match embed_call.embed {
-                            FlatEmbed::BitArrayLe => Ok(None), // todo
+                            FlatEmbed::FieldToBoolUnsafe => Ok(None), // todo
+                            FlatEmbed::BoolToField => Ok(None),       // todo
```
Schaeff: `BoolToField` is not necessary as an embed, it's optimal with `c ? 1 : 0`

zokrates_analysis/src/propagation.rs
```diff
@@ -372,7 +401,9 @@ impl<'ast, 'a, T: Field> ResultFolder<'ast, T> for Propagator<'ast, 'a, T> {
                 match embed_call.arguments.iter().all(|a| a.is_constant()) {
                     true => {
                         let r: Option<TypedExpression<'ast, T>> = match embed_call.embed {
-                            FlatEmbed::BitArrayLe => Ok(None), // todo
+                            FlatEmbed::FieldToBoolUnsafe => Ok(None), // todo
```
Schaeff: This would make sense to implement here, throwing for anything not 0 or 1

zokrates_analysis/src/propagation.rs
```diff
@@ -827,11 +858,135 @@ impl<'ast, 'a, T: Field> ResultFolder<'ast, T> for Propagator<'ast, 'a, T> {
...
+                        box e2.annotate(UBitwidth::B32),
                     )),
                 }
             }
+            FieldElementExpression::Xor(box e1, box e2) => {
+                let e1 = self.fold_field_expression(e1)?;
+                let e2 = self.fold_field_expression(e2)?;
+
+                match (e1, e2) {
+                    (FieldElementExpression::Number(n1), FieldElementExpression::Number(n2)) => {
```
Schaeff, line 938: Shall we also have

X XOR 0 == 0 XOR X == X

zokrates_analysis/src/propagation.rs
```diff
@@ -827,11 +858,135 @@ impl<'ast, 'a, T: Field> ResultFolder<'ast, T> for Propagator<'ast, 'a, T> {
...
+                        UExpression {
+                            inner: UExpressionInner::Value(by),
+                            ..
+                        },
+                    ) => {
+                        let two = BigUint::from(2usize);
+                        let mask: BigUint = two.pow(T::get_required_bits()).sub(1usize);
+
+                        Ok(FieldElementExpression::Number(
+                            T::try_from(n.to_biguint().mul(two.pow(by as usize)).bitand(mask))
```
Schaeff: It seems like we can use `shl` instead of multiplication here?

zokrates_analysis/src/propagation.rs
```diff
@@ -827,11 +858,135 @@ impl<'ast, 'a, T: Field> ResultFolder<'ast, T> for Propagator<'ast, 'a, T> {
...
+                            ..
+                        },
+                    ) if by == 0 => Ok(e),
+                    (
+                        _,
+                        UExpression {
+                            inner: UExpressionInner::Value(by),
+                            ..
+                        },
+                    ) if by as usize >= T::get_required_bits() => {
```
Schaeff, line 1009: Isn't this case covered by the next one?

zokrates_analysis/src/propagation.rs
```diff
@@ -827,11 +858,135 @@ impl<'ast, 'a, T: Field> ResultFolder<'ast, T> for Propagator<'ast, 'a, T> {
...
```
Schaeff, line 1046: Isn't this case covered by the next one?

zokrates_analysis/src/reducer/shallow_ssa.rs
```diff
@@ -124,6 +124,25 @@ impl<'ast, 'a> ShallowTransformer<'ast, 'a> {
...
 
 impl<'ast, 'a, T: Field> Folder<'ast, T> for ShallowTransformer<'ast, 'a> {
+    fn fold_assembly_statement(
+        &mut self,
+        s: TypedAssemblyStatement<'ast, T>,
+    ) -> TypedAssemblyStatement<'ast, T> {
+        match s {
+            TypedAssemblyStatement::Assignment(a, e) => {
+                let e = self.fold_field_expression(e);
+                let a = match a {
```
Schaeff: I realise we may be able to put this in `self.fold_assignee`. To be confirmed though. Then this whole `fold_assembly_statement` would disappear?

zokrates_analysis/src/zir_propagation.rs
```diff
@@ -50,6 +57,24 @@ impl<'ast, T: Field> ZirPropagator<'ast, T> {
...
     type Error = Error;
 
+    fn fold_assembly_statement(
+        &mut self,
+        s: ZirAssemblyStatement<'ast, T>,
+    ) -> Result<ZirAssemblyStatement<'ast, T>, Self::Error> {
+        match s {
+            ZirAssemblyStatement::Assignment(assignees, function) => {
+                for a in &assignees {
+                    self.constants.remove(&a.id);
```
Schaeff: At this point I believe there should be no redefinitions. So this whole implementation could be removed. To be confirmed though (and the same removal could be avoided for `ZirStatement::Definition` too)

zokrates_ast/src/common/embed.rs
```diff
@@ -49,6 +52,12 @@ pub enum FlatEmbed {
 impl FlatEmbed {
     pub fn signature(&self) -> UnresolvedSignature {
         match self {
+            FlatEmbed::FieldToBoolUnsafe => UnresolvedSignature::new()
+                .inputs(vec![UnresolvedType::FieldElement.into()])
+                .output(UnresolvedType::Boolean.into()),
+            FlatEmbed::BoolToField => UnresolvedSignature::new()
```
Schaeff: To be removed

zokrates_ast/src/common/error.rs
```diff
@@ -3,6 +3,7 @@ use std::fmt;
 
 #[derive(Debug, Clone, Serialize, Deserialize, Hash, PartialEq, Eq)]
 pub enum RuntimeError {
+    UserConstraint,
```
Schaeff: I think we should have a more consistent naming with `SourceAssertion`. Maybe `SourceAssemblyConstraint` ?

Schaeff: And ideally include an optional message like for `SourceAssertion`. This would require a syntax change, so for now a line number would be good.

zokrates_ast/src/common/solvers.rs
```diff
@@ -11,19 +12,24 @@ pub enum Solver {
...
+    #[serde(borrow)]
+    Zir(ZirFunction<'ast, T>),
     #[cfg(feature = "bellman")]
     Sha256Round,
     #[cfg(feature = "ark")]
     SnarkVerifyBls12377(usize),
 }
 
-impl fmt::Display for Solver {
+impl<'ast, T: fmt::Debug + fmt::Display> fmt::Display for Solver<'ast, T> {
```
Schaeff: Debug should not be required here

zokrates_ast/src/typed/mod.rs
```diff
@@ -1196,6 +1254,73 @@ pub enum FieldElementExpression<'ast, T> {
...
+    fn from(assignee: TypedAssignee<'ast, T>) -> Self {
+        match assignee {
+            TypedAssignee::Identifier(v) => {
+                let inner = TupleExpression::identifier(v.id);
+                match v._type {
+                    GType::Tuple(tuple_ty) => inner.annotate(tuple_ty),
+                    _ => unreachable!(),
+                }
+            }
+            TypedAssignee::Select(box a, box index) => TupleExpression::select(a.into(), index),
```
Schaeff, line 1250: Shouldn't there be a check of the type here? Or is this checking recursively?

zokrates_ast/src/untyped/mod.rs
```diff
@@ -382,6 +382,33 @@ impl<'ast> fmt::Display for Assignee<'ast> {
     }
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum AssemblyStatement<'ast> {
+    Assignment(AssigneeNode<'ast>, ExpressionNode<'ast>, bool),
```
Schaeff, line 387: This could be made clearer with an enum instead of a boolean

Schaeff: Not a huge deal though.

zokrates_ast/src/zir/lqc.rs
```diff
@@ -0,0 +1,132 @@
...
+            FieldElementExpression::Mult(box left, box right) => {
+                let left = Self::try_from(left)?;
+                let right = Self::try_from(right)?;
+
+                left.try_mul(right)
+            }
+            _ => Err(()),
+        }
+    }
+}
```
Schaeff, line 135: We need some unit tests for this

zokrates_ast/src/zir/mod.rs
```diff
@@ -113,8 +117,41 @@ impl RuntimeError {
     }
 }
 
+#[derive(Clone, PartialEq, Hash, Eq, Debug, Serialize, Deserialize)]
+pub enum ZirAssemblyStatement<'ast, T> {
+    Assignment(
+        #[serde(borrow)] Vec<ZirAssignee<'ast>>,
```
Schaeff, line 123: newline

zokrates_ast/src/zir/mod.rs
```diff
@@ -129,6 +166,7 @@ pub enum ZirStatement<'ast, T> {
         FormatString,
         Vec<(ConcreteType, Vec<ZirExpression<'ast, T>>)>,
     ),
+    Assembly(#[serde(borrow)] Vec<ZirAssemblyStatement<'ast, T>>),
```
Schaeff: this looks weird, is there a way to put this tag somewhere else?

zokrates_ast/src/zir/mod.rs
```diff
@@ -392,16 +449,57 @@ pub enum FieldElementExpression<'ast, T> {
         Box<FieldElementExpression<'ast, T>>,
     ),
     Pow(
+        Box<FieldElementExpression<'ast, T>>,
+        #[serde(borrow)] Box<UExpression<'ast, T>>,
```
Schaeff, line 455: newline

zokrates_cli/examples/book/assembly/bool_to_field.zok
```diff
@@ -0,0 +1,8 @@
+from "EMBED" import bool_to_field;
+
+def main(bool x) -> field {
+    // `x` is constrained by the compiler automatically so we can safely
```
Schaeff: This always the case, bool_to_field is always free and safe, it can be removed from the embeds. This test is not necessary.

zokrates_cli/examples/book/assembly/division.zok
```diff
@@ -0,0 +1,7 @@
+def main(field a, field b) {
+    field mut c = 0;
+    asm {
+        c <-- a / b;
+        a === b * c;
```
Schaeff, line 8: This is an example of the dangers of assembly: here you can set any value as the result of `0 / 0`
In the compiler, we first calculate `1 / b`, checked by `invb * b == 1` (which is unsatisfiable if `b == 0`)
Then return `invb * a`

zokrates_cli/examples/book/assembly/field_to_bool.zok
```diff
@@ -0,0 +1,12 @@
+from "EMBED" import field_to_bool_unsafe;
+
+def main(field x) -> bool {
+    // we constrain `x` to be 0 or 1
+    asm {
+        x * (x - 1) === 0;
+    }
+    // we can treat `x` as `bool` afterwards, as we constrained it properly
+    // `field_to_bool_unsafe` call does not produce any extra constraints
```
Schaeff:
```suggestion
    // we can convert `x` to `bool` afterwards, as we constrained it properly
    // if we failed to constrain `x` to `0` or `1`, the call to `field_to_bool_unsafe` introduces undefined behavior
    // `field_to_bool_unsafe` call does not produce any extra constraints
```

Schaeff: Something along those lines

zokrates_cli/src/bin.rs
```diff
@@ -258,7 +259,7 @@ mod tests {
 
             let interpreter = zokrates_interpreter::Interpreter::default();
 
-            let res = interpreter.execute(artifacts.prog(), &[Bn128Field::from(0)]);
+            let res = interpreter.execute(artifacts.prog(), &[Bn128Field::from(0u32)]);
```
Schaeff: why is this change required?

Schaeff (commented):
Partial review

zokrates_analysis/src/constant_argument_checker.rs
dark64: I had to move this check here from propagation, this was an unrelated issue to this PR but one of my assembly tests was failing because of it. We had this check in propagation which would fail in this case:

```
def test<N>() -> field {
    field mut a = 1;
    for u32 i in 0..N {
        a = a * 2;
    }
    field b = 2**N; //  Non-constant exponent `N_1` detected during static analysis
    return b;
}
```

The code above will error out on develop. It seems we call propagation a few times but in one of the calls `N_1` is not available as a constant. Also, it seems like this is the correct place to put this check as we do check shifts in this module.

zokrates_codegen/src/lib.rs
```diff
@@ -2215,6 +2220,44 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+        &mut self,
+        statements_flattened: &mut FlatStatements<'ast, T>,
+        stat: ZirAssemblyStatement<'ast, T>,
+    ) {
+        match stat {
+            ZirAssemblyStatement::Assignment(assignees, function) => {
+                let outputs: Vec<Variable> = assignees
+                    .iter()
+                    .map(|a| {
+                        self.layout
```
Schaeff: The assignees should not already be defined, same as for `ZirStatement::Definition`. So we can call `use_variable` directly.

zokrates_core/src/imports.rs
```diff
@@ -147,6 +147,14 @@ impl Importer {
                     id: symbol.get_alias(),
                     symbol: Symbol::Flat(FlatEmbed::Unpack),
                 },
+                "field_to_bool_unsafe" => SymbolDeclaration {
+                    id: symbol.get_alias(),
+                    symbol: Symbol::Flat(FlatEmbed::FieldToBoolUnsafe),
+                },
+                "bool_to_field" => SymbolDeclaration {
```
Schaeff: to remove

zokrates_core/src/semantics.rs
```diff
@@ -1132,12 +1132,18 @@ impl<'ast, T: Field> Checker<'ast, T> {
                     // for declaration signatures, generics cannot be ignored
                     generics.0.insert(
                         generic.clone(),
-                        UExpression::identifier(self.id_in_this_scope(generic.name()).into())
-                            .annotate(UBitwidth::B32),
+                        UExpression::identifier(
+                            self.id_in_this_scope(generic.name().into()).into(),
```
Schaeff: Can we instead change `id_in_this_scope` to take `I: Into<_>` as input so we don't have to change each call?

zokrates_core/src/semantics.rs
```diff
@@ -1132,12 +1132,18 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+                        )
+                        .annotate(UBitwidth::B32),
                     );
 
                     //we don't have to check for conflicts here, because this was done when checking the signature
-                    self.insert_into_scope(generic.name(), Type::Uint(UBitwidth::B32), false);
+                    self.insert_into_scope(
```
Schaeff: Same here

zokrates_core/src/semantics.rs
```diff
@@ -1173,7 +1179,7 @@ impl<'ast, T: Field> Checker<'ast, T> {
                         ty,
                         is_mutable,
                     };
-                    match self.scope.insert(id, info) {
+                    match self.scope.insert(id.into(), info) {
```
Schaeff: Same

zokrates_core/src/semantics.rs
```diff
@@ -1782,6 +1788,79 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+            AssemblyStatement::Assignment(assignee, expression, constrained) => {
+                let assignee = self.check_assignee(assignee, module_id, types)?;
+                let checked_e = self.check_expression(expression, module_id, types)?;
+
+                let e = match checked_e {
+                    TypedExpression::FieldElement(e) => Ok(e),
+                    TypedExpression::Int(e) => Ok(FieldElementExpression::try_from_int(e).unwrap()),
+                    _ => Err(ErrorInner {
+                        pos: Some(pos),
+                        message: "Only field element expressions are allowed in the assembly block"
```
Schaeff: The right hand side of an assembly assignment must be a field element, found {}

zokrates_core/src/semantics.rs
```diff
@@ -1782,6 +1788,79 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+        let pos = stat.pos();
+
+        match stat.value {
+            AssemblyStatement::Assignment(assignee, expression, constrained) => {
+                let assignee = self.check_assignee(assignee, module_id, types)?;
+                let checked_e = self.check_expression(expression, module_id, types)?;
+
+                let e = match checked_e {
+                    TypedExpression::FieldElement(e) => Ok(e),
+                    TypedExpression::Int(e) => Ok(FieldElementExpression::try_from_int(e).unwrap()),
```
Schaeff: Is it fine to unwrap here?

zokrates_core/src/semantics.rs
```diff
@@ -1782,6 +1788,79 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+                    true => {
+                        let e = FieldElementExpression::block(vec![], e);
+                        match assignee.get_type() {
+                            Type::FieldElement => Ok(vec![
+                                TypedAssemblyStatement::Assignment(assignee.clone(), e.clone()),
+                                TypedAssemblyStatement::Constraint(assignee.into(), e),
+                            ]),
+                            _ => Err(ErrorInner {
+                                pos: Some(pos),
+                                message: "Assignee must be of type `field`".to_string(),
```
Schaeff: found {}

zokrates_core/src/semantics.rs
```diff
@@ -1782,6 +1788,79 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+            AssemblyStatement::Constraint(lhs, rhs) => {
+                let lhs = self.check_expression(lhs, module_id, types)?;
+                let rhs = self.check_expression(rhs, module_id, types)?;
+
+                let (lhs, rhs) = match (lhs, rhs) {
+                    (TypedExpression::FieldElement(lhs), TypedExpression::FieldElement(rhs)) => {
+                        Ok((lhs, rhs))
+                    }
+                    (TypedExpression::FieldElement(lhs), TypedExpression::Int(rhs)) => {
+                        Ok((lhs, FieldElementExpression::try_from_int(rhs).unwrap()))
```
Schaeff: Is it fine to unwrap here?

zokrates_core/src/semantics.rs
```diff
@@ -1782,6 +1788,79 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+                    (TypedExpression::Int(lhs), TypedExpression::FieldElement(rhs)) => {
+                        Ok((FieldElementExpression::try_from_int(lhs).unwrap(), rhs))
+                    }
+                    (TypedExpression::Int(lhs), TypedExpression::Int(rhs)) => Ok((
+                        FieldElementExpression::try_from_int(lhs).unwrap(),
+                        FieldElementExpression::try_from_int(rhs).unwrap(),
+                    )),
+                    _ => Err(ErrorInner {
+                        pos: Some(pos),
+                        message: "Only field element expressions are allowed in the assembly block"
```
Schaeff: found {}

zokrates_core/src/semantics.rs
```diff
@@ -1791,6 +1870,18 @@ impl<'ast, T: Field> Checker<'ast, T> {
         let pos = stat.pos();
 
         match stat.value {
+            Statement::Assembly(statements) => {
+                let mut checked_statements = vec![];
+                for s in statements {
+                    checked_statements.push(
```
Schaeff: I think we can call `extend` here so that we don't need to `flatten` later

zokrates_core/src/semantics.rs
```diff
@@ -2049,7 +2140,7 @@ impl<'ast, T: Field> Checker<'ast, T> {
         let pos = assignee.pos();
         // check that the assignee is declared
         match assignee.value {
-            Assignee::Identifier(variable_name) => match self.scope.get(&variable_name) {
+            Assignee::Identifier(variable_name) => match self.scope.get(&variable_name.into()) {
```
Schaeff: Same, it seems like `scope` itself should deal with the conversions, not the caller

zokrates_core/src/semantics.rs
```diff
@@ -4495,7 +4596,7 @@ mod tests {
 
         let mut scope = Scope::default();
         scope.insert(
-            "b",
+            std::borrow::Cow::Borrowed("b"),
```
Schaeff: Same, then the tests wouldn't have to change

zokrates_core_test/tests/tests/assembly/binary_sub.zok
```diff
@@ -0,0 +1,34 @@
+def bin_sub<N>(field[N] a, field[N] b) -> field[N] {
```
Schaeff, line 3: Can we specify what this function does? Also, what happens if the inputs are not 0 or 1?

zokrates_core_test/tests/tests/assembly/division.json
```diff
@@ -0,0 +1,26 @@
+{
+  "curves": ["Bn128"],
+  "max_constraint_count": 2,
+  "tests": [
```
Schaeff, line 4: This needs a test for `[0, 0]` which fails at a constraint (cf comment in other review)

zokrates_core_test/tests/tests/assembly/gates/not.zok
```diff
@@ -0,0 +1,7 @@
+def main(field inp) -> field {
+    field mut out = 0;
+    asm {
+        out <== 1 + inp - 2*inp;
```
Schaeff: Why not just `1 - inp` ?

zokrates_core_test/tests/tests/assembly/sha256_maj.json
```diff
@@ -0,0 +1,5 @@
+{
+  "curves": ["Bn128"],
+  "max_constraint_count": 2,
+  "tests": []
```
Schaeff: Let's add tests here

zokrates_core_test/tests/tests/embed/bool_to_field.json
```diff
@@ -0,0 +1,26 @@
+{
+  "curves": ["Bn128"],
```
Schaeff: to remove with the embed

zokrates_core_test/tests/tests/embed/bool_to_field.zok
```diff
@@ -0,0 +1,6 @@
+from "EMBED" import bool_to_field;
```
Schaeff: to remove with the embed

zokrates_core_test/tests/tests/embed/field_to_bool.zok
```diff
@@ -0,0 +1,9 @@
+from "EMBED" import field_to_bool_unsafe;
+
+def main(field x) -> bool {
+    asm {
+        x * (x - 1) === 0;
```
Schaeff, line 5: this doesnt need to be wrapped in assembly

zokrates_interpreter/src/lib.rs
```diff
@@ -157,11 +160,79 @@ impl Interpreter {
...
         assert_eq!(inputs.len(), expected_input_count);
 
         let res = match solver {
+            Solver::Zir(func) => {
+                use zokrates_ast::zir::result_folder::ResultFolder;
+                assert_eq!(func.arguments.len(), inputs.len());
+
+                let constants = func
+                    .arguments
+                    .iter()
```
Schaeff, line 178: I think we can move here, `into_iter`, so then we don't have to clone

zokrates_interpreter/src/lib.rs
```diff
@@ -157,11 +160,79 @@ impl Interpreter {
...
+                                bitwidth
+                            )),
+                        },
+                    })
+                    .collect::<Result<HashMap<_, _>, _>>()?;
+
+                let mut propagator = zokrates_analysis::ZirPropagator::with_constants(constants);
+
+                let folded_function = propagator
+                    .fold_function(func.clone())
```
Schaeff, line 218: Let's add a function to the propagator impl so that we do not need the folder trait in scope here

zokrates_js/index.d.ts
```diff
@@ -60,6 +60,7 @@ declare module "zokrates-js" {
     snarkjs?: {
       program: Uint8Array;
     };
+    constraintCount?: number;
```
Schaeff, line 63: Is this related to this PR?

zokrates_js/lib.js
```diff
@@ -16,6 +16,7 @@ module.exports = (pkg) => {
         {
           program: ptr.program(),
           abi: ptr.abi(),
+          constraintCount: ptr.constraint_count(),
```
Schaeff, line 19: Same

zokrates_js/src/lib.rs
```diff
@@ -43,6 +41,7 @@ pub struct CompilationResult {
     program: Vec<u8>,
     abi: Abi,
     snarkjs_program: Option<Vec<u8>>,
+    constraint_count: u32,
```
Schaeff, line 44: Same

zokrates_js/tests/tests.js
```diff
@@ -44,6 +44,7 @@ describe("tests", () => {
         );
         assert.ok(artifacts);
         assert.ok(artifacts.snarkjs === undefined);
+        assert.equal(artifacts.constraintCount, 1);
```
Schaeff, line 47: same

zokrates_parser/src/zokrates.pest
```diff
@@ -52,7 +52,7 @@ _mut = {"mut"}
 
 
 // Statements
-statement = { (iteration_statement // does not require semicolon
+statement = { (iteration_statement | asm_statement // does not require semicolon
```
Schaeff, line 55: Why not have semi colons? Doesn't their absence make parsing more annoying?

zokrates_parser/src/zokrates.pest
```diff
@@ -66,6 +66,15 @@ return_statement = { "return" ~ expression? }
 definition_statement = { typed_identifier_or_assignee ~ "=" ~ expression }
 assertion_statement = {"assert" ~ "(" ~ expression ~ ("," ~ quoted_string)? ~ ")"}
 
+op_asm_assign = @{"<--"}
+op_asm_assign_constrain = @{"<=="}
+
+asm_assignment = { assignee ~ (op_asm_assign | op_asm_assign_constrain) ~ expression }
+asm_constraint = { expression ~ "===" ~ expression }
+
+asm_statement_inner = { (asm_assignment | asm_constraint) ~ semicolon ~ NEWLINE* }
```
Schaeff: These NEWLINEs :/

Schaeff: More of a question though, I like it without semi colons visually, except the lack of consistency

zokrates_pest_ast/src/lib.rs
```diff
@@ -431,6 +433,60 @@ mod ast {
         pub span: Span<'ast>,
     }
 
+    #[derive(Debug, PartialEq, Eq, Clone)]
+    pub enum AssignmentOperator {
+        Assign,
+        AssignConstrain,
+    }
+
+    impl<'ast> FromPest<'ast> for AssignmentOperator {
```
Schaeff: I think this can be derived the same as way as line 477 below?

zokrates_test/tests/out_of_range.rs
```diff
@@ -42,7 +42,7 @@ fn lt_field() {
     assert!(interpreter
         .execute(
             res.prog(),
-            &[Bn128Field::from(10000), Bn128Field::from(5555)]
+            &[Bn128Field::from(10000u32), Bn128Field::from(5555u32)]
```
Schaeff: why is this change required?

Schaeff (commented):
Rest of the review, except docs

zokrates_analysis/src/propagation.rs
dark64: The case after is if both sides are constants, this one will return `0` no matter the expression on the left if `by >= T::get_required_bits()`

zokrates_analysis/src/propagation.rs
dark64: Same comment as above

zokrates_ast/src/common/solvers.rs
dark64: it is needed because of `T`, otherwise we cannot use `{:?}` in this implementation

zokrates_ast/src/typed/mod.rs
dark64: its recursive

zokrates_analysis/src/reducer/shallow_ssa.rs
dark64: Got a few failing tests after doing this, something is going on that messes with the versions. This can be improved in a different PR maybe.

zokrates_ast/src/zir/mod.rs
dark64: yeah outside Assembly,

```
    #[serde(borrow)]
    Assembly(Vec<ZirAssemblyStatement<'ast, T>>),
 ```

File: zokrates_analysis/src/flatten_complex_types.rs (hunk 6 above)
Comment (dark64) [reply zokrates_analysis/src/flatten_complex_types.rs:485]: seems so, otherwise compiler struggles to infer T

File: zokrates_analysis/src/flatten_complex_types.rs (hunk 8 above)
Comment (dark64) [reply zokrates_analysis/src/flatten_complex_types.rs]: I went with a default value, visibility doesn't make much sense here so I'm not sure what is "better".

File: zokrates_parser/src/zokrates.pest (hunk 57 above)
Comment (dark64) [reply zokrates_parser/src/zokrates.pest:55]: because we don't expect `;` after `{ }`, same as for the for-loop block

File: zokrates_parser/src/zokrates.pest (hunk 58 above)
Comment (dark64) [reply zokrates_parser/src/zokrates.pest]: > These NEWLINEs :/

good catch on this one, I changed the newline thing on develop and merged it without changing this as well

File: zokrates_test/tests/out_of_range.rs (hunk 60 above)
Comment (dark64) [reply zokrates_test/tests/out_of_range.rs]: I need to revert this back, I somehow changed something in a way that I had to do this which is not the case anymore

File: zokrates_pest_ast/src/lib.rs (hunk 59 above)
Comment (dark64) [reply zokrates_pest_ast/src/lib.rs]: I couldn't find a way to do this, so I took a similar approach as we do for binary operators.

Comment (Schaeff) [reply zokrates_pest_ast/src/lib.rs]: You can add an `op_asm` rule to the parser which is an OR of the two, and with a small change to `from_pest` you can use:

```rust
    #[pest_ast(rule(Rule::op_asm))]
    pub enum AssignmentOperator {
        Assign(AssignOperator),
        AssignConstrain(AssignConstrainOperator),
    }

    #[pest_ast(rule(Rule::op_asm_assign))]
    pub struct AssignOperator;

    #[pest_ast(rule(Rule::op_asm_assign_constrain))]
    pub struct AssignConstrainOperator;    
```

I'd favor consistency with what we already use.

zokrates_parser/src/zokrates.pest
Schaeff: Oh I see, thought this was after each internal asm statement

zokrates_parser/src/zokrates.pest
Schaeff: does it work without semi colons on each inner assembly statement?

zokrates_core_test/tests/tests/embed/field_to_bool.zok
dark64: The purpose of this was to show the constraint in the book (in the assembly section)

zokrates_analysis/src/flatten_complex_types.rs
Schaeff: Yeah it's possible that you have to pass `T` explicitly, but it could be better than having this extra field.

zokrates_js/index.d.ts
dark64: This was a change you introduced so we can show constraint count in the playground, we can keep this in, no?

zokrates_analysis/src/flatten_complex_types.rs
Schaeff: I think private is better

zokrates_parser/src/zokrates.pest
dark64: no, each inner statement must have the semi-colon

zokrates_interpreter/src/lib.rs
dark64: cannot move `func.arguments` because its behind a shared reference

zokrates_ast/src/common/solvers.rs
Schaeff: Do we have to use Debug in the impl? In general in the codebase we keep Debug and Display separate

zokrates_core_test/tests/tests/embed/field_to_bool.zok
Schaeff: This file is not in the book folder though?

zokrates_parser/src/zokrates.pest
Schaeff: Oh right, that makes sense. Feel free to close if you're happy with this then.

zokrates_analysis/src/propagation.rs
dark64: I think we agreed on not allowing this because having an unknown index will cause the constraint to be non-quadratic

zokrates_ast/src/common/solvers.rs
dark64: Yeah I'll just implement display separately without using debug stuff

zokrates_core_test/tests/tests/embed/field_to_bool.zok
dark64: Discussed this in the call, the reason this is here is to test the code with actual values. This code is present in the book examples.

zokrates_core/src/semantics.rs
dark64: refactored this part a bit so there are no unsafe unwraps

zokrates_core_test/tests/tests/assembly/bitify.zok
```diff
@@ -0,0 +1,21 @@
+def bitify<N>(field num) -> field[N] {
+    field[N] mut out = [0; N];
+    field mut lc1 = 0;
+    field mut e2 = 1;
+    for u32 i in 0..N {
+        asm {
+            out[i] <-- (num >> i) & 1;
```
Schaeff: This is using little-endian, while the rest of the codebase uses big-endian. I think it would make sense to have consistency and use big-endian everywhere.

Schaeff (commented):
Just this little endian thing!

Schaeff (approved):

PR #1254, opened 2022-12-12
Allow user-provided randomness in setup

dark64:


zokrates_js/src/util.rs
```diff
@@ -26,3 +28,23 @@ pub fn normalize_path(path: PathBuf) -> PathBuf {
     }
     ret
 }
+
+pub fn get_seeded_rng(entropy: &str) -> StdRng {
```
Schaeff: Would it make sense to put this in `zokrates_proof_systems` instead of duplicating it in `zokrates_js` and `zokrates_cli` ?

Schaeff: If this happens, some dependencies can be removed

Schaeff (commented):
Nice!

Schaeff (approved):

Issue #1251, opened 2022-12-06
Add JubJub curve to stdlib to support signature verification on curve bl12_381


alv-around:
## Abstract

Zokrates supports proof generation based on the `bl12_381` curve. However, in order verify a signature on the `bl12_381` a user have to implement the `jubjub` embedded curve from scratch. There is also 

## Motivation

Similarly to the `baby jujub`, the adding an instance of the `jubjub` curve  in the stdlib would simplify the signature verification for the curve `bl12_381`.

## Specification

Add `jubjubParams.zok` in the `ecc` package of zokrates_stdlib

## Backwards Compatibility

This feature should not introduce any backward compatibility. Users who would like to use this feature would have to update Zokrates to a newer version, or copy and edit the source code  according to their local version.

alv-around:
@dark64 I was going to work on this feature, but I see you have come far already. I can overtake if you are currently working on other things, if not let me know how I can support.

PR #1259, opened 2023-01-04
Add jubjub twisted edwards curve to stdlib

dark64:
https://github.com/Zokrates/ZoKrates/issues/1251


zokrates_stdlib/stdlib/hashes/pedersen/512bitBool.zok
```diff
@@ -1,8 +1,8 @@
 import "utils/multiplexer/lookup3bitSigned" as sel3s;
 import "utils/multiplexer/lookup2bit" as sel2;
-import "ecc/edwardsAdd" as add;
-import "ecc/edwardsCompress" as edwardsCompress;
-from "ecc/babyjubjubParams" import BABYJUBJUB_PARAMS;
+import "ecc/babyjubjub/compress";
+from "ecc/edwards" import add;
+from "ecc/babyjubjub/params" import EDWARDS_A, EDWARDS_D;
```
alv-around: The pedersen hash works currently only with `bn128` but from my understanding it should also work with bls12_381 ([see](https://github.com/iden3/iden3-docs/blob/master/source/iden3_repos/research/publications/zkproof-standards-workshop-2/pedersen-hash/pedersen.rst#background).

I would say, that implementing the pedersen hash for both curves is out of the scope of this ticket. Should I maybe create a follow-up ticket for the it? I would also add the `#pragma curve bn128` macro in the mean time for the pedersen hash, to avoid misunderstandings.

dark64: Yes! Feel free to open a PR pointing to this branch

alv-around:
@dark64 rebasing the source branch should solve the failing check

alv-around (approved):
I keep in mind format and changelog for next time ✍️ everything else looks good to me

alv-around (approved):

alv-around:
@Schaeff can you please review the branch? 👾

zokrates_stdlib/stdlib/utils/pack/bool/nonStrictUnpack256.zok
```diff
@@ -1,11 +1,11 @@
...
-
 import "./unpack_unchecked";
 
 // Unpack a field element as 256 big-endian bits
 // Note: uniqueness of the output is not guaranteed
 // For example, `0` can map to `[0, 0, ..., 0]` or to `bits(p)`
-def main(field i) -> bool[256] {
-    bool[254] b = unpack_unchecked(i);
-    return [false, false, ...b];
+def main(field i, u32 bit_size) -> bool[256] {
```
dark64, line 6: TODO: After a second look, I figure this function should have better documentation

cc @alvaro-alonso

alv-around: @dark64 good point!

What about:

```rust
import "./unpack_unchecked";

// Unpack a field element as 256 big-endian bits.
// Since the size of p can differ among different fields, 
// p_size has to be given to derive the correct padding.
// For example, for bn128 |p|=254, for bls12-381 |p|=255.
// Note: uniqueness of the output is not guaranteed
// For example, `0` can map to `[0, 0, ..., 0]` or to `bits(p)`
def main(field i, u32 p_size) -> bool[256] {
    assert(p_size <= 256);
    u32 padding_size = 256 - p_size;
    bool[p_size] b = unpack_unchecked(i);
    return [...[false; padding_size], ...b];
}
```

I can not push the changes to the branch. And I think a PR is overkill 💀

PR #1268, opened 2023-01-26
Optimize assembly solvers

dark64:
- Changed a file structure to a section-based one
- Moved zir solver to a separate section of the file to improve output size and performance
- Optimized evaluation of linear combinations in the interpreter (~15-20% faster)


zokrates_ast/src/zir/substitution.rs
```diff
@@ -0,0 +1,26 @@
+use super::{Folder, Identifier};
+use std::{collections::HashMap, marker::PhantomData};
+use zokrates_field::Field;
+
+pub struct ZirSubstitutor<'a, 'ast, T> {
+    substitution: &'a HashMap<Identifier<'ast>, Identifier<'ast>>,
+    _phantom: PhantomData<T>,
```
Schaeff: This is not required

zokrates_codegen/src/lib.rs
```diff
@@ -2237,11 +2237,36 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
                 let outputs: Vec<Variable> = assignees
                     .into_iter()
                     .map(|assignee| self.use_variable(&assignee))
                     .collect();
-                let directive = FlatDirective::new(outputs, Solver::Zir(function), inputs);
+
+                let mut substitution_map = HashMap::default();
+                for (index, p) in function.arguments.iter().enumerate() {
+                    let new_id = format!("i{}", index).into();
+                    substitution_map.insert(p.id.id.clone(), new_id);
```
dark64: This substitution is done so that we can deduplicate function calls later. We want functions with the same logic to hash to the same thing eg. `(a, b) -> a + b` is logically the same as `(c, d) -> c + d`. Both functions are transformed to `(i0, i1) -> i0 + i1`

zokrates_ast/src/ir/serialize.rs
```diff
@@ -58,33 +63,195 @@ impl<
...
+    pub return_count: u32,
+    pub sections: Vec<Section>,
+}
+
+impl ProgHeader {
+    pub fn write<W: Write>(&self, mut w: W) -> std::io::Result<()> {
+        w.write_all(&self.magic)?;
+        w.write_all(&self.version)?;
+        w.write_all(&self.curve_id)?;
+        w.write_u32::<LittleEndian>(self.constraint_count)?;
```
dark64, line 124: I included this in the header so we don't have to `collect` the `Prog` to read the constraint count

zokrates_ast/src/ir/serialize.rs
```diff
@@ -1,14 +1,19 @@
...
 use serde_cbor::{self, StreamDeserializer};
-use std::io::{Read, Write};
+use std::io::{Read, Seek, Write};
 use zokrates_field::*;
 
 type DynamicError = Box<dyn std::error::Error>;
 
 const ZOKRATES_MAGIC: &[u8; 4] = &[0x5a, 0x4f, 0x4b, 0];
-const ZOKRATES_VERSION_2: &[u8; 4] = &[0, 0, 0, 2];
+const FILE_VERSION: &[u8; 4] = &[3, 0, 0, 0];
```
dark64, line 13: I feel this doesn't really represent a zokrates version, it is more of a file version. I changed it to a little-endian order to align with everything else.

zokrates_analysis/src/flatten_complex_types.rs
```diff
@@ -481,7 +481,7 @@ impl<'ast, T: Field> Flattener<T> {
 // }
 #[derive(Default)]
 pub struct ArgumentFinder<'ast, T> {
-    pub identifiers: HashMap<zir::Identifier<'ast>, zir::Type>,
+    pub identifiers: BTreeMap<zir::Identifier<'ast>, zir::Type>,
```
dark64, line 484: Changed this to `BTreeMap` because it turns out the order is important in some cases

zokrates_core/src/optimizer/duplicate.rs
```diff
@@ -39,14 +39,18 @@ impl<'ast, T: Field> Folder<'ast, T> for DuplicateOptimizer {
...
-        let hashed = hash(&s);
-        let result = match self.seen.get(&hashed) {
-            Some(_) => vec![],
-            None => vec![s],
-        };
-
-        self.seen.insert(hashed);
-        result
+        match s {
+            Statement::Block(s) => s.into_iter().flat_map(|s| self.fold_statement(s)).collect(),
```
dark64, line 43: It seems like we didn't go through the blocks here, which would not optimize in the case of assembly blocks.

zokrates_ast/src/ir/solver_indexer.rs
```diff
@@ -0,0 +1,50 @@
...
+fn hash<T: Field>(f: &ZirFunction<T>) -> Hash {
+    use std::hash::Hash;
+    use std::hash::Hasher;
+    let mut hasher = DefaultHasher::new();
+    f.hash(&mut hasher);
+    hasher.finish()
+}
+
+#[derive(Debug, Default)]
+pub struct SolverIndexer<'ast, T> {
```
dark64, line 22: @Schaeff Do you think we should move this to a different directory?

zokrates_ast/src/zir/identifier.rs
```diff
@@ -34,10 +35,17 @@ impl<'ast> fmt::Display for Identifier<'ast> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             Identifier::Source(s) => write!(f, "{}", s),
+            Identifier::Internal(s) => write!(f, "{}", s),
```
dark64: We have some compiler-introduced variables in this PR (check other comments for clarification) so I assumed it would be best to introduce a new variant here, as the Source one has a `'ast` lifetime.

zokrates_ast/src/ir/serialize.rs
```diff
@@ -58,33 +63,195 @@ impl<
     }
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum SectionType {
+    Arguments = 1,
```
Schaeff: Why doesn't this start from 0?

zokrates_ast/src/ir/serialize.rs
```diff
@@ -58,33 +63,195 @@ impl<
...
+}
+
+#[derive(Debug, Clone)]
+pub struct ProgHeader {
+    pub magic: [u8; 4],
+    pub version: [u8; 4],
+    pub curve_id: [u8; 4],
+    pub constraint_count: u32,
+    pub return_count: u32,
+    pub sections: Vec<Section>,
```
Schaeff: Can this be a static array of size 3, or is it designed to be arbitrary? For example, can you ignore the solver section?

zokrates_ast/src/ir/serialize.rs
```diff
@@ -117,124 +284,108 @@ impl<'de, R: Read>
...
+                let solvers: Vec<Solver<'de, Bls12_381Field>> = {
+                    let mut p = serde_cbor::Deserializer::from_reader(r.by_ref());
+                    Vec::deserialize(&mut p).map_err(|_| String::from("Cannot read solvers"))?
+                };
+
+                let section = header.get_section(SectionType::Constraints).unwrap();
+                r.seek(std::io::SeekFrom::Start(section.offset)).unwrap();
+
+                let p = serde_cbor::Deserializer::from_reader(r);
+                let s = p.into_iter::<Statement<Bls12_381Field>>();
```
Schaeff: Could this could be extracted into a function? Something like

```rust
fn read<'de, T: Field, R: Read>(r: R) -> ProgIterator<'de, T, impl IntoIterator<Item = Statement<'de, T>>>
```

zokrates_ast/src/ir/solver_indexer.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub solvers: Vec<Solver<'ast, T>>,
+    pub index_map: HashMap<u64, usize>,
```
Schaeff:
```suggestion
    pub index_map: HashMap<Hash, usize>,
```

zokrates_ast/src/ir/solver_indexer.rs
Schaeff: I think the best would be to do this earlier in the pipeline, because currently we create a lot of `Solver::Zir` in flattening to then remove them here. Also It would make the `Solver` enum cleaner. Basically this logic could be introduced during flattening where we can repurpose `Solver::Zir` variant to do what `IndexedCall` does here.

zokrates_ast/src/zir/identifier.rs
```diff
@@ -4,13 +4,14 @@ use std::fmt;
 
 use crate::typed::Identifier as CoreIdentifier;
 
-#[derive(Debug, PartialEq, Clone, Hash, Eq, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Clone, Hash, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub enum Identifier<'ast> {
     #[serde(borrow)]
     Source(SourceIdentifier<'ast>),
+    Internal(String),
```
Schaeff: Would this be clearer and sufficient?

```suggestion
    SolverArgument(u8),
```

zokrates_ast/src/zir/substitution.rs
```diff
@@ -0,0 +1,26 @@
+use super::{Folder, Identifier};
+use std::{collections::HashMap, marker::PhantomData};
+use zokrates_field::Field;
+
+pub struct ZirSubstitutor<'a, 'ast, T> {
```
Schaeff: Why is the `'a` lifetime required? Why can't `substitution` be moved into `ZirSubstitutor`?

zokrates_codegen/src/lib.rs
```diff
@@ -2237,11 +2237,36 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+                        .into_iter()
+                        .map(|p| substitutor.fold_parameter(p))
+                        .collect(),
+                    statements: function
+                        .statements
+                        .into_iter()
+                        .flat_map(|s| substitutor.fold_statement(s))
+                        .collect(),
+                    signature: function.signature,
+                };
```
Schaeff: I suspect all this logic can be put into `ZirSubstitutor` instead of reimplementing a visitor here

zokrates_interpreter/src/lib.rs
```diff
@@ -487,7 +498,7 @@ mod tests {
     #[test]
     fn five_hundred_bits_of_1() {
         let inputs = vec![Bn128Field::from(1)];
-        let res = Interpreter::execute_solver(&Solver::Bits(500), &inputs).unwrap();
+        let res = Interpreter::execute_solver(&Solver::Bits(500), &inputs, &[]).unwrap();
```
Schaeff, line 501: The usage of Zir solvers is not covered in these tests

changelogs/unreleased/1268-dark64
```diff
@@ -0,0 +1 @@
+Optimize assembly solver
```
Schaeff:
```suggestion
Reduce compiled program size by deduplicating assembly solvers
```

zokrates_ast/Cargo.toml
```diff
@@ -9,6 +9,7 @@ bellman = ["zokrates_field/bellman", "pairing_ce", "zokrates_embed/bellman"]
 ark = ["ark-bls12-377", "zokrates_embed/ark"]
 
 [dependencies]
+byteorder = "1.4.3"
```
Schaeff, line 12: Is this used?

zokrates_ast/src/common/solvers.rs
```diff
@@ -14,6 +14,7 @@ pub enum Solver<'ast, T> {
     EuclideanDiv,
     #[serde(borrow)]
     Zir(ZirFunction<'ast, T>),
+    IndexedCall(usize, usize),
```
Schaeff: This would benefit from a bit more documentation or a struct

zokrates_ast/src/ir/from_flat.rs
```diff
@@ -24,6 +24,7 @@ pub fn from_flat<'ast, T: Field, I: IntoIterator<Item = FlatStatement<'ast, T>>>
         statements: flat_prog_iterator.statements.into_iter().map(Into::into),
         arguments: flat_prog_iterator.arguments,
         return_count: flat_prog_iterator.return_count,
+        solvers: vec![],
```
Schaeff, line 27: Is this empty because it gets populated later?

zokrates_ast/src/zir/identifier.rs
```diff
@@ -34,10 +35,17 @@ impl<'ast> fmt::Display for Identifier<'ast> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             Identifier::Source(s) => write!(f, "{}", s),
+            Identifier::Internal(s) => write!(f, "{}", s),
         }
     }
 }
 
+impl<'ast> From<String> for Identifier<'ast> {
```
Schaeff: This would be confusing because I'd expect such a conversion to create a `SourceIdentifier`. A method on the Identifier impl seems better.

zokrates_ast/src/zir/substitution.rs
```diff
@@ -0,0 +1,26 @@
...
+            _phantom: PhantomData::default(),
+        }
+    }
+}
+
+impl<'a, 'ast, T: Field> Folder<'ast, T> for ZirSubstitutor<'a, 'ast, T> {
+    fn fold_name(&mut self, n: Identifier<'ast>) -> Identifier<'ast> {
+        match self.substitution.get(&n) {
+            Some(v) => v.clone(),
+            None => n,
```
Schaeff: Shouldn't the substitution fail if the key isn't found?

zokrates_ast/src/zir/substitution.rs
```diff
@@ -0,0 +1,26 @@
+use super::{Folder, Identifier};
+use std::{collections::HashMap, marker::PhantomData};
+use zokrates_field::Field;
+
+pub struct ZirSubstitutor<'a, 'ast, T> {
+    substitution: &'a HashMap<Identifier<'ast>, Identifier<'ast>>,
```
Schaeff: Aren't the values always of the Internal variant? If so this should be made stricter

zokrates_codegen/src/lib.rs
Schaeff: How about variables introduced inside the body though? Shouldn't they also be added to the substitution? 

Also I suspect this should be done earlier in the pipeline, when the `ZirFunction`s are created: they could be created directly in this canonical form, then we don't have to deal with this substitution.

Schaeff (changes requested):

zokrates_ast/Cargo.toml
dark64: yes, in `zokrates_ast/src/ir/serialize.rs`

zokrates_ast/src/ir/from_flat.rs
dark64: yes

zokrates_ast/src/zir/substitution.rs
dark64: Hmm I don't think so, we don't necessarily substitute all variables

zokrates_ast/src/zir/substitution.rs
dark64: I aimed to make this substitutor generic so it could be reused easily, that includes potential cases  in the future where an identifier is not necessarily "internal"

zokrates_ast/src/ir/serialize.rs
dark64: Because this number does not represent an index, but rather an identifier of a section type. I took a similar approach to this https://github.com/iden3/r1csfile/blob/master/doc/r1cs_bin_format.md#sectiontype

zokrates_ast/src/ir/serialize.rs
dark64: I thought about using a static array here, but in some cases, we would write empty sections (eg. if no solvers are used). So we can ignore those yes, even the section for parameters if none are present

zokrates_ast/src/zir/substitution.rs
Schaeff: Why? Can you give an example?

dark64: with the last changes the `None` should be unreachable

zokrates_ast/src/zir/substitution.rs
dark64: Refactored this based on the discussion on the call

zokrates_codegen/src/lib.rs
```diff
@@ -2237,10 +2237,15 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
                     .map(|p| self.layout.get(&p.id.id).cloned().unwrap().into())
                     .collect();
+
                 let outputs: Vec<Variable> = assignees
                     .into_iter()
                     .map(|assignee| self.use_variable(&assignee))
                     .collect();
+
+                let mut substitutor = ZirSubstitutor::default();
+                let function = substitutor.fold_function(function);
```
Schaeff: Should we rename `substitutor` to `canonicaliser` everywhere?

Schaeff (commented):
LGTM, just a small naming nitpick!

zokrates_codegen/src/lib.rs
dark64: yeah makes sense, changed

Schaeff (approved):

PR #1282, opened 2023-02-21
Add Jubjub to ECC stdlib

alv-around:
Add jubjub to stdlib #1251

Features:
- [x] Proof-Of-Ownership
- [x] VerifyEddsa


alv-around:
@dark64 @Schaeff. I finished the implementation for the Jubjub curve in the stdlib.
 
  I had to re-arrange some stuff, because many ecc things were written for 254-bit values, and in the jubjub many are 255. 
 
 I also wrote in a way so in the future it is easy to integrate the [signatures for bls12-377](https://pkg.go.dev/github.com/consensys/gnark-crypto/ecc/bls12-377/twistededwards) and other Edwardstwisted curves

PR #1283, opened 2023-02-22
Refactor reducer to reduce memory usage and runtime

Schaeff:
- Introduce `FrameIdentifier` to isolate variables related to a given call frame
- Simplify the reducer
- Remove call push and pop statements

### DIFF
--- a/zokrates_analysis/src/reducer/inline.rs
+++ b/zokrates_analysis/src/reducer/inline.rs
@@ -26,13 +26,10 @@
 // - The body of the function is in SSA form
 // - The return value(s) are assigned to internal variables
 
-use crate::reducer::ShallowTransformer;
-use crate::reducer::Versions;
-
 use zokrates_ast::common::FlatEmbed;
 use zokrates_ast::typed::types::{ConcreteGenericsAssignment, IntoType};
 use zokrates_ast::typed::CoreIdentifier;
-use zokrates_ast::typed::Identifier;
+
 use zokrates_ast::typed::TypedAssignee;
 use zokrates_ast::typed::UBitwidth;
 use zokrates_ast::typed::{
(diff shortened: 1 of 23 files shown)


zokrates_analysis/src/reducer/inline.rs
```diff
@@ -58,35 +57,40 @@ pub enum InlineError<'ast, T> {
...
     match &s.symbol {
-        TypedFunctionSymbol::There(key) => get_canonical_function(key.clone(), program),
+        TypedFunctionSymbol::There(key) => get_canonical_function(key, program),
         _ => s.clone(),
     }
 }
 
 type InlineResult<'ast, T> = Result<
-    Output<(Vec<TypedStatement<'ast, T>>, TypedExpression<'ast, T>), Vec<Versions<'ast>>>,
+    (
```
dark64: I think we should have a struct for this, it's a bit unclear by just looking at this type what the elements represent

zokrates_analysis/src/reducer/mod.rs
```diff
@@ -325,74 +275,70 @@ impl<'ast, 'a, T: Field> ResultFolder<'ast, T> for Reducer<'ast, 'a, T> {
...
+                    .into_iter()
+                    .flatten();
+
+                let statements = statements
+                    .map(|s| self.propagator.fold_statement(s))
+                    .collect::<Result<Vec<_>, _>>()?
+                    .into_iter()
+                    .flatten();
+
+                statements.collect()
```
dark64, line 326: Just checking, is it possible to collect only once here?

Schaeff: We collect a single time into `Vec`, but we need more collects into `Result<Vec<>, _>` for error handling. I had a look and did not find another easy way to do `vec.into_iter().collect::<Result<Vec<_>, _>>().flatten()` which does not require going through `Vec<_>`. This is something we could improve across the codebase if we have proof that it actually makes a difference and isn't optimised by rustc already. I don't think it's necessary in this PR.

dark64 (approved):

PR #1285, opened 2023-03-01
Add sourcemaps

Schaeff:
Goal: each ir statement has a span linking back to the original source
Non goal: each ir expression has a span (can be a separate PR, and is less useful for analysis as we care mostly about constraints)

Changes:
- factor a lot of the common ast code into `zokrates_ast/src/common`
- make the ast more granular by introducing structs inside enum variants
- add span on all structs and implement `WithSpan` trait for (almost) all AST nodes
- refactor folders to provide some degree of automatic sourcemap propagation (implementing `*_cases` for expressions and statements allows to forget about the span and the output gets assigned to span of the input), though manual span setting is still required in many cases
- add `module_map` at the program level to store the file names for sourcemaps
- some smaller refactors:
  - remove immutable from typed
  - remove gt and ge from typed
  - ~simplify zir identifier (it uses an enum with a single variant before this PR)~
  - rename number to value for `{typed, zir}::FieldElementExpression` and `FlatExpression`


zokrates_analysis/src/assembly_transformer.rs
```diff
@@ -28,143 +32,135 @@ impl AssemblyTransformer {
 impl<'ast, T: Field> ResultFolder<'ast, T> for AssemblyTransformer {
     type Error = Error;
```
Schaeff, line 34: The diff looks big but I only extracted `fold_assembly_statement` into the new folder methods for constraint

zokrates_ast/src/flat/mod.rs
```diff
@@ -47,7 +47,7 @@ pub struct FlatFunctionIterator<'ast, T, I: IntoIterator<Item = FlatStatement<'a
     pub arguments: Vec<Parameter>,
     /// Vector of statements that are executed when running the function
     pub statements: I,
-    /// Number of outputs
+    /// Value of outputs
```
Schaeff: revert

zokrates_ast/src/ir/clean.rs
```diff
@@ -21,8 +24,13 @@ impl<'ast, T: Field, I: IntoIterator<Item = Statement<'ast, T>>> ProgIterator<'a
 
 impl<'ast, T: Field> Folder<'ast, T> for Cleaner {
     fn fold_statement(&mut self, s: Statement<'ast, T>) -> Vec<Statement<'ast, T>> {
+        if s.get_span().is_none() {
+            log::error!("Internal compiler warning: found a statement without source information. Please open an issue https://github.com/Zokrates/ZoKrates/issues/new?template=bug_report.md");
```
dark64: why not `log::warn!` if it's a warning?

zokrates_ast/src/ir/serialize.rs
```diff
@@ -231,13 +235,24 @@ impl<'ast, T: Field, I: IntoIterator<Item = Statement<'ast, T>>> ProgIterator<'a
             section
         };
 
+        // write module map section
+        let module_map = {
+            let mut section = Section::new(SectionType::Solvers);
+            section.set_offset(w.stream_position()?);
+
+            serde_cbor::to_writer(&mut w, &self.module_map)?;
```
dark64, line 243: Just as a reminder, we need to make sure no absolute paths are in this module map. This can be addressed in a different PR.

zokrates_ast/src/typed/mod.rs
```diff
@@ -556,6 +564,23 @@ impl<'ast, T: fmt::Display> fmt::Display for TypedAssignee<'ast, T> {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Hash, Eq, PartialOrd, Ord)]
+pub struct AssertionMetadata {
+    pub file: String,
```
dark64: This field might not be needed anymore, not sure completely but it could be worth looking into it. This can also in some cases contain absolute paths

dark64 (dismissed):
Nice one, I left some comments here but they can be addressed in a different PR. I think this is good as-is.

zokrates_ast/src/ir/clean.rs
Schaeff: changed to eprintln

zokrates_ast/src/ir/serialize.rs
Schaeff: Remap user code and stdlib paths in cli

zokrates_ast/src/typed/mod.rs
Schaeff: This struct is not used but the equivalent in `SourceMetadata` is a bit trickier to remove. I think this should be a separate PR.

Schaeff:
Let's make sure we squash before merge

dark64 (approved):

PR #1287, opened 2023-03-19
Sudoku example fix

Turupawn:
**The problem**

the logic behind `countDuplicates` is invalid because it's doing this:

(incorrect)
```
if(duplicates + e11 == e21)
  duplicates = 1;
else
  duplicates = 0;
```

Instead of this:
(correct)
```
if(e11 == e21)
  duplicates = duplicates + 1;
else
  duplicates = duplicates;
```

**The solution**

I'm using an auxiliary variable to fix this

**Alternative solution**

We can also fix it with an array and a loop. But I went for the auxiliary variable for no particular reason.

```
def countDuplicates(field e11, field e12, field e21, field e22) -> u32 {
    u32[6] mut duplicates = [0,0,0,0,0,0];
    duplicates[0] = e11 == e12 ? 1 : 0;
    duplicates[1] = e11 == e21 ? 1 : 0;
    duplicates[2] = e11 == e22 ? 1 : 0;
    duplicates[3] = e12 == e21 ? 1 : 0;
    duplicates[4] = e12 == e22 ? 1 : 0;
    duplicates[5] = e21 == e22 ? 1 : 0;

    u32 mut count = 0;
    for u32 i in 0..5 {
        count = count + duplicates[i];
    }

    return count;
}
```

And btw, I also added an assert to make sure we check for this validation. I've tested this on the playground and on chain on goerli and with this changes it works correctly.

Thanks for putting adding the examples. It has helped me a lot.
Cheers!

### DIFF
--- a/zokrates_cli/examples/sudoku/sudoku_checker.zok
+++ b/zokrates_cli/examples/sudoku/sudoku_checker.zok
@@ -11,13 +11,20 @@
 // We use a naive encoding of the values as `[1, 2, 3, 4]` and rely on if-else statements to detect duplicates
 
 def countDuplicates(field e11, field e12, field e21, field e22) -> field {
-    field mut duplicates = e11 == e12 ? 1 : 0;
-    duplicates = duplicates + e11 == e21 ? 1 : 0;
-    duplicates = duplicates + e11 == e22 ? 1 : 0;
-    duplicates = duplicates + e12 == e21 ? 1 : 0;
-    duplicates = duplicates + e12 == e22 ? 1 : 0;
-    duplicates = duplicates + e21 == e22 ? 1 : 0;
-    return duplicates;
+    field mut duplicates = 0;
+    field mut duplicatesAux = e11 == e12 ? 1 : 0;
+    duplicates = duplicates + duplicatesAux;
+    duplicatesAux = e11 == e21 ? 1 : 0;
+    duplicates = duplicates + duplicatesAux;
+    duplicatesAux = e11 == e22 ? 1 : 0;
+    duplicates = duplicates + duplicatesAux;
+    duplicatesAux = e12 == e21 ? 1 : 0;
+    duplicates = duplicates + duplicatesAux;
+    duplicatesAux = e12 == e22 ? 1 : 0;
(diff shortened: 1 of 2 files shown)


Schaeff:
Oh good catch, seems like a precedence issue, would a smaller change like the following work?

```diff
-   duplicates = duplicates + e11 == e21 ? 1 : 0;
+   duplicates = duplicates + (e11 == e21 ? 1 : 0);
```

zokrates_cli/examples/sudoku/sudoku_checker.zok
```diff
@@ -74,5 +81,6 @@ def main(field a21, field b11, field b22, field c11, field c22, field d21, priva
     duplicates = duplicates + countDuplicates(b12, b22, d12, d22);
 
     // the solution is correct if and only if there are no duplicates
+    assert(duplicates == 0);
```
Schaeff: This is fine but then let's return nothing

Turupawn:
The smaller change worked! I tested it on the playground and works well :+1: 
I'm also returning nothing on the main function.
Thanks for reviewing this @Schaeff !

Turupawn:
Woops, closed this by accident. Just reopened

Turupawn:
After further thought, I went back returning a bool instead of nothing. I think it makes more sense. Specially for newcomers like myself.
So in this PR I'm only fixing the precedence issue.

Turupawn:
Hey guys, I wanted to know if you guys think something else should be done on this PR. I would greatly appreciate if we merge this because I'm doing educational content and it would be nice to point to this official repo instead of my own.
This is the first piece of documentation for reference, but I plan on doing more:
https://dev.to/turupawn/circuitos-zksnarks-paso-a-paso-4i6n
It rewards the players with tokens when they submit a valid proof that they solved the sudoku.
Cheers!

Schaeff:
Yes we can merge it soon. Can you please just add a changelog at `changelogs/unreleased/1287-Turupawn`? Thanks!

Turupawn:
Added the changelog @Schaeff 
Very appreciated :pray:

Schaeff (approved):

PR #1289, opened 2023-03-24
Optimize backend integrations

dark64:
- Changed witness format to a binary one
- Added `--json` flag to `compute-witness` for debugging purposes
- Removed unnecessary `into_canonical` calls
- Improved performance of `into_bellman` calls


zokrates_ast/src/common/variable.rs
```diff
@@ -33,6 +34,20 @@ impl Variable {
...
+    pub fn read<R: Read>(mut reader: R) -> std::io::Result<Self> {
+        let mut buf = [0; std::mem::size_of::<isize>()];
+        reader.read_exact(&mut buf)?;
+
+        Ok(Variable {
+            id: isize::from_le_bytes(buf),
+        })
+    }
+
     pub fn try_from_human_readable(s: &str) -> Result<Self, &str> {
```
Schaeff: Can this method be removed?

zokrates_ast/src/ir/witness.rs
```diff
@@ -41,54 +41,78 @@ impl<T: Field> Witness<T> {
...
+            .0
+            .iter()
+            .map(|(k, v)| (k.to_string(), serde_json::json!(v.to_dec_string())))
+            .collect::<serde_json::Map<String, serde_json::Value>>();
+
+        serde_json::to_writer_pretty(writer, &map)?;
+        Ok(())
+    }
+
+    pub fn read_json<R: Read>(reader: R) -> io::Result<Self> {
```
Schaeff: If the json is just for debugging purposes I don't think we should give the option to read it

zokrates_ast/src/ir/witness.rs
```diff
@@ -136,34 +160,5 @@ mod tests {
 
             assert_eq!(w, r);
         }
-
```
Schaeff, line 139: Let's have a test for json here

zokrates_ast/src/ir/witness.rs
dark64: We do need this in the integration test, check my changes in `integration.rs`. We only check the output variables in witness (which I assume is because other variables could change). This way we can still easily specify parts of a witness for the tests

Schaeff: I see, then all the relevant code should be moved to the tests as helper functions or tagged with `#[cfg(test)]`

zokrates_ast/src/common/variable.rs
Schaeff: If not then moved to the tests where it's needed

Schaeff (approved):

PR #1304, opened 2023-04-26
Big integer PoC

dark64:
Work in progress


zokrates_stdlib/stdlib/algebra/biginteger.zok
```diff
@@ -0,0 +1,110 @@
...
+        a.limbs[j] = a.limbs[j - n];
+    }
+    for u32 i in 0..n {
+        a.limbs[i] = 0;
+    }
+    return a;
+}
+
+// Big integer multiplication
+def bigint_mul<N, M>(BigInteger<N> a, BigInteger<N> b) -> BigInteger<N> {
```
dark64: A self-note: use a more optimized method, the output should be `BigInteger<2*N>`

PR #1309, opened 2023-05-16
Fix range check

### LINKED ISSUE #1295: Cannot use <= on u32 in assert
## Description

When using `<=` in `assert`, I get a compilation error:

```
The compiler unexpectedly panicked
panicked at 'assertion failed: res.len() <= to', zokrates_codegen/src/lib.rs:2020:21
```

This does not happen for `<`, `>`, or `>=`.

## Environment

- Compiler version: 0.8.5, 0.8.6
- Operating system: macOS 13.1

## Steps to Reproduce

```zokrates
def main(u32 n) {
    assert(n <= 5);
    return;
}
```

Compiling this code returns the error above. `assert(n < 5)` does not trigger the same error. Changing the type of `n` to `field` also does not trigger the error. `assert(n <= 5 ? true : false)` works as a workaround.

dark64:
Closes https://github.com/Zokrates/ZoKrates/issues/1295 #1329


zokrates_codegen/src/lib.rs
```diff
@@ -2014,20 +2020,49 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+
+                        // update the entry
+                        entry.insert(
+                            (0..res.len() - to)
+                                .map(|_| FlatExpression::value(T::zero()))
+                                .chain(bits.clone())
+                                .collect(),
+                        );
+
+                        return bits;
```
Schaeff: This seems a bit more complicated than it could be:

```
// only keep the last `to` values and return the sum of the others
let sum = res.drain(0..res.len().saturating_sub(to)).sum();

// force the sum to be 0
statements_flattened.push_back(sum == 0);

// sanity check that we have at most `to` values
assert_eq!(res.len() <= to);

// return the result left-padded to `to` values
std::iter::repeat(FlatExpression::value(T::zero())).take(to - res.len()).chain(res.clone()).collect()
```

Schaeff: I think this works in all cases (removing the need for branching)

zokrates_codegen/src/lib.rs
```diff
@@ -2593,32 +2640,84 @@ impl<'ast, T: Field> Flattener<'ast, T> {
...
+                        match (lhs, rhs) {
+                            (e, FlatExpression::Value(c)) => self.enforce_constant_le_check(
                                 statements_flattened,
-                                FlatExpression::sub(T::max_value().into(), e),
-                                T::max_value() - c.value,
+                                e,
+                                c.value,
                                 error.into(),
                             ),
+                            // c < e <=> 2^bw - 1 - e < 2^bw - 1 - c
```
Schaeff: The comment refers to `<` but we're processing `<=`

Schaeff (commented):
Nice, some simplifications and typos. I think we should have more test for this also, see recent issue.

zokrates_core_test/tests/tests/range_check/assert_lt_u8_constant.json
```diff
@@ -0,0 +1,67 @@
+{
+  "entry_point": "./tests/tests/range_check/assert_lt_u8_constant.zok",
```
Schaeff, line 2: Should we have a test with a value which is outside the u8 range and check that it fails, or is it handled in some other test for usize inputs?

zokrates_core_test/tests/tests/compare_min_to_max.zok
```diff
@@ -1,10 +0,0 @@
-from "field" import FIELD_MAX;
-
-// /!\ should be called with a = 0
-// as `|a - FIELD_MAX| < 2**(N-2)` the comparison should succeed
-
-def main(field a) -> bool {
-    field p = FIELD_MAX + a;
-    // we added a = 0 to prevent the condition to be evaluated at compile time
-    return a < p;
```
Schaeff, line 9: The other tests down below all do `assert(condition)`. Do we cover all cases which just use the condition without asserting it, for example returning it? They are handled differently in code generation so they should be also tested separately.

Schaeff (approved):

PR #1312, opened 2023-05-20
Fixed a small typo, the right bracket next to bitwidth was mistyped.

only4sim:
Corrected a typo in the documentation, the right bracket next to bitwidth was typed as a left bracket. This has been fixed.


Schaeff:
I think this change is incorrect: `2**bitwidth` is excluded, hence `[`

only4sim:
> I think this change is incorrect: `2**bitwidth` is excluded, hence `[`

Then I think it should be ')' to avoid misunderstanding.

PR #1313, opened 2023-05-23
Propagate embed call in reducer

### LINKED ISSUE #1306: Expected loop bounds to be constant
## Description

Loop bounds are not propagated correctly

## Environment

- Compiler version: 0.8.7
- Operating system: Ubuntu

## Steps to Reproduce

```zokrates
import "utils/casts/field_to_u32";

def main() {
    u32 bound = field_to_u32(2);
    for u32 i in 0..bound {
        log("{}", i);
    }
}
```

Error: `Expected loop bounds to be constant, found 0..bound˙`

Schaeff:
closes #1306

### DIFF
--- a/zokrates_ast/src/typed/identifier.rs
+++ b/zokrates_ast/src/typed/identifier.rs
@@ -8,7 +8,7 @@ pub type SourceIdentifier<'ast> = std::borrow::Cow<'ast, str>;
 pub enum CoreIdentifier<'ast> {
     #[serde(borrow)]
     Source(ShadowedIdentifier<'ast>),
-    Call(usize),
+    Call,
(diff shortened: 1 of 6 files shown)


zokrates_ast/src/typed/identifier.rs
```diff
@@ -8,7 +8,7 @@ pub type SourceIdentifier<'ast> = std::borrow::Cow<'ast, str>;
 pub enum CoreIdentifier<'ast> {
     #[serde(borrow)]
     Source(ShadowedIdentifier<'ast>),
-    Call(usize),
+    Call,
```
Schaeff, line 11: Just a simplification as this is namespaced by the call frame anyway

dark64 (approved):

PR #1318, opened 2023-05-24
Allow composite field type in witness assignment

dark64:


zokrates_core/src/semantics.rs
```diff
@@ -1794,16 +1794,16 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
-                        e.get_type(),
-                    ),
-                })?;
-
                 match constrained {
                     true => {
+                        let e = FieldElementExpression::try_from_typed(e).map_err(|e| ErrorInner {
+                            span: Some(span),
+                            message: format!(
+                                "Expected right hand side of an assembly assignment to be of type field, found {}",
```
Schaeff: This is confusing, did you mean something like "a field element or a complex type containing only field elements"

dark64: This is the case of `<==` where we do expect only `field`

zokrates_analysis/src/zir_propagation.rs
```diff
@@ -78,7 +80,7 @@ impl<'ast, T: Field> ResultFolder<'ast, T> for ZirPropagator<'ast, T> {
                 .into_iter()
                 .flatten()
                 .collect(),
-            ..f
+            signature: f.signature.inputs(inputs),
```
dark64, line 83: in some cases, arguments were filtered out without updating the signature

zokrates_ast/src/typed/types.rs
```diff
@@ -942,6 +942,19 @@ impl<S> GType<S> {
     }
 }
 
+impl<S: PartialEq> GType<S> {
+    pub fn is_composite_of(&self, ty: &Self) -> bool {
```
Schaeff: This seems to only work for basic types, so I think we should either make that clear in the function name or make it work for any type.

zokrates_core/src/semantics.rs
```diff
@@ -1793,20 +1793,32 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+                    }
                 })?;
 
                 match constrained {
-                    true => {
-                        let e = FieldElementExpression::block(vec![], e);
-                        match assignee.get_type() {
-                            Type::FieldElement => Ok(vec![
+                    true => match assignee.get_type() {
+                        // in case of `lhs <== rhs` we expect lhs and rhs to be of type field
```
Schaeff, line 1810: This should be updated in the book

zokrates_core/src/semantics.rs
```diff
@@ -1793,20 +1793,32 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+                        Type::FieldElement => {
+                            let e = FieldElementExpression::try_from_typed(e).map_err(|e| ErrorInner {
+                                    span: Some(span),
+                                    message: format!(
+                                        "Expected right hand side of an assembly assignment to be of type field, found {}",
```
Schaeff: assembly *constrained* assignment?

zokrates_core/src/semantics.rs
```diff
@@ -1818,18 +1830,28 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+                            span: Some(span),
+                            message: format!("Assignee must be of type field, found {}", ty),
+                        }),
+                    },
                     false => {
-                        let e = FieldElementExpression::block(vec![], e);
-                        Ok(vec![
-                            TypedAssemblyStatement::assignment(assignee, e.into()).with_span(span)
-                        ])
+                        // we can allow composite types in case of `<--` as no constraints are generated from this type of statement
```
Schaeff, line 1841: Also add this to the docs

zokrates_core/src/semantics.rs
```diff
@@ -1818,18 +1830,28 @@ impl<'ast, T: Field> Checker<'ast, T> {
...
+                        // a composite type should only consist of field types
+                        match e.get_type().is_composite_of(&Type::FieldElement) {
```
Schaeff: we're using `is_composite_of` only for fields so it would be fine to specialise it for that

Schaeff (changes requested):
Looks clean! Just a few comments

Schaeff (approved):

PR #1328, opened 2023-06-26
Fixing pest-ast exact version

lvella:
Compilation fails if it tries to build against 0.3.4


titaniumtraveler:
Until this pr is landed, the workaround is to explicitly state pest-ast v0.3.3 as dependency.
That way cargo uses the correct version.
```toml
[dependencies]
pest-ast = "=0.3.3"
```

dark64 (approved):

PR #1333, opened 2023-08-02
Add Jubjub to ECC stdlib

alv-around:
Add jubjub to stdlib #1251

Features:
- [x] Proof-Of-Ownership
- [x] VerifyEddsa


alv-around:
I had to re-arrange some stuff, because ecc points for the babyjubjub are 254-bit while for the jubjub 255.

I also wrote in a way so in the future it is easy to integrate the [signatures for bls12-377](https://pkg.go.dev/github.com/consensys/gnark-crypto/ecc/bls12-377/twistededwards) and other Edwardstwisted curves

alv-around:
@dark64 I extended your work to add the JubJub signatures in the stdlib. could please review it? looking forward your feedback :)

zokrates_stdlib/stdlib/utils/pack/bool/nonStrictUnpack256.zok
```diff
@@ -1,11 +1,11 @@
...
 import "./unpack_unchecked";
 
 // Unpack a field element as 256 big-endian bits
 // Note: uniqueness of the output is not guaranteed
 // For example, `0` can map to `[0, 0, ..., 0]` or to `bits(p)`
-def main(field i) -> bool[256] {
-    bool[254] b = unpack_unchecked(i);
-    return [false, false, ...b];
+def main(field i, u32 bit_size) -> bool[256] {
+    assert(bit_size == 254 || bit_size == 255);
```
dark64: Is there a reason why this two bit sizes are only allowed here? Maybe `assert(bit_size <= 256);` would fit better?

alv-around: @dark64 Good point. I found out that for the baby jubjub and jubjub the field bit size if 254 and 255 respectively ([resource](https://neuromancer.sk/std/other/JubJub)). So I went for a strict assertion as the nonStrictUnpack is only used for signatures in the code. 

But you are right that line does not help with code readability. I applied the changes :)

PR #1353, opened 2023-10-23
Remove branch isolation

dark64:
This feature turned out to be expensive and buggy in some cases. After a discussion with @Schaeff, we decided to remove the experimental feature for now.


zokrates_codegen/src/lib.rs
```diff
@@ -275,9 +270,8 @@ impl<T: Field> FlatUExpression<T> {
 
 impl<'ast, T: Field> Flattener<'ast, T> {
     /// Returns a `Flattener` with fresh `layout`.
-    fn new(config: CompileConfig) -> Flattener<'ast, T> {
```
Schaeff, line 278: this should be removed and we should use `Default` instead

Schaeff (changes requested):
Very small suggestion

zokrates_codegen/src/lib.rs
dark64: oh good catch!

Schaeff (approved):

PR #1368, opened 2024-04-03
Fix input/output size in staticcall for addition and scalar_mul

### LINKED ISSUE #1364: Need Clarification on Input/Output Size for `addition`/`scalar_mul` in Generated `verifier.sol`
## Description

The `zokrates export-verifier` command generates a `verifier.sol` file, sourcing its code templates from [solidity.rs](https://github.com/Zokrates/ZoKrates/blob/develop/zokrates_proof_systems/src/solidity.rs). Upon closer inspection of the code, there seems to be something wrong in input and output sizes:
- The `input` is defined as `uint[4]`, totaling `32*4 = 128` bytes, or `0x80` in hexadecimal.
- The `r` variable represents a `G1Point`, amounting to `32*2 = 64` bytes, or `0x40` in hexadecimal.

However, the `staticcall` specifies `0xc0` as the input size and `0x60` as the output size, both of which are larger than necessary.

A similar discrepancy is observed in `scalar_mul`. I haven't exhaustively inspected other calls/static calls.

It appears that this code has been functioning without issues for years, arguably due to luck, as EVM precompiled contracts ignore the input and output sizes for `Bn256Add` and `Bn256ScalarMul`, as seen in the [Bn256Add](https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go#L442) and [Bn256ScalarMul](https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go#L484).

Could anyone confirm this? Thanks.

Earliest commit: fc60aa97e88d7c195517d8d69226763d088268ce

```
    /// @return r the sum of two points of G1
    function addition(G1Point memory p1, G1Point memory p2) internal view returns (G1Point memory r) {
        uint[4] memory input;
        input[0] = p1.X;
        input[1] = p1.Y;
        input[2] = p2.X;
        input[3] = p2.Y;
        bool success;
        assembly {
            success := staticcall(sub(gas(), 2000), 6, input, 0xc0, r, 0x60)
            // Use "invalid" to make gas estimation work
            switch success case 0 { invalid() }
        }
        require(success);
    }
```

```
    /// @return r the product of a point on G1 and a scalar, i.e.
    /// p == p.scalar_mul(1) and p.addition(p) == p.scalar_mul(2) for all points p.
    function scalar_mul(G1Point memory p, uint s) internal view returns (G1Point memory r) {
        uint[3] memory input;
        input[0] = p.X;
        input[1] = p.Y;
        input[2] = s;
        bool success;
        assembly {
            success := staticcall(sub(gas(), 2000), 7, input, 0x80, r, 0x60)
            // Use "invalid" to make gas estimation work
            switch success case 0 { invalid() }
        }
        require (success);
    }
```

dark64:
Closes https://github.com/Zokrates/ZoKrates/issues/1364, merge after https://github.com/Zokrates/ZoKrates/pull/1369

Note: This is not a huge issue as the sizes are clamped in the precompile to the correct values anyways (see [here](https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go#L447))

### DIFF
--- a/zokrates_proof_systems/src/solidity.rs
+++ b/zokrates_proof_systems/src/solidity.rs
@@ -481,7 +481,7 @@ library Pairing {
         input[2] = s;
         bool success;
         assembly {
-            success := staticcall(sub(gas(), 2000), 7, input, 0x80, r, 0x60)
+            success := staticcall(sub(gas(), 2000), 7, input, 0x60, r, 0x40)
             // Use "invalid" to make gas estimation work
             switch success case 0 { invalid() }
         }
--- a/zokrates_proof_systems/src/solidity.rs
+++ b/zokrates_proof_systems/src/solidity.rs
@@ -456,7 +456,7 @@ library Pairing {
         input[3] = p2.Y;
         bool success;
         assembly {
-            success := staticcall(sub(gas(), 2000), 6, input, 0xc0, r, 0x60)
+            success := staticcall(sub(gas(), 2000), 6, input, 0x80, r, 0x40)
             // Use "invalid" to make gas estimation work
             switch success case 0 { invalid() }
         }
--- a/changelogs/unreleased/1368-dark64
+++ b/changelogs/unreleased/1368-dark64
@@ -1,0 +1,1 @@
+Fix staticcall input/output size in `addition` & `scalar_mul`


zokrates_proof_systems/src/solidity.rs
```diff
@@ -481,7 +481,7 @@ library Pairing {
         input[2] = s;
         bool success;
         assembly {
-            success := staticcall(sub(gas(), 2000), 7, input, 0x80, r, 0x60)
+            success := staticcall(sub(gas(), 2000), 7, input, 0x60, r, 0x40)
```
dark64, line 484: https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go#L488

`Bn256ScalarMul` takes 96 bytes (0x60) and outputs G1 which is 64 bytes (0x40)

zokrates_proof_systems/src/solidity.rs
```diff
@@ -456,7 +456,7 @@ library Pairing {
         input[3] = p2.Y;
         bool success;
         assembly {
-            success := staticcall(sub(gas(), 2000), 6, input, 0xc0, r, 0x60)
+            success := staticcall(sub(gas(), 2000), 6, input, 0x80, r, 0x40)
```
dark64, line 459: https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go#L446

`Bn256Add` takes 128 bytes (0x80) and outputs G1 which is 64 bytes (0x40)

Schaeff (approved):

PR #1382, opened 2024-12-14
Fix broken documentation link

futreall:
Fix broken documentation links
  - Replaced `https://github.com/zcash/zips/blob/master/protocol/sapling.pdf`
   with `https://github.com/zcash/zips/blob/main/rendered/protocol/sapling.pdf`.


futreall:
@Schaeff sorry , could you please take a look at my PR so I know what to do next? Thank you.

